PR #14, opened 2017-03-19
Update kit/log To New API

Gouthamve:
NewContext has been removed couple of weeks back.
Ref: https://github.com/go-kit/kit/releases/tag/v0.4.0


fabxc:
Good catch, thanks. That's actually awesome. I found this pattern rather ugly \o/

Issue #9, opened 2017-03-15
Allow fast access to most recent sample


fabxc:
For federation in Prometheus we want fast access to the most recent sample. This is currently not served by the interface we have which only allows forward-streaming access to a series for a time window.
That's the interface actually reflecting properly how data is stored in the database. We do have an efficient cache of the most recent sample but that is strictly speaking an implementation detail. Meaning: extending the `SeriesIterator` interface by a `Last()` call would seem like a workaround at best.

First, need for such a fast access should be benchmarked again in context of the new storage. If it turns out to still be necessary, I'm leaning towards exposing the in-memory block type and making it accessible – potentially as yet another interface.
The federation layer would then use this instead of the general `SeriesIterator` interface.

Sounds sane? @beorn7 @brian-brazil

brian-brazil:
With atomicity and staleness handling, which I hope to get in with the new storage for 2.0, I can see this requiring more involved logic which makes considerations around the current implementation largely moot.

Let's get something working, and then benchmark.

fabxc:
Federation does work modulo #10 which can cause duplicate entries for overlapping selectors right now. So yes, this is is a low priority issue for now.

beorn7:
Just two thoughts at this point (not much time to dig deeper):

- The fast access path for the last sample was not only motivated by federation but also by the fact that rule evaluation are often just accessing the last sample of a series.

- Given that allocation churn of the query layer seems a problem right now, I could imagine that a low-alloc access path to the last sample would help here (note that many advanced setups have a huge amount of rule evaluations going on all the time). (This is based on naïve code reading if how the iteration for for federation works. Might be all wrong.)

fabxc:
> The fast access path for the last sample was not only motivated by federation but also by the fact that rule evaluation are often just accessing the last sample of a series. 

To my knowledge rule evaluations are just regular queries and do not make us of the particular storage call. The query engine does not use it internally either.
I'd argue that almost all rule evaluations that are relevant in cost, use time aggregation anyway and it doesn't apply.

> Given that allocation churn of the query layer seems a problem right now, I could imagine that a low-alloc access path to the last sample would help here (note that many advanced setups have a huge amount of rule evaluations going on all the time). (This is based on naïve code reading if how the iteration for for federation works. Might be all wrong.)

Not so much of a problem. Just way higher than it would need to be. The allocs don't relate so much to access but to what the QE allocates itself. v2 also allocates internally when being queried, but that's not happening in v3.

[...]

PR #17, opened 2017-03-21
[WIP]: Add A Method For Latest Value Per Series

Gouthamve:
This is a very rough stab at #9. 

This is to get my head around the code-base and will primarily be used for benchmarking the need for the latest value retrieval.

@fabxc I have no clue if this works or not :) Need some pointers as to how I can test this.

### DIFF
--- a/db.go
+++ b/db.go
@@ -188,6 +188,25 @@ func Open(dir string, l log.Logger, r prometheus.Registerer, opts *Options) (db 
 	return db, nil
 }
 
+// QueryLatest will return the matching series with their latest value.
+// Useful for federation.
+func (db *DB) QueryLatest(cutoff int64, matchers ...labels.Matcher) []SeriesLatest {
+	db.headmtx.RLock()
+	defer db.headmtx.RUnlock()
+
+	sl := make([]SeriesLatest, 0)
+	for _, h := range db.heads {
+		m := h.Meta()
+		if m.MaxTime < cutoff {
+			continue
+		}
+
+		sl = append(sl, h.SelectLatest(cutoff, matchers...)...)
+	}
+
+	return sl
+}
+
 func (db *DB) run() {
 	defer close(db.donec)
 
--- a/head.go
+++ b/head.go
@@ -228,6 +228,50 @@ func (h *headBlock) Busy() bool {
 	return atomic.LoadUint64(&h.activeWriters) > 0
 }
 
+// SeriesLatest is series with its latest value.
+type SeriesLatest struct {
+	Lset      labels.Labels
+	LastValue float64
+}
+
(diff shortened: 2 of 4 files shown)


head.go
```diff
@@ -228,6 +228,50 @@ func (h *headBlock) Busy() bool {
...
+		}
+
+		for _, abs := range absent {
+			if ms.lset.Get(abs) != "" {
+				continue Outer
+			}
+		}
+
+		sl = append(sl, SeriesLatest{Lset: ms.lset, LastValue: ms.lastValue})
+	}
```
fabxc: After the loop exits we have to check `p.Err()`.

head.go
```diff
@@ -228,6 +228,50 @@ func (h *headBlock) Busy() bool {
 	return atomic.LoadUint64(&h.activeWriters) > 0
 }
 
+// SeriesLatest is series with its latest value.
+type SeriesLatest struct {
+	Lset      labels.Labels
+	LastValue float64
+}
+
+func (h *headBlock) SelectLatest(cutoff int64, matchers ...labels.Matcher) []SeriesLatest {
```
fabxc: As there can be arrow per my other comment, this also needs to return an errors.
It should probably follow convention of other interfaces and return an iterator rather than a materialized slice.

For example, a federation query probably should not but generally may return all series. With 5+ million series, we don't really want to allocate all that space just to subsequently iterate over it to write out on an HTTP connection.

db.go
```diff
@@ -188,6 +188,25 @@ func Open(dir string, l log.Logger, r prometheus.Registerer, opts *Options) (db
...
+
+	sl := make([]SeriesLatest, 0)
+	for _, h := range db.heads {
+		m := h.Meta()
+		if m.MaxTime < cutoff {
+			continue
+		}
+
+		sl = append(sl, h.SelectLatest(cutoff, matchers...)...)
+	}
```
fabxc: This logically doesn't seem correct.

The `heads` list is _not_ sharded across all our series but represents blocks sequential in time, each covering the entire series space.
What we do here is checking every head for the last sample, which generally means we get several samples for the same series, and throwing them all into a big result.

We have to first consider only heads with `m.MaxTime >= cutoff` as you correctly do here.
But then we have to start from the last block, look which series exist there, and take their last sample.
Then go to the block before that and do the same, but only add those series and samples to the result, which are not already in our result so far. Repeat for all considered blocks.

Going with slices here is fine for sketching things out, but we will have to make this all iterator based.
Especially if we want to "merge with priority" across blocks, with sorted iterators we can do so in O(n) whereas the naive approach will be quadratic, cubic, etc... depending on how many heads have to be considered.

PR #22, opened 2017-03-26
Add Sample Back

Gouthamve:
The compilation and tests are broken as head.go requires sample which
has been moved to another package while moving BufferedSeriesIterator.

Duplication seemed better compared to exposing sample from tsdbutil.


fabxc:
Thanks. Forgot to push that fixup from local.

PR #23, opened 2017-03-26
Implement Postings Iterator Over Bytes

Gouthamve:
Closes fabxc/tsdb#18

@fabxc Umm, I think its time to add tests to this. No way for me to know if my changes are working unless I vendor this into Prom and run it. But I have no clue how to test this. Any pointers?

### DIFF
--- a/postings.go
+++ b/postings.go
@@ -1,6 +1,7 @@
 package tsdb
 
 import (
+	"encoding/binary"
 	"sort"
 	"strings"
 )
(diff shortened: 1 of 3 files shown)


Gouthamve:
Ran it with Prometheus over-night, looks like its working (blocks have been persisted and able to query well).

postings.go
```diff
@@ -240,6 +241,40 @@ func (it *listPostings) Err() error {
 	return nil
 }
 
+type bytePostings struct {
```
fabxc: This should probably be named `bigEndianPostings` to be more specific about the contents of the byte slice.
Later, we might want to improve further by using compression schemes on the postings. It will still be a `bytePostings` but with a different encoding scheme.

postings.go
```diff
@@ -240,6 +241,40 @@ func (it *listPostings) Err() error {
...
+}
+
+func (it *bytePostings) Seek(x uint32) bool {
+	num := len(it.list) / 4
+	// Do binary search between current position and end.
+	it.idx += sort.Search(num-it.idx, func(i int) bool {
+		idx := 4 * (it.idx + i)
+		val := binary.BigEndian.Uint32(it.list[idx : idx+4])
+		return val >= x
+	})
```
fabxc: As another little project, what would be interesting here is setting up some benchmarking around seeking for `x` if the current value is close to `x`.

If we seek for 10 and are currently at 8, initiating a binary search might be way more expensive than just streaming forward sequentially.
Aside from proper benchmarking, this probably also needs some understanding around how frequent that case is in the real world when intersecting postings for example.

At this point it's probably a bit premature to think about.

fabxc:
Looks good in general 👍 

The iterator of the byte slice should be easy enough to write unit tests for. For testing it embedded into the `indexWriter` it's indeed a bit more complicated. We would probably end up testing symmetry of `indexWriter` and `indexReader`, i.e. do I get the same data out that I wrote into it?
That of course doesn't protect against violations of the specified serialization format (not properly specified yet actually) if writer and reader both do it.

Gouthamve:
Added tests!

fabxc:
👍

PR #27, opened 2017-03-29
Add mockIndex And Refactor Tests To Use That

Gouthamve:
The test takes a noticeable amount of time to run due to the large size of the data. I think we should have a smaller dataset for this test.

PS: This was done as an exercise to understand the indexing layer better, its OK if this PR is turned down.


fabxc:
That's a more complete test, thanks.

PR #28, opened 2017-04-01
Test headBlock

Gouthamve:
Wrote the mock to understand the implementation. The mock looks a lot more complicated than I fathomed.

Did not write any tests yet, but given the complexity of the mock itself, do we want to test via mocks here?

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/28)

### DIFF
--- a/head.go
+++ b/head.go
@@ -565,6 +565,8 @@ func (h *headBlock) create(hash uint64, lset labels.Labels) *memSeries {
 		lset: lset,
 		ref:  uint32(len(h.series)),
 	}
+	// create the initial chunk and appender
+	s.cut()
 
 	// Allocate empty space until we can insert at the given index.
 	h.series = append(h.series, s)
@@ -627,7 +629,7 @@ func (s *memSeries) append(t int64, v float64) bool {
 
 	var c *memChunk
 
-	if s.app == nil || s.head().samples > 2000 {
+	if s.head().samples > 2000 {
 		c = s.cut()
 		c.minTime = t
 	} else {
--- a/head.go
+++ b/head.go
@@ -410,22 +410,11 @@ func (a *headAppender) Commit() error {
 		return err
 	}
 
-	var (
-		total = uint64(len(a.samples))
-		mint  = int64(math.MaxInt64)
-		maxt  = int64(math.MinInt64)
-	)
+	total := uint64(len(a.samples))
 
 	for _, s := range a.samples {
 		if !a.series[s.ref].append(s.t, s.v) {
 			total--
-		}
-
-		if s.t < mint {
-			mint = s.t
-		}
-		if s.t > maxt {
-			maxt = s.t
 		}
 	}
 
--- a/head_test.go
+++ b/head_test.go
@@ -2,6 +2,7 @@ package tsdb
 
 import (
 	"io/ioutil"
+	"math"
 	"os"
 	"sync"
 	"sync/atomic"
@@ -361,6 +362,9 @@ func (m mockHeadAppender) Commit() error {
 
 	m.h.Lock()
 	defer m.h.Unlock()
+	mint := int64(math.MaxInt64)
+	maxt := int64(math.MinInt64)
+
 	for _, v := range m.lref.lref {
 		for _, rlset := range v {
 			// If series exists then append, else add series and append.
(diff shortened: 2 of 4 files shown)


Gouthamve:
The latest commit exposes a panic on running the e2e test.

Run: ```go test -run Head -v .``` leaves this:
```
➜  tsdb git:(headblock-test) ✗ go test -run Head -v .
=== RUN   TestHeadBlock_e2e
panic: runtime error: index out of range

goroutine 25 [running]:
github.com/fabxc/tsdb.(*headAppender).AddFast(0xc42265c2d0, 0x233d1, 0x15b2fd72a15, 0x3fe3cceb54258383, 0x14, 0xc429a81d90)
	/home/goutham/gocode/src/github.com/fabxc/tsdb/head.go:339 +0x266
github.com/fabxc/tsdb.(*headAppender).Add(0xc42265c2d0, 0xc42009a000, 0xe, 0x14, 0x15b2fd72a15, 0x3fe3cceb54258383, 0x0, 0x0, 0x0)
	/home/goutham/gocode/src/github.com/fabxc/tsdb/head.go:289 +0x3f9
github.com/fabxc/tsdb.ingestScrapes(0xc420077790, 0xc4210d2000, 0x3a98, 0x4000, 0xc427b08018, 0x1, 0x15b2fd72a15, 0x9042c0, 0xc42265c2d0, 0x904400, ...)
	/home/goutham/gocode/src/github.com/fabxc/tsdb/head_test.go:125 +0x11a
created by github.com/fabxc/tsdb.TestHeadBlock_e2e
	/home/goutham/gocode/src/github.com/fabxc/tsdb/head_test.go:101 +0x661
exit status 2
FAIL	github.com/fabxc/tsdb	1.212s
```

The probability of hitting this panic increases as the [number of metrics per scrape](https://github.com/Gouthamve/tsdb/blob/c9206afdce0c852b703961afadca18672ed638c4/head_test.go#L86) are increases.

cc @fabxc

Gouthamve:
This is ready for review @fabxc.

This is a longer E2E test like the one on Index and adds a lot of baggage with a huge mock implementation. The next step would be to test if the queries are returning the right values.

But to keep the codebase clean, I propose we move E2E tests to their own folder along with the mock implementations. Ideally, I would want head_test.go to only have the "unit" tests.

Let me know your thoughts.

Gouthamve:
Also if this direction is good, need some pointers on how we would want to test queries (kind of queries, etc.).

head_test.go
```diff
@@ -72,3 +79,410 @@ func readPrometheusLabels(fn string, n int) ([]labels.Labels, error) {
...
+	values   map[string]stringset
+
+	// The series for this head. The refs are stored in index.
+	series []mockSeries
+	lref   lmap
+
+	activeWriters uint64
+
+	sync.RWMutex
+}
```
fabxc: So this goes pretty far into reproducing the same logic as the implementation it is testing. That is worrying me a bit as we get close to testing the implementation with itself. The simplifications mostly stem from not handling all edge cases the actual implementation does. As we want to test those, we will extend the mock further.

Really wondering whether we should stick with unit tests largely and run e2e tests purely on whether we can query the exact values we wrote in.

head.go
```diff
@@ -410,23 +410,12 @@ func (a *headAppender) Commit() error {
...
 		if !a.series[s.ref].append(s.t, s.v) {
 			total--
 		}
-
-		if s.t < mint {
-			mint = s.t
-		}
-		if s.t > maxt {
-			maxt = s.t
-		}
```
Gouthamve: They were not being used. The MaxTime and MinTime are inputs when creating the headblock.

head_test.go
Gouthamve: I completely agree. So the whole point of this was to actually get the querier working. But it became quite bulky and I couldn't get to querier yet.

The primary motivation to implement a mockHead was for me to understand the behaviour better. By trying to implement a naaive version and then looking at the implementation to figure out the optimisations and details. It helped me understand the codebase really well.

If you think this is unnecessary then it is completely fine to not merge this in. And a huge thanks for reviewing.

Gouthamve:
Closing this due to the issues mentioned above and because an e2e test we merged. If there is anything useful, please specify and I will open a PR with those changes.

PR #31, opened 2017-04-05
Fix Panic When Accessing Uncut memorySeries

Gouthamve:
When calling AddFast, we check the details of the head chunk of the
referred memorySeries. But it could happen that there are no chunks in
the series at all.

Currently, we are deferring chunk creation to when we actually append
samples, but we can be sure that there will be samples if the series is
created. We will be consuming no extra memory by cutting a chunk when we
create the series.

@fabxc This is not seen until now as we are never calling AddFast twice in a 
transaction, but looks like my (broken) tests went this way.

Ref: #28 comment 2

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/31)

### DIFF
--- a/head.go
+++ b/head.go
@@ -565,6 +565,8 @@ func (h *headBlock) create(hash uint64, lset labels.Labels) *memSeries {
 		lset: lset,
 		ref:  uint32(len(h.series)),
 	}
+	// create the initial chunk and appender
+	s.cut()
 
 	// Allocate empty space until we can insert at the given index.
 	h.series = append(h.series, s)
@@ -627,7 +629,7 @@ func (s *memSeries) append(t int64, v float64) bool {
 
 	var c *memChunk
 
-	if s.app == nil || s.head().samples > 2000 {
+	if s.head().samples > 2000 {
 		c = s.cut()
 		c.minTime = t
 	} else {


fabxc:
To be honest, I still don't understand what is happening here.

PromQL tests append several samples for a single series in one transaction. They always use Add but that one uses AddFast under the hood regardless. No panics there.
Also, with appending several samples to a new series in the same transaction we should never be executing L339 where the panic was triggered (https://github.com/prometheus/tsdb/blob/master/head.go#L322).

Gouthamve:
Aah, that, I never realised that we will not hit L322 at all. Okay, this was being triggered by two appenders trying to append to the same series. I fixed my tests after I pushed this fix. 

I know that the invariant is that two appenders never concurrently append to the same series. But when I check again, there is another edgecase in which the panic can be triggered. We create the series but never append a sample to it. 

This can happen if Commit fails here: https://github.com/prometheus/tsdb/blob/master/head.go#L408
If we exit there, then the series would not have a chunk and when a new Appender tries to append to it, we will panic.

Gouthamve:
@fabxc Can you take a look at this when you get time? I think my previous comment specifies a situation when a panic could occur.

A better link to where the commit abort could cause this: https://github.com/prometheus/tsdb/blob/09cd2021de58a41fe51f9f7eb1a8d76bf32f5fc5/head.go#L429-L430

So to clarify again, we try to get the series `headChunk` in `AddFast`, but trying to [retrieve a headChunk](https://github.com/prometheus/tsdb/blob/09cd2021de58a41fe51f9f7eb1a8d76bf32f5fc5/head.go#L356) will panic if there are no chunks.

Now we usually do end up creating a chunk if we create the series but sometimes when we abort on the the case above, we will end up with a series with no chunks.

fabxc:
Yes, your explanation make total sense. I'll resolve the merge conflict locally and push it to master.

PR #34, opened 2017-04-07
Add fast path for latest values

Gouthamve:
Incorporated most of the feedback from previous PR. Re-wrote again as the previous approach was flawed.

Need feedback on the interfaces.

Tests pending.

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/34)

### DIFF
--- a/block.go
+++ b/block.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/oklog/ulid"
 	"github.com/pkg/errors"
+	"github.com/prometheus/tsdb/labels"
 )
 
 // DiskBlock handles reads against a Block of time series data.
@@ -40,6 +41,8 @@ type Block interface {
 type HeadBlock interface {
 	Block
 	Appendable
+
+	SelectLast(cutoff int64, matchers ...labels.Matcher) SeriesSet
 }
 
 // Appendable defines an entity to which data can be appended.
--- a/db.go
+++ b/db.go
@@ -639,7 +639,12 @@ func (l *latestSeriesSet) Next() bool {
 		return false
 	}
 
-	l.cur = l.heads[len(l.heads)-1-l.headIdx].SelectLast(l.cutoff, l.matchers...)
+	h := l.heads[len(l.heads)-1-l.headIdx]
+	if h.Meta().MaxTime < l.cutoff {
+		return false
+	}
+
+	l.cur = h.SelectLast(l.cutoff, l.matchers...)
 	return l.Next()
 }
 
--- a/db.go
+++ b/db.go
@@ -587,102 +587,52 @@ func (a *dbAppender) appenderFor(t int64) (*metaAppender, error) {
 	return nil, ErrNotFound
 }
 
-type latestSeriesSet struct {
-	seen map[uint64][]labels.Labels
-
-	heads   []HeadBlock
-	headIdx int
-
-	cur SeriesSet
-
-	cutoff   int64
-	matchers []labels.Matcher
+type lastValSeriesSet struct {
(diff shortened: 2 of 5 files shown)


Gouthamve:
@fabxc This is ready for review. Fixes #9

Gouthamve:
@fabxc How should I proceed benchmarking this?

fabxc:
Best way is probably to just fill up 20k series with about 60 samples (should roughly be the avg chunk-fullness) and have one benchmark run retrieval of the last sample by a) just doing a regular query and seeking to the the end and b) using the new fast-path.

That should give roughly the comparison we are interested in.

Gouthamve:
@fabxc @brian-brazil 

This is the verdict:
```
➜  tsdb git:(latest-val2) ✗ go test -run=XXX -bench=QueryFor
BenchmarkQueryForLatestVals-8   	  100000	     10499 ns/op
PASS
ok  	github.com/prometheus/tsdb	6.618s
➜  tsdb git:(latest-val2) ✗ go test -run=XXX -bench=Fast    
BenchmarkFastPathForLatestVals-8   	 1000000	      1076 ns/op
PASS
ok  	github.com/prometheus/tsdb	6.498s
```

fabxc:
That's definitely significant. Thanks for adding benchmarks.
Now the remaining question is how much this shares with the total load of a typical federation request. You mentioned that regexp matching slows it down significantly. The benchmark only receives a single series. Was this when adding the regexp lookup overhead for a single series every time?

@mwitkow can hopefully give us some real-world federation stats soon.

Then the question is whether we want to make access to last sample a first class concept or just make it a transparent optimisation if we `Seek` a series >= it's last sample.

mwitkow:
For our federation use case, we federate all metrics that match a given regex naming pattern. In our case it is:
```
'match[]': ['{__name__=~".*::.*"}']
```
(`::`=winning)

We'd love to give you some real life comparsions for federation, but our federation is blocked by 
https://github.com/prometheus/prometheus/issues/2688
and the P2s are dropping like flies due to 
https://github.com/prometheus/prometheus/issues/2689
hence it is hard to provide a comparison directly since the P2 stack doesn't have any data ;)

But even then, we expect our `simleaf` prometheus instances (scrape + short retain + aggregate) to benefit from this massively. They keep a *ton* of metrics and even witht he P1 optimisation of federation scrapes that @beorn7 did for P1 (that resulted in a 10x reduction in CPU) they're barely keeping.

Gouthamve:
There is another issue here. We only consider the headBlocks. I think sometimes a valid `lastValue` could be inside a persisted block, but is that case too esoteric and can be ignored?

Now regarding making it a transparent optimisation on `Seek()`, I am not sure as any `*Iterator` interface implementation has been passed a `chunks.Chunk` at max and we cannot expect a sane way to extract a fast-path from that.

brian-brazil:
It's likely that staleness handling will require the last two samples, possibly more if things get fancy.

Gouthamve:
I am not sure how staleness is being handled, will read up on it today. But my guess is, if it requires the last two samples or more, it should be handled during the append stage directly and not by using a fast-path to retrieve the last values.

i.e, it should be handled [here](https://github.com/prometheus/tsdb/blob/master/head.go#L436).

brian-brazil:
Staleness hasn't even been fully implemented yet, I think we should hold off on optimisations like this until it's looking more solid as otherwise we could be wasting your efforts.

Gouthamve:
The federation already performs better and this optimisation is not required. Closing it now.

PR #35, opened 2017-04-08
Provide functionality to merge SeriesSets

### LINKED ISSUE #10: Provide functionality to merge SeriesSets
Currently we can query multiple SeriesSets via label selectors but cannot merge two of them together while eliminating duplicates.
Internally we have similar functionality, so this is merely about making it accessible. We mostly need this for federation where we can have multiple AND'ed selectors (e.g. `abc{de=~"fg"}`) which are OR'd together.

Either we provide an external method OR'ing SeriesSets or extend the querying interface to accept more complex selectors, which would generally be trees.

Gouthamve:
Fixes #10 Depends on #34 

Does chaining by recursion on mergedSets, unlike implementing [chains on lists like here](https://github.com/prometheus/tsdb/blob/master/querier.go#L482-L484).

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/35)

### DIFF
--- a/block.go
+++ b/block.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/oklog/ulid"
 	"github.com/pkg/errors"
+	"github.com/prometheus/tsdb/labels"
 )
 
 // DiskBlock handles reads against a Block of time series data.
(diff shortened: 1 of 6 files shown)


Gouthamve:
@fabxc This is ready for review. But this extends the PR #34 hence the big diff. Merging that in first will make it easier to check this.

Gouthamve:
Closing this as its implemented in prometheus itself. @fabxc Do we want the functionality inside tsdb itself?

fabxc:
Maybe if it is requested at a later point. Let's not introduce more code
dups right now maybe.

PR #36, opened 2017-04-08
Tests For Querier and Postings

Gouthamve:
Tests for Querier and Postings.

### DIFF
--- a/index_test.go
+++ b/index_test.go
@@ -264,6 +264,14 @@ func TestPersistence_index_e2e(t *testing.T) {
 		}
 		i++
 	}
+
+	for k, v := range values {
+		vals := v.slice()
+
+		require.NoError(t, iw.WriteLabelIndex([]string{k}, vals))
+		require.NoError(t, mi.WriteLabelIndex([]string{k}, vals))
+	}
+
 	all := make([]uint32, len(lbls))
 	for i := range all {
 		all[i] = uint32(i)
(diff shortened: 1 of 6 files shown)


querier.go
```diff
@@ -466,8 +466,8 @@ func (s *chunkSeries) Iterator() SeriesIterator {
 // SeriesIterator iterates over the data of a time series.
 type SeriesIterator interface {
 	// Seek advances the iterator forward to the given timestamp.
-	// If there's no value exactly at ts, it advances to the last value
-	// before tt.
+	// If there's no value exactly at t, it advances to the first value
```
Gouthamve: Fixes #37

Gouthamve:
The commit `b60c206` fixes #43 and I can open a separate PR for it if required.

querier_test.go
```diff
@@ -200,3 +203,641 @@ func expandSeriesIterator(it SeriesIterator) (r []sample, err error) {
...
+
+	cases := []struct {
+		dataIdx int
+
+		mint, maxt int64
+		ms         []labels.Matcher
+		exp        SeriesSet
+	}{
+		{
+			dataIdx: 0,
```
fabxc: No need to complicate things for multiple datasets if we are not using it yet. We never might.

index_test.go
```diff
@@ -306,6 +314,24 @@ func TestPersistence_index_e2e(t *testing.T) {
...
+			require.NoError(t, err)
+
+			strsRes, err := tplsRes.At(i)
+			require.NoError(t, err)
+
+			require.Equal(t, strsExp, strsRes)
+		}
+	}
+
+	require.NoError(t, ir.Close())
```
Gouthamve: Hmm, removed index.go from the checklist of done things. Will test index once I get some free time. But I don't think this should block this PR.

Gouthamve:
Incorporated the feedback, PTAL. 

Also I am caught up right now and will resume adding tests after next week, so it would be great if this could be merged as is.

fabxc:
👍

PR #42, opened 2017-04-12
When checking for amended points, do it in terms of bits.

brian-brazil:
NaN != NaN, so the previous code would incorrectly report
it as changed.

There's also plans to take advantage of the NaN payload,
so look at the entire value.

@fabxc

### DIFF
--- a/head.go
+++ b/head.go
@@ -357,7 +357,7 @@ func (a *headAppender) AddFast(ref uint64, t int64, v float64) error {
 		if t < c.maxTime {
 			return ErrOutOfOrderSample
 		}
-		if c.maxTime == t && ms.lastValue != v {
+		if c.maxTime == t && math.Float64bits(ms.lastValue) != math.Float64bits(v) {
 			return ErrAmendSample
 		}
 	}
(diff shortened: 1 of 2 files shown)


fabxc:
👍

Gouthamve:
Just curious, why are we allowing duplicates (same `ts` and `value`)? Shouldn't we ideally drop any metric that comes with a `ts` that already exists?

Here the comment says we drop duplicate samples, but we are storing exact duplicates: https://github.com/prometheus/tsdb/blob/master/head.go#L648-L651

fabxc:
Yes, you are right! That's a bug.

Gouthamve:
Wait, does it mean we stop comparing the value altogether in the above PR and in `memSeries.append` I quoted above?

Just compare the `ts` and if it matches return `ErrAmendSample`?

Gouthamve:
> There's also plans to take advantage of the NaN payload - @brian-brazil 

Do the plans have anything to do with this?

brian-brazil:
Do we even need this if? Timestamps are already checked in AddFast.

PR #45, opened 2017-04-14
db: Add comment: Appender not goroutine safe

### LINKED ISSUE #44: Concurrent map write in (*headAppender).Add()
I'm seeing a panic due to concurrent map writes in: `(*headAppender).Add()`:
https://github.com/prometheus/tsdb/blob/25d45465189ea9b7f2c894188c06d1853bfa79ba/head.go#L322-L323

Is the intention that the caller to `Add()` should implement their own locking? Is there a reason that `Add()` shouldn't manage a mutex itself?

Is this is a documentation issue, happy to raise a PR add a comment to clarify.

mattbostock:
Fixes #44.

### DIFF
--- a/db.go
+++ b/db.go
@@ -73,6 +73,8 @@ type Options struct {
 
 // Appender allows appending a batch of data. It must be completed with a
 // call to Commit or Rollback and must not be reused afterwards.
+//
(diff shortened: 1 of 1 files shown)


Gouthamve:
:+1:

PR #51, opened 2017-04-18
Make rollback actually rollback.

### LINKED ISSUE #39: dbAppender.Rollback actually Commits
It should be calling rollback.

https://github.com/prometheus/tsdb/blob/778103b45060697e8452f3d00a6e8fe1f11306da/db.go#L657

brian-brazil:
Also add tests for basic commit/rollback
functionality.

Fixes #39

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -12,3 +12,78 @@
 // limitations under the License.
 
 package tsdb
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/prometheus/tsdb/labels"
+)
+
+func TestDataAvailableOnlyAfterCommit(t *testing.T) {
+	tmpdir, _ := ioutil.TempDir("", "test")
+	defer os.RemoveAll(tmpdir)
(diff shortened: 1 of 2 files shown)


db_test.go
```diff
@@ -12,3 +12,81 @@
...
+	if !seriesSet.Next() {
+		t.Fatalf("Error, was expecting a matching series.")
+	}
+}
+
+func TestDataNotAvailableAfterRollback(t *testing.T) {
+	tmpdir, _ := ioutil.TempDir("", "test")
+	defer os.RemoveAll(tmpdir)
+
+	db, err := Open(tmpdir, log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr)), prometheus.NewRegistry(), nil)
```
fabxc: Those should both accept a `nil` and do a sane default.
Logger could arguably useful for test debugging, registry seems generally unnecessary.

brian-brazil: Done

PR #52, opened 2017-04-19
Add vendoring.

brian-brazil:
This fixes CI breaking due to a dependency on the dev-2.0
branch of Prometheus.


brian-brazil:
@fabxc Looks like that was the only thing broken with the CI.

grobie:
It's not common practice to vendor dependencies of libraries.

brian-brazil:
Do you have a suggestion for another way to get the CI passing while dev-2.0 is not master?

grobie:
AFAIK only github.com/prometheus/prometheus/pkg/labels needs to be in
master, we could merge this already.

fabxc:
Yea, packporting pkg/ into master for the transition is the sanest approach

brian-brazil:
https://github.com/prometheus/prometheus/pull/2646

PR #54, opened 2017-04-20
Test we get back out the data we put in.

brian-brazil:
This also adds a utility that'll aid future tests.

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -19,43 +19,59 @@ import (
 	"testing"
 
 	"github.com/prometheus/tsdb/labels"
+	"github.com/stretchr/testify/require"
 )
+
+// Convert a SeriesSet into a form useable with reflect.DeepEqual.
+func readSeriesSet(ss SeriesSet) (map[string][]sample, error) {
+	result := map[string][]sample{}
+
+	for ss.Next() {
+		series := ss.At()
+
+		samples := []sample{}
+		it := series.Iterator()
+		for it.Next() {
+			t, v := it.At()
+			samples = append(samples, sample{t: t, v: v})
+		}
+
+		name := series.Labels().String()
+		result[name] = samples
+		if err := ss.Err(); err != nil {
+			return nil, err
+		}
+	}
+	return result, nil
+}
 
 func TestDataAvailableOnlyAfterCommit(t *testing.T) {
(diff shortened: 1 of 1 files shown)


db_test.go
```diff
@@ -16,11 +16,37 @@ package tsdb
 import (
 	"io/ioutil"
 	"os"
+	"reflect"
 	"testing"
 
 	"github.com/prometheus/tsdb/labels"
 )
 
+type Sample struct {
```
fabxc: Make type and attributes unexported.

db_test.go
```diff
@@ -16,11 +16,37 @@ package tsdb
...
+		samples := []Sample{}
+		it := series.Iterator()
+		for it.Next() {
+			t, v := it.At()
+			samples = append(samples, Sample{T: t, V: v})
+		}
+
+		name := series.Labels().String()
+		result[name] = samples
+	}
```
fabxc: This has to end with `if err := ss.Err(); ...`

db_test.go
```diff
@@ -52,9 +78,12 @@ func TestDataAvailableOnlyAfterCommit(t *testing.T) {
...
 	defer querier.Close()
 
-	seriesSet = querier.Select(labels.NewEqualMatcher("foo", "bar"))
-	if !seriesSet.Next() {
-		t.Fatalf("Error, was expecting a matching series.")
+	seriesSet = readSeriesSet(querier.Select(labels.NewEqualMatcher("foo", "bar")))
+	expected := map[string][]Sample{
+		`{foo="bar"}`: []Sample{{T: 0, V: 0}},
+	}
+	if !reflect.DeepEqual(seriesSet, expected) {
```
fabxc: So we are using the `require` package in a bunch of places in our projects by now.
Many people advocate against only using stdlib stuff, but practically, it does exactly the same and is less noisy. I think we should use it somewhat consistently.

db_test.go
brian-brazil: Done

db_test.go
brian-brazil: I don't see a big problem with using it in tests, the client library is the only place I'd worry about deps.

brian-brazil:
Updated, PTAL.

Issue #50, opened 2017-04-17
Improve time complexity of Postings merge


telendt:
> The procedure for more than two lists of different set operations works similarly. So the number of k set operations merely modifies the factor (O(k*n)) instead of the exponent (O(n^k)) of our worst-case lookup runtime. A great improvement.

-- https://fabxc.org/blog/2017-04-10-writing-a-tsdb/

That's not the most efficient way of performing [k-way merge](https://en.wikipedia.org/wiki/K-Way_Merge_Algorithms). [Your implementation](https://github.com/prometheus/tsdb/blob/778103b45060697e8452f3d00a6e8fe1f11306da/postings.go#L148-L158) has indeed `O(nk)` time complexity, but this short drop-in replacement:

```
func Merge(its ...Postings) Postings {
	l := len(its)
	switch l {
	case 0:
		return nil
	case 1:
		return its[0]
	default:
		m := l / 2
		return newMergePostings(Merge(its[:m]...), Merge(its[m:]...))
	}
}
```
has `O(n log k)` (hope I got it right). There are many other efficient implementations of k-way merge, one of the most popular (I think) is based on binary heap ([`container/heap`](https://golang.org/pkg/container/heap/) in Go land).

I could prepare a couple of implementations and benchmark them. Alternatively, we can take a look how other search engines do that (https://github.com/blevesearch/bleve ?) and go with that.

@fabxc

fabxc:
Awesome! Thanks for digging so deep.

The proper construction of the iterator tree is of course an easy and welcome improvement and my initial oversight.

As stated in the blog post, there seems to be tons of material on this. I roughly went with what was easiest for now, in particular because k is typically very small for our use case. For merge (i.e, regexp matchers in Prometheus), it might indeed get larger than intersections but still `k << n`.

It would indeed be interesting to benchmark the impact of different solutions for the typical inputs we encounter.  
Heaps are certainly interesting, the caveat being that its cache locality might be a lot worse than for a simple list and it would require a re-implementation that allows us to hold the heap in a continuous byte slice (for reading the data structure directly from our mmaped index file). From my understanding sorted lists are pretty established in that regard – which may partially be because they compress really well, which is not as important for us, but also because adding elements is very cheap if happening in-order.

telendt:
> For merge (i.e, regexp matchers in Prometheus), it might indeed get larger than intersections but still k << n.

[...]

PR #57, opened 2017-04-24
Fix various postings implementations

telendt:
* newIntersectPostings and newMergePostings should not advance passed
  postings automatically.
* listPostings and bigEndianPostings should not panic with index out
  of range when At() is called before Next()/Seek().
* Seek() call should advance listPostings/bigEndianPostings.

### DIFF
--- a/postings.go
+++ b/postings.go
@@ -33,7 +33,7 @@ func (p *memPostings) get(t term) Postings {
 	if l == nil {
 		return emptyPostings
 	}
-	return &listPostings{list: l, idx: -1}
+	return newListPostings(l)
 }
 
 // add adds a document to the index. The caller has to ensure that no
@@ -70,18 +70,13 @@ func (e errPostings) Seek(uint32) bool { return false }
 func (e errPostings) At() uint32       { return 0 }
 func (e errPostings) Err() error       { return e.err }
 
-func expandPostings(p Postings) (res []uint32, err error) {
-	for p.Next() {
-		res = append(res, p.At())
-	}
-	return res, p.Err()
-}
+var emptyPostings = errPostings{}
 
 // Intersect returns a new postings list over the intersection of the
 // input postings.
 func Intersect(its ...Postings) Postings {
 	if len(its) == 0 {
-		return errPostings{err: nil}
+		return emptyPostings
 	}
 	a := its[0]
 
@@ -91,8 +86,6 @@ func Intersect(its ...Postings) Postings {
 	return a
 }
 
-var emptyPostings = errPostings{}
-
 type intersectPostings struct {
 	a, b     Postings
 	aok, bok bool
@@ -100,41 +93,44 @@ type intersectPostings struct {
 }
 
 func newIntersectPostings(a, b Postings) *intersectPostings {
-	it := &intersectPostings{a: a, b: b}
-	it.aok = it.a.Next()
-	it.bok = it.b.Next()
-
-	return it
+	return &intersectPostings{a: a, b: b}
 }
 
 func (it *intersectPostings) At() uint32 {
 	return it.cur
 }
 
-func (it *intersectPostings) Next() bool {
+func (it *intersectPostings) doNext(id uint32) bool {
 	for {
-		if !it.aok || !it.bok {
+		if !it.b.Seek(id) {
 			return false
 		}
-		av, bv := it.a.At(), it.b.At()
-
-		if av < bv {
-			it.aok = it.a.Seek(bv)
-		} else if bv < av {
-			it.bok = it.b.Seek(av)
-		} else {
-			it.cur = av
-			it.aok = it.a.Next()
-			it.bok = it.b.Next()
-			return true
+		if vb := it.b.At(); vb != id {
+			if !it.a.Seek(vb) {
+				return false
+			}
+			id = it.a.At()
+			if vb != id {
+				continue
+			}
 		}
+		it.cur = id
+		return true
 	}
 }
 
+func (it *intersectPostings) Next() bool {
+	if !it.a.Next() {
+		return false
+	}
+	return it.doNext(it.a.At())
+}
+
 func (it *intersectPostings) Seek(id uint32) bool {
-	it.aok = it.a.Seek(id)
-	it.bok = it.b.Seek(id)
-	return it.Next()
+	if !it.a.Seek(id) {
+		return false
+	}
+	return it.doNext(it.a.At())
 }
 
 func (it *intersectPostings) Err() error {
@@ -158,17 +154,14 @@ func Merge(its ...Postings) Postings {
 }
 
 type mergedPostings struct {
-	a, b     Postings
-	aok, bok bool
-	cur      uint32
+	a, b        Postings
+	initialized bool
+	aok, bok    bool
+	cur         uint32
 }
 
 func newMergedPostings(a, b Postings) *mergedPostings {
-	it := &mergedPostings{a: a, b: b}
-	it.aok = it.a.Next()
-	it.bok = it.b.Next()
-
-	return it
+	return &mergedPostings{a: a, b: b}
 }
 
 func (it *mergedPostings) At() uint32 {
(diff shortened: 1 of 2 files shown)


postings.go
```diff
@@ -70,18 +70,13 @@ func (e errPostings) Seek(uint32) bool { return false }
 func (e errPostings) At() uint32       { return 0 }
 func (e errPostings) Err() error       { return e.err }
 
-func expandPostings(p Postings) (res []uint32, err error) {
```
telendt: This is only used in tests so I moved it there.

telendt:
@fabxc Imho `merge*` should either:
1. be renamed to `union` (in IR system it's called `disjunction`), or
2. just merge, without removing duplicates

Since it's only used in `blockQuerier.selectSingle` for merging postings associated with the same label but having different values I would go with option 2, since there can be any duplicates between matched postings anyway.

Gouthamve:
Some of these things caught me off-guard too, but I think after getting used to the code-base, these make sense. My thoughts:

1. Once we pass the postings, if we are not okay with advancing the postings, it would mean we want to use the passed postings after passing them. But do we want/allow that?

2. Having to call `Next()` has caught me off-guard in the beginning too, but the pattern is consistent across the other iterators in the repo. But given we are returning an invalid value (a posting which might not exist in the list), do we want people to call `At()` before `Next()`?

3. `Seek()` should not advance postings. It is not consistent with the definition.


Removing de-duplication for `merge*` is not as straight forward. We `Intersect()` the different postings later and we would need to de-duplicate in the same postings inside `Intersect`. I think its better to rename it to `union`.

telendt:
@Gouthamve 
 > 1. Once we pass the postings, if we are not okay with advancing the postings, it would mean we want to use the passed postings after passing them. But do we want/allow that?

In order to improve performance of intersection (https://github.com/prometheus/tsdb/issues/50) I'd like to introduce `Postings.Cost()` (similar to lucene's [`DocIdSetIterator.cost()`](https://lucene.apache.org/core/6_5_0/core/org/apache/lucene/search/DocIdSetIterator.html#cost--)). Now, I can imagine specialization of querier that tries to optimize bool query (sequence of merges/intersections) by minimizing total cost. It the process, it might try to arrange postings in a different ways (constructing temporary `(Intersect/Merge)Postings`, not necessary returned as a final value). If the mentioned constructor would advance passed postings that would make things trickier and IMHO counter-intuitive.

> 2. Having to call Next() has caught me off-guard in the beginning too, but the pattern is consistent across the other iterators in the repo. But given we are returning an invalid value (a posting which might not exist in the list), do we want people to call At() before Next()?

We definitely don't want people to call `At()` before `Next()`. It's fine to return some arbitrary value in such case, but IMHO panic too hard punish ([lucene returns -1](https://lucene.apache.org/core/6_5_0/core/org/apache/lucene/search/DocIdSetIterator.html#docID--) in such case, but they can do it since they use signed ints).

> 3. Seek() should not advance postings. It is not consistent with the definition.

This definition?

	// Seek advances the iterator to value v or greater and returns
	// true if a value was found.
	Seek(v uint32) bool

It's not clear whether it's "advance" or "maybe advance". I would go with the former and make it more obvious by renaming it to `Advance` and documenting that it "advances to the first beyond the current [...]"). But again, this is just my opinion. Is there a case where "maybe advance" behavior makes things easier?

Speaking of which, I would also rename `At()` to something like `Current()`, `Value()` or `TSID()`. I find seek/at confusing -- to me it suggests that it's seekable in both direction (like random access file), which is not true.

@fabxc 
> I have a hunch that switching from holding the values in the iterator rather than indices and creating a shortened slice on every iteration could be more costly in general.

Slice is just a [struct of 3 primitive values](https://github.com/golang/go/blob/87e48c5afdcf5e01bb2b7f51b7643e8901f4b7f9/src/runtime/slice.go#L11-L15) and creating a slice on top of slice is relatively inexpensive (it's just a different view on the same array). But yes, I also don't have any benchmarks to back this up at this point.

postings.go
```diff
@@ -158,24 +154,31 @@ func Merge(its ...Postings) Postings {
...
 func (it *mergePostings) Next() bool {
+	if !it.initialized {
+		if !it.a.Next() {
+			it.aok = false
+		}
+		if !it.b.Next() {
+			it.bok = false
+		}
+		it.initialized = true
+	}
```
telendt: > it seems to just make things more compliciated by adding initialized and checking it on every iteration.

Yes, unfortunately Go does not support coroutines. (In Python, it would be as simple as [this](https://gist.github.com/telendt/3fe7f3db9301d40cb3291c1d05428c21)).

The alternative to flag checking on each iteration is to swap function pointer swap. Would you prefer that?

> Is there any thing fundamentally speaking the one element peeking?
> Also we should at least be able to simplify the ifs to it.aok = it.a.Next() like in the former constructor, right?

I'm not sure I understand. Are you asking what are the disadvantages of advancing `it.(a|b)` in the constructor? If so, I already gave some example to @Gouthamve.

fabxc: Swapping function is likely going to be the slowest alternative.

I was referring to the usage of

```
if !it.a.Next() {
  it.aok = false
}
if !it.b.Next() {
  it.bok = false
}
```

Previously, in the constructor we just did:


```
  it.aok = it.a.Next()
  it.bok = it.b.Next()
```

Semantically seem equivalent. What's speaking against keeping it like that.

And yes, also wondering why the additional attribute instead of keeping the initialization in the constructor. I'm not sure I know which part of your other response you meant.

fabxc:
Background on naming. `At()` used to be `Value()`. But we also have a bunch of other forward-only iterators some of which point to more than a single value. I wanted to converge on some general pattern of iterator within the package to make it more easy to use.

Go doesn't have first class iterators and thus devs are not exactly used to a particular usage. Keeping the naming consistent is an attempt to at least make it clear that all our iterators are to be used in the same way.
So I'd rather stick with `At`.  

I agree that `Seek` does not convey the forward-only semantics all that well and `Advance` seems like a good alternative. It should be changed consistently across all iterators though by reasoning above. 

If we were to change the "maybe" aspect, this should also happen uniformly everywhere. I think
(tm) that always advancing should not break anything. But honestly, I don't share the intuition that not advancing if we are at the value already is an edge case – I think of it as "advance by 0", which seems perfectly valid.  
Is my intuition against what is commonly done in IR systems?

postings.go
telendt: Ah, I get it now.

> And yes, also wondering why the additional attribute instead of keeping the initialization in the constructor. I'm not sure I know which part of your other response you meant.

I meant my response to pt. 1 in this comment:
https://github.com/prometheus/tsdb/pull/57#issuecomment-296604719
(passing single postings to many other constructors, just to calculate it's cost).

fabxc: Oh I see, didn't see the bigger picture including the cost.
What would the cost function be based on? Number of elements or anything further as well?

Given relevant optimisations we can of course store additional meta data as well.

telendt:
@fabxc You're right that slicing on each `Next()` comes with some cost. I'll revert it back to indexing, but I would add simple boundary checks to avoid panic (even if one should not call `At()` before `Next()`), since -- according to my tests -- it does not cost anything:

https://gist.github.com/telendt/b7a9fc0524059dcbf0da1698a3142c98#file-slice_test-go-L44

> I agree that Seek does not convey the forward-only semantics all that well and Advance seems like a good alternative.

Cool, I rename it to `Advance()` then. It should also solve this `go vet` annoyance:

```
$ go vet querier.go
querier.go:491: method Seek(t int64) bool should have signature Seek(int64, int) (int64, error)
querier.go:531: method Seek(t int64) bool should have signature Seek(int64, int) (int64, error)
querier.go:597: method Seek(t int64) (ok bool) should have signature Seek(int64, int) (int64, error)
exit status 1
```

> But honestly, I don't share the intuition that not advancing if we are at the value already is an edge case – I think of it as "advance by 0", which seems perfectly valid.
Is my intuition against what is commonly done in IR systems?

Honest response to that is -- I don't know. `Advance()` is an extension to "regular" iterator. I would rather align with lucene here, since it would make it easier to port some of their code.

fabxc:
While I'm still a bit indecisive on always advancing on calling `Seek`, this contains general changes that fix severe bugs in our current implementation. So merging this – we'll figure out soon enough whether the advancing has any downsides. Probably not :)

We should just make it consistent across iterators eventually.

Thanks a lot for looking into this so deeply and all your comments!

PR #61, opened 2017-04-26
IndexWriter improvements

fabxc:
I started documenting the on-disk format and ran into a bunch of places where I wanted to polish the format a bit.

A bunch of version flags were removed for now as the top-level version guards sufficiently. I wanted to get rid of some complexity we don't know to be needed in the future.

I also added decoding/encoding helpers to make the code more readable and less prone to regressions due to failed bounds checking etc. when making changes.

More checksums were added. Series and chunks, each between 50 and 300 bytes typically now come with a CRC32 each. That is pretty excessive, OTOH the storage overhead is still pretty low and smaller checksums lack hardware support, making them overall more expensive computationally.
By some basic benchmarking, we should be able to validate them on every query without notable overhead. But that is subject to proper benchmarks using prombench later.

We also now properly enforce the call order of IndexWriter methods required by the interface contract and have a table of contents, which can be useful for static analysis too.  
I refrained from making offsets pointing around in the index relative to the sections captured by the TOC. I planned that initially, but it makes potential data recovery very hard/impossible if we lose the base that has to be added to the relative offset.

I've yet to decide on whether to put the CRCs next to the data or to the pointer of the data in the general case. The latter allows us to verify that the data being pointed to is actually the one intended to be referenced.  
However, it does make independent traversal and validation of data harder. It is also at risk of putting several CRCs of data across the file in close proximity and burst errors could destroy several of them at once.

Ideas and opinions on any of those points are greatly appreciated.

@Gouthamve @telendt @brian-brazil


index.go
```diff
@@ -39,10 +41,48 @@ const (
...
 
+type indexWriterSeries struct {
+	labels labels.Labels
+	chunks []*ChunkMeta // series file offset of chunks
+	offset uint32       // index file offset of series reference
+}
+
+type indexWriterSeriesSlice []*indexWriterSeries
+
+func (s indexWriterSeriesSlice) Len() int      { return len(s) }
```
brian-brazil: Do you still need this with the go 1.8 sorting?

index.go
```diff
@@ -120,40 +168,78 @@ func newIndexWriter(dir string) (*indexWriter, error) {
...
-	return err
+func (w *indexWriter) write(bufs ...[]byte) error {
+	for _, b := range bufs {
+		n, err := w.fbuf.Write(b)
+		w.pos += uint64(n)
+		if err != nil {
+			return err
+		}
+		// For now the index file must not grow beyond 4GiB. Some of the fixed-sized
+		// offset references in v1 are only 4 byte large.
```
brian-brazil: bytes

index.go
```diff
@@ -120,40 +168,78 @@ func newIndexWriter(dir string) (*indexWriter, error) {
...
-	wr := io.MultiWriter(w.crc32, w.bufw)
-
-	b := [5]byte{flag, 0, 0, 0, 0}
-	binary.BigEndian.PutUint32(b[1:], uint32(l))
-
-	if err := w.write(wr, b[:]); err != nil {
-		return errors.Wrap(err, "writing header")
+// ensureStage handles transitions between write stages and ensures that IndexWriter
+// methods are called in an order valid for the implementation.
+func (w *indexWriter) ensureStage(s indexWriterStage) error {
```
brian-brazil: Will this behave correctly if you skip a stage?

index.go
```diff
@@ -181,33 +267,27 @@ func (w *indexWriter) writeSymbols() error {
...
+	w.buf2.reset()
 
 	for _, s := range symbols {
-		w.symbols[s] = base + uint32(len(w.b))
+		w.symbols[s] = uint32(w.pos) + headerSize + uint32(w.buf2.len())
 
-		n := binary.PutUvarint(buf[:], uint64(len(s)))
-		w.b = append(w.b, buf[:n]...)
-		w.b = append(w.b, s...)
+		// NOTE: len(s) gives the number of runes, not the number of bytes.
```
brian-brazil: This should be mentioned in the format docs

Documentation/format/index.md
```diff
@@ -0,0 +1,198 @@
...
+│ │                  │ └──────────────────────────┘     │ │
+│ ├──────────────────┼──────────────────────────────────┤ │
+│ │                  │ ┌──────────────────────────┐     │ │
+│ │                  │ │ c_i.mint <varint>        │     │ │
+│ │                  │ ├──────────────────────────┤     │ │
+│ │                  │ │ c_i.maxt <varint>        │     │ │
+│ │ #chunks <varint> │ ├──────────────────────────┤ ... │ │
+│ │                  │ │ ref(c_i.data) <varint>   │     │ │
+│ │                  │ ├──────────────────────────┤     │ │
+│ │                  │ │ crc32(c_i.data) <varint> │     │ │
```
fabxc: Why? The point for varints is to reduce the total size. Otherwise a big endian uint32 would have done.
But series entries are about 70% of our total index file size.

Documentation/format/index.md
```diff
@@ -0,0 +1,198 @@
...
+│ │                  │ └──────────────────────────┘     │ │
+│ └──────────────────┴──────────────────────────────────┘ │
+├─────────────────────────────────────────────────────────┤
+│ CRC32 <4 byte>                                          │
+└─────────────────────────────────────────────────────────┘
+```
+
+The CRC checksum is calculated over the series contents of the index concatenated with the data of its chunks (with encoding byte, without length).
```
fabxc: I should've mentioned that I didn't finish the texts yet. Will update those.

Documentation/format/index.md
brian-brazil: I'm saying that if there's no alignment, that should be explicitly documented.

index.go
fabxc: `sort.Slice` has a notable overhead. As we do this a lot I figured it's worth giving a native implementation.

index.go
fabxc: Yes, if we leave the populate stage we always write the symbols and series no matter which one is next.

brian-brazil: If I jump to Done from Populating, would postings and label index be handled correctly?

Documentation/format/index.md
Gouthamve: Curious, what is meant by alignment here?

brian-brazil: If we were at say byte 5 we might write 3 empty bytes to get us to 8, which is 8 byte aligned and generally faster/more atomic for a CPU to access.

fabxc: Yea, but then we would just use 4 or 8 byte ints to begin with. The whole point here is compression. With the known tradeoff of performance. But this shouldn't matter here.

Our chunk compression obviously does the same and that at a much higher cardinality.

Documentation/format/index.md
```diff
@@ -0,0 +1,198 @@
...
+### Postings
+
+Postings are postings lists that map label pairs to series they occur in.
+
+```
+┌─────────────────────────────────────────────────┐
+│ len <varint>                                    │
+├─────────────────────────────────────────────────┤
+│ ┌─────────────────────────────────────────────┐ │
+│ │ ref(series[0]) <4 byte>                     │ │
```
Gouthamve: Why is this not Varint? Or aligned Varint is CPU is an issue?

fabxc: Because our `Seek()` method does binary search, which needs constant spacing.
If we go with compressed postings, e.g. by using delta varint encoding, we need a skip table unless we want to always decode every single entry.

Suppose you want to intersect two postings and one start with 1,000,000 and the other one with 1. We potentially have to decode 1e6 entries before even starting with intersecting properly.

Gouthamve: Aah, damn. Did not realise that. Now I know when to choose what.

index.go
```diff
@@ -219,64 +299,52 @@ func (w *indexWriter) writeSeries() error {
...
+
+		w.buf2.putHash(w.crc32)
+
+		if err := w.write(w.buf1.get(), w.buf2.get()); err != nil {
+			return errors.Wrap(err, "write series data")
+		}
 	}
-	w.started = true
 
 	return nil
```
Gouthamve: So the 4GB limit imposed above has been constantly been on my mind. If the serialised index can be greater than 4GB, the it is highly likely that the series labels and chunks info inside `w.series` would be a significant value (>2GB by a very conservative estimate?).

But we are keeping this info around even when it is not required. Having another map with `[ref]offset` would give us significant savings as we will be able to purge `w.series` after this call. I am not sure if this optimisation is worth it (it has been bugging me since last night :smile: ) but if it is, I think a TODO would be good here.

fabxc: Yes it could be significant. But it would still be relative to the total memory consumption and some of it, like labels, are just references to data that is in the head block anyway.
When compacting multiple older blocks, the data backing the label sets will point to mmaped regions – so also not as bad a pure heap memory.

I think as we have to keep the data around up until here, the maximum wouldn't change if we purge it a few seconds earlier.  
Compaction spikes are a lot smaller than they used to be. But would be interesting to benchmark further in the future. Would be great if we could eliminate them almost entirely of course.

Documentation/format/index.md
brian-brazil: If we're doing binary search, this data should be aligned as it'll be a little bit faster. It'll cost a max of 3 bytes.

fabxc: Good point!

fabxc: Considering the options where and how to put that padding. I think it's best to make it concern of the serializer and not of the format, i.e. don't make padding part of the entry itself.

The writer can just add padding before and the reference will be correct. This makes sequential reading of all postings not work directly anymore. But the simple rule that 0 bytes should just be skipped until the next entries' header, which must never start with 0, is encountered seems to address this just fine.

This should generally allow us to work with inter-entry paddings without bumping the format.

index.go
fabxc: Yes, they'd simply be omitted entirely. No offset tables would be written and the TOC entries end up zeroed out.

fabxc:
Changed some of the format again – mostly to allow for the alignment of postings and value indices without format extensions. Also made docs a bit clearer and actually added the padding.

Chunk CRC32 moved back next to chunk.

Documentation/format/index.md
```diff
@@ -0,0 +1,205 @@
...
+An offset table stores a sequence of entries that maps a list of strings to an offset. They are used to track label index and postings sections. They are read into memory when an index file is loaded.
+
+```
+┌─────────────────────┬────────────────────┐
+│ len <4b>            │ #entries <4b>      │
+├─────────────────────┴────────────────────┤
+│ ┌──────────────────────────────────────┐ │
+│ │  n = #strs <uvarint>                 │ │
+│ ├──────────────────────┬───────────────┤ │
+│ │ len(str_1) <uvarint> │ str_1 <bytes> │ │
```
Gouthamve: Why are we using full strings instead of using the symbol table?

chunks.go
```diff
@@ -204,21 +210,24 @@ func (w *chunkWriter) WriteChunks(chks ...*ChunkMeta) error {
...
-		if err := w.write(wr, chk.Chunk.Bytes()); err != nil {
+		if err := w.write(chk.Chunk.Bytes()); err != nil {
 			return err
 		}
-		chk.Chunk = nil
-	}
 
-	if err := w.write(w.wbuf, w.crc32.Sum(nil)); err != nil {
-		return err
+		w.crc32.Reset()
```
Gouthamve: Could be replaced by `func (cm *ChunkMeta) hash(h hash.Hash) error` defined above?

Documentation/format/chunks.md
```diff
@@ -0,0 +1,15 @@
...
+The following describes the format of a single chunks file, which is created in the `chunks/` directory of a block. The maximum size per segment file is 512MiB.
+
+Chunks in the files are referenced from the index by the in-file offset in the 4 LSB and the segment sequence number in the bigher 4 MSBs.
+
+```
+┌────────────────────────────────────────┬─────────────────────┐
+│ magic(0x85BD40DD) <4 byte>             │ version(1) <1 byte> │
+├────────────────────────────────────────┴─────────────────────┤
+│ ┌──────────────┬───────────────────┬──────┬────────────────┐ │
+│ │ len <varint> │ encoding <1 byte> │ data │ CRC32 <4 byte> │ │
```
Gouthamve: Curious, why are lengths `varints` and not `uvarints`?

fabxc: Doc error. Are actually uvarints.

Documentation/format/index.md
fabxc: Good question, in the end, there's no real answer I suppose. The tables are just very small so compression is really not needed. Sp complexity-wise, it seemed easier/more robust to just inline it if there's no benefit in using pointers.

chunks.go
fabxc: Oh right.

Documentation/format/index.md
```diff
@@ -0,0 +1,205 @@
...
+│ └───────────────────────────────────┘ │
+└───────────────────────────────────────┘
+```
+
+Every series entry first holds its number of labels, followed by tuples of symbol table references that resemble label name and value. The label pairs are lexicographically sorted.  
+After the labels, the number of indexed chunks is encoded, followed by a sequence of metadata entries containing the chunks minimum and maximum timestamp and a reference to its position in the chunk file. Holding the time range data in the index allows dropping chunks irrelevant to queried time ranges without accessing them directly.  
+
+```
+┌─────────────────────────────────────────────────────────┐
+│ len <varint>                                            │
```
Gouthamve: `uvarint` here too

Documentation/format/index.md
```diff
@@ -0,0 +1,205 @@
...
+│ │                 Postings Table               │ │
+│ ├──────────────────────────────────────────────┤ │
+│ │                      TOC                     │ │
+│ └──────────────────────────────────────────────┘ │
+└──────────────────────────────────────────────────┘
+```
+
+When the index is written, an arbitrary number of padding bytes may be added between the lined out main sections above. When sequentially scanning through the file, any zero bytes after a section's specified length must be skipped.
+
+Most of the sections described below start with a `len` field. It always specifies the number of bytes after them up until the trailing CRC32 checksum. The checksum is always calculated over those `len` bytes.
```
brian-brazil: "just before the trailing" would be a tad clearer.

Documentation/format/index.md
```diff
@@ -0,0 +1,205 @@
...
+
+
+### Symbol Table
+
+The symbol table holds a sorted list of deduplicated strings that occurred in label pairs of the stored series. They can be referenced from subsequent sections and significantly reduce the total index size.
+
+The section contains a sequence of the string entries, each prefixed with the string's length in raw bytes.
```
brian-brazil: You should mention that these are utf-8 encoded

Documentation/format/index.md
```diff
@@ -0,0 +1,205 @@
...
+│ │   series_1                        │ │
+│ ├───────────────────────────────────┤ │
+│ │                 . . .             │ │
+│ ├───────────────────────────────────┤ │
+│ │   series_n                        │ │
+│ └───────────────────────────────────┘ │
+└───────────────────────────────────────┘
+```
+
+Every series entry first holds its number of labels, followed by tuples of symbol table references that resemble label name and value. The label pairs are lexicographically sorted.
```
brian-brazil: s/resemble/contain the/

Documentation/format/index.md
```diff
@@ -0,0 +1,205 @@
...
+├─────────────────────────────────────────────────────────┤
+│ CRC32 <4b>                                              │
+└─────────────────────────────────────────────────────────┘
+```
+
+
+
+### Label Index
+
+The label index indexes holds lists of possible values for label names. Each label index can be a composite index over more than a single label name, which is tracked by `#names`, followed by the total number of entries.
```
brian-brazil: It's not clear to me whether this is a list of label names, or label values. The structure appears to indicate names, but then what's this about composites?

Documentation/format/index.md
```diff
@@ -0,0 +1,205 @@
...
+
+```
+┌─────────────────────────────────────────┐
+│ ref(symbols) <8b>                       │
+├─────────────────────────────────────────┤
+│ ref(series) <8b>                        │
+├─────────────────────────────────────────┤
+│ ref(label indices start) <8b>           │
+├─────────────────────────────────────────┤
+│ ref(label indices table) <8b>           │
```
brian-brazil: This isn't described above

chunks.go
```diff
@@ -43,6 +42,16 @@ type ChunkMeta struct {
 	MinTime, MaxTime int64 // time range the data covers
 }
 
+func (cm *ChunkMeta) hash(h hash.Hash) error {
```
brian-brazil: This comment remains

encoding_helpers.go
```diff
@@ -0,0 +1,157 @@
...
+func (e *encbuf) len() int    { return len(e.b) }
+
+func (e *encbuf) putString(s string) { e.b = append(e.b, s...) }
+func (e *encbuf) putBytes(b []byte)  { e.b = append(e.b, b...) }
+func (e *encbuf) putByte(c byte)     { e.b = append(e.b, c) }
+
+func (e *encbuf) putBE32int(x int)      { e.putBE32(uint32(x)) }
+func (e *encbuf) putBE64int(x int)      { e.putBE64(uint64(x)) }
+func (e *encbuf) putUvarint32(x uint32) { e.putUvarint64(uint64(x)) }
+func (e *encbuf) putUvarint(x int)      { e.putUvarint64(uint64(x)) }
```
brian-brazil: This comment remains

index.go
brian-brazil: That needs to be documented then, though I'd suggest instead making this method do what it's description says it does.

fabxc: It does what it says it does.

brian-brazil: Not by my reading, it can fail to write sections and per your above comment fail to fully populate the TOC.

fabxc: It's not "failing" to do so. It omits them because there's literally no data if people decide not to write anything in those sections. This is transparently handled on read and equals to no data being found.

brian-brazil: I'd expect sections with a reported length of 0 to be written, with the TOC pointing to them.

fabxc: Why does it matter? If a null pointer in the TOC indicates the exact same thing? We cannot get there without the TOC anyway.

brian-brazil: Because it's a) not what the spec says and b) not what the function description says.

fabxc: The function description doesn't say anything how this will be expressed in the serialization.

fabxc: There is literally nothing we could point to if we wanted to. If there are e.g. no postings, there is no outer structure to capture that (because we don't want to buffer all postings in memory). They are just written sequentially, there is no start or zero-entry list we could point to.

brian-brazil: I think it'd be simplest for there to be a zero-entry list, that way there's no need to special case this.

fabxc: I literally just told you there is no outer structure such as a list. *There is nothing we could possible reference*. I updated the TOC docs on how to treat null pointers.

fabxc:
I'm merging this now. The above discussion is kinda revolving in circles and I've further changes to make I don't want to stack onto this.

We can revisit any points in subsequent PRs as they'll touch the same area anyway.

PR #64, opened 2017-04-28
Fix and test ref handling in DB appender

### LINKED ISSUE #40: dbAppender.Add says it's looking at last byte, actually just inverting it
At https://github.com/prometheus/tsdb/blob/778103b45060697e8452f3d00a6e8fe1f11306da/db.go#L522

There's:

```
	// Store last byte of sequence number in 3rd byte of refernece.
	return ref | (uint64(h.meta.Sequence^0xff) << 40), nil
```

The `^` should be a `&`. Similarly in AddFast.

fabxc:
Fixes #40 @brian-brazil


juliusv:
:+1:

brian-brazil:
:+1:

PR #67, opened 2017-04-28
Remove unused skiplist types

juliusv:
@fabxc These aren't used anywhere - of course it's entirely possible that you intend to use them somewhere in the future, in which case feel free to close the PR.


fabxc:
Probably not in the near future. So this can go.

PR #68, opened 2017-04-28
Simplify Labels.Equals()

juliusv:
Also extend the Compare() tests to also test Labels.Equals(). In case that's too ugly, let me know, but it seemed like a convenient reuse of already existing test scenarios for a very similar method.

### DIFF
--- a/labels/labels.go
+++ b/labels/labels.go
@@ -84,7 +84,7 @@ func (ls Labels) Equals(o Labels) bool {
 		return false
 	}
 	for i, l := range ls {
-		if l.Name != o[i].Name || l.Value != o[i].Value {
+		if o[i] != l {
 			return false
 		}
 	}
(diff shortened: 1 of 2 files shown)


labels/labels.go
```diff
@@ -83,8 +83,8 @@ func (ls Labels) Equals(o Labels) bool {
 	if len(ls) != len(o) {
 		return false
 	}
-	for i, l := range ls {
-		if l.Name != o[i].Name || l.Value != o[i].Value {
+	for i, ls := range ls {
```
fabxc: Why changing the name to "ls"? That makes it shadow the label set being iterated over.

juliusv: Whoops, groggy airplane code. Changed back to `l`.

fabxc:
Thanks 👍

PR #70, opened 2017-04-28
[DONTMERGE] Various minor questions

juliusv:
This is not meant for merging, but a PR is a good way for commenting
on lines of code... just some questions in the form of comments before
sending actual changes.

### DIFF
--- a/block.go
+++ b/block.go
@@ -25,6 +25,7 @@ import (
 )
 
 // DiskBlock handles reads against a Block of time series data.
+// XXX: Should we rename "*Block" to "*Segment", as that's what we call these pieces in chunks.go?
 type DiskBlock interface {
 	// Directory where block data is stored.
 	Dir() string
@@ -95,6 +96,7 @@ type BlockMeta struct {
 }
 
 const (
+	// XXX: What is the meaning of these flags?
 	flagNone = 0
 	flagStd  = 1
 )
--- a/chunks.go
+++ b/chunks.go
@@ -61,9 +61,11 @@ type ChunkWriter interface {
 type chunkWriter struct {
 	dirFile *os.File
 	files   []*os.File
-	wbuf    *bufio.Writer
-	n       int64
-	crc32   hash.Hash
+	// XXX: In wal.go, we call the equivalent fields "cur" and "curN".
+	//      Should we make the naming consistent between both?
+	wbuf  *bufio.Writer
+	n     int64
+	crc32 hash.Hash
 
 	segmentSize int64
 }
@@ -91,6 +93,8 @@ func newChunkWriter(dir string) (*chunkWriter, error) {
 	return cw, nil
 }
 
+// XXX: should this be called head instead of tail? We also call the
+// currently active block the "head" instead of the "tail".
 func (w *chunkWriter) tail() *os.File {
 	if len(w.files) == 0 {
 		return nil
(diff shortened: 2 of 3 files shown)


block.go
```diff
@@ -25,6 +25,7 @@ import (
 )
 
 // DiskBlock handles reads against a Block of time series data.
+// XXX: Should we rename "*Block" to "*Segment", as that's what we call these pieces in chunks.go?
```
fabxc: Segment is one of multiple chunk files within a block.
In the end they feel pretty synonymous.  I wouldn't care too much either way. But "block-based" design kinda became the catch phrase for the whole thing already.

block.go
```diff
@@ -95,6 +96,7 @@ type BlockMeta struct {
 }
 
 const (
+	// XXX: What is the meaning of these flags?
```
fabxc: Original serialization format had a lot more versioning flags. As there was only the first version of everything I just added these standard constants.

Ultimately, I cut down on a fair bit of complexity and most version flags in #61.
So those can soon be removed.

chunks.go
```diff
@@ -61,9 +61,11 @@ type ChunkWriter interface {
 type chunkWriter struct {
 	dirFile *os.File
 	files   []*os.File
-	wbuf    *bufio.Writer
-	n       int64
-	crc32   hash.Hash
+	// XXX: In wal.go, we call the equivalent fields "cur" and "curN".
+	//      Should we make the naming consistent between both?
```
fabxc: Yea, more naming consistency cannot hurt.
In #61 I named them `fbuf` and `pos` – those might be the most expressive ones.

chunks.go
```diff
@@ -91,6 +93,8 @@ func newChunkWriter(dir string) (*chunkWriter, error) {
 	return cw, nil
 }
 
+// XXX: should this be called head instead of tail? We also call the
+// currently active block the "head" instead of the "tail".
```
fabxc: Yea, probably makes sense. Or rename current heads to tails?
I cannot see an immediate wrong/right way for this. I kinda went with head based on what the 1.x storage has.

chunks.go
```diff
@@ -216,6 +220,8 @@ func (w *chunkWriter) WriteChunks(chks ...*ChunkMeta) error {
 		chk.Chunk = nil
 	}
 
+	// XXX: Shouldn't the CRC32 length be included in maxLen?
+	// XXX: Shouldn't w.n be incremented here?
```
fabxc: `w.n` is incremented in every call to `write` as below.
CRC32s are generally never included in the length. The general concept is `<len(data), data, crc(data)>`, which is also found in the index plenty of times.

wal.go
```diff
@@ -77,6 +77,8 @@ const (
 // The table gets initialized with sync.Once but may still cause a race
 // with any other use of the crc32 package anywhere. Thus we initialize it
 // before.
+// XXX: Once should already do the right thing here if two callers
+//      construct the table at the same time, since it uses a lock?
```
fabxc: It's something apparently broken in the stdlib... the race detector was giving me shit about it.
This is the workaround.

chunks.go
juliusv: I think head makes most sense to me intuitively. When I'm thinking about head+tail, I'm thinking of the classic snake game, where the head is at the front and basically moving forwards, and old stuff at the tail gets cut off :)

fabxc: And yet you tail log files :P

chunks.go
juliusv: Hurdur, of course (about the n increment).

But if we use `maxLen` to determine whether we still have space in the current segment, but we also write the CRC32 to the segment buffer, shouldn't it be included in that size calculation? Otherwise we'd need to preallocate to `segmentSize + 4`? But I'm probably just jet-lagged and dense again :)

chunks.go
juliusv: DAMN ;) Well, I'll just pull the 1.x consistency card then :P

juliusv:
Ok, cool on the other points. The segment vs. block naming still confuses me, but I'll give it more thought when I'm awake again.

PR #71, opened 2017-05-01
Fix #59

### LINKED ISSUE #59: code bug in querier.go
```
func (s *populatedChunkSeries) Next() bool {
	for s.set.Next() {
		lset, chks := s.set.At()

		for i, c := range chks {
			if c.MaxTime < s.mint {
				chks = chks[1:]
				continue
			}
```
chks length is changed above, below code will not work 
as you want. It will panic for "slice bounds out of range".
I think it should be something  like:
```
             var out int
             for i, c := range chks {
			if c.MaxTime < s.mint {
				out++
				continue
			}
................
              chks=chks[out:]
               if len(chks) == 0 {
			continue
		}
```

Gouthamve:
Mutating the chunks can change their length. Hence referencing using
previous indices can cause panics.

cc @fabxc @qinguoan

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/71)

### DIFF
--- a/querier.go
+++ b/querier.go
@@ -400,7 +400,7 @@ func (s *populatedChunkSeries) Next() bool {
 	for s.set.Next() {
 		lset, chks := s.set.At()
 
-		from := 0
+		from := -1
 		for i, c := range chks {
 			if c.MaxTime < s.mint {
 				from = i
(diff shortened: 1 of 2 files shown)


fabxc:
👍

PR #72, opened 2017-05-01
Fix handling of duplicate datapoints

Gouthamve:
Currently if we allow appending two equal samples (both t, v). This
fixes the behaviour by not allowing two samples with the same ts.

Also we could previously append out of order or equal samples when in
the same transaction. This fixes it.

cc @fabxc @brian-brazil before we forget the discussion in #42

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/72)

### DIFF
--- a/head_test.go
+++ b/head_test.go
@@ -15,6 +15,7 @@ package tsdb
 
 import (
 	"io/ioutil"
+	"math"
 	"os"
 	"testing"
 	"unsafe"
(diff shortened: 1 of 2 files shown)


brian-brazil:
I think the unittests should stay, as they're testing NaN handling rather than just float handling.

Gouthamve:
Even though we are not checking the value at all?

brian-brazil:
We might change the code again in future, and the unittests will catch any breakage from that.

Gouthamve:
@brian-brazil Added it back. PTAL

head_test.go
```diff
@@ -97,44 +97,23 @@ func TestAmendDatapointCausesError(t *testing.T) {
...
+	require.NoError(t, err)
 
 	app = hb.Appender()
-	_, err = app.Add(labels.Labels{}, 0, 1)
-	if err != ErrAmendSample {
-		t.Fatalf("Expected error amending sample, got: %s", err)
-	}
-}
 
-func TestDuplicateNaNDatapointNoAmendError(t *testing.T) {
```
brian-brazil: This test should remain too, as it'll catch if we switch to float comparisons as NaN != NaN

head_test.go
```diff
@@ -148,16 +127,24 @@ func TestNonDuplicateNaNDatapointsCausesAmendError(t *testing.T) {
...
+	require.Equal(t, ErrAmendSample, err)
+
+	_, err = app.Add(labels.Labels{}, 1, math.NaN())
+	require.NoError(t, err)
+	err = app.Commit()
+	require.NoError(t, err)
+
+	app = hb.Appender()
+	_, err = app.Add(labels.Labels{}, 1, math.NaN())
+	require.Equal(t, ErrAmendSample, err)
```
Gouthamve: @brian-brazil I moved the NaN != NaN here. Should I put it in its own test?

brian-brazil: This shouldn't cause an amend error (it doesn't currently), so it belongs in a differently named test.

Gouthamve: Hmm, but all this is assuming that we will add float comparisons later. I am pretty sure that won't happen. @fabxc WDYT?

If we have two tests doing the same thing, it is bound to confuse those w/o context and any minor refactor will merge the two tests. Will add another if you are sure, though.

brian-brazil: Unittests are one way we ensure that doesn't happen.

fabxc:
> Currently if we allow appending two equal samples (both t, v)

This was a deliberate decision unless there was a discussion I missed. You can frequently run into this e.g. by federating metrics at a higher frequency than they are updated. So this erroring would be extremely noisy for no real value.

Gouthamve:
Aah, https://github.com/prometheus/tsdb/pull/42#issuecomment-293812868 this made me open this.

So instead of erroring out, we simply skip appending? That makes sense. 

But there is another bug here we could end up with out of order samples if we append that way in the same transaction. Will open a new PR the docs on the exact duplicates and the fix for out of order samples.

PR #73, opened 2017-05-02
Handle duplicate & out of order values in same txn

Gouthamve:
Add docs about not erroring out on exact dupes.
Moved tests to `require.*`

cc @fabxc @brian-brazil

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/73)

### DIFF
--- a/head_test.go
+++ b/head_test.go
@@ -145,7 +145,7 @@ func TestSkippingInvalidValuesInSameTxn(t *testing.T) {
 	hb, err := createHeadBlock(tmpdir+"/hb", 0, nil, 0, 1000)
 	require.NoError(t, err)
 
-	// Append AmendedValue
+	// Append AmendedValue.
 	app := hb.Appender()
 	_, err = app.Add(labels.Labels{{"a", "b"}}, 0, 1)
 	require.NoError(t, err)
@@ -157,17 +157,16 @@ func TestSkippingInvalidValuesInSameTxn(t *testing.T) {
 	// Make sure the right value is stored.
 	q := hb.Querier(0, 10)
 	ss := q.Select(labels.NewEqualMatcher("a", "b"))
-	require.True(t, ss.Next())
-	it := ss.At().Iterator()
-	require.True(t, it.Next())
-	ts, v := it.At()
-	require.Equal(t, int64(0), ts)
-	require.Equal(t, float64(1), v)
-	require.False(t, it.Next())
-	require.False(t, ss.Next())
+	ssMap, err := readSeriesSet(ss)
+	require.NoError(t, err)
+
+	require.Equal(t, map[string][]sample{
+		labels.New(labels.Label{"a", "b"}).String(): []sample{{0, 1}},
+	}, ssMap)
+
 	require.NoError(t, q.Close())
 
-	// Append Out of Order Value
+	// Append Out of Order Value.
 	app = hb.Appender()
 	_, err = app.Add(labels.Labels{{"a", "b"}}, 10, 3)
 	require.NoError(t, err)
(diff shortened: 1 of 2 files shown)


head_test.go
```diff
@@ -92,72 +92,103 @@ func TestAmendDatapointCausesError(t *testing.T) {
...
+	_, err = app.Add(labels.Labels{{"a", "b"}}, 0, 1)
+	require.NoError(t, err)
+	_, err = app.Add(labels.Labels{{"a", "b"}}, 0, 2)
+	require.NoError(t, err)
+	require.NoError(t, app.Commit())
+	require.Equal(t, uint64(1), hb.Meta().Stats.NumSamples)
+
+	// Make sure the right value is stored.
+	q := hb.Querier(0, 10)
+	ss := q.Select(labels.NewEqualMatcher("a", "b"))
```
brian-brazil: You can use `readSeriesSet` here to make this simpler

head_test.go
```diff
@@ -92,72 +92,103 @@ func TestAmendDatapointCausesError(t *testing.T) {
...
+	require.Equal(t, float64(1), v)
+	require.False(t, it.Next())
+	require.False(t, ss.Next())
+	require.NoError(t, q.Close())
+
+	// Append Out of Order Value
+	app = hb.Appender()
+	_, err = app.Add(labels.Labels{{"a", "b"}}, 10, 3)
+	require.NoError(t, err)
+	_, err = app.Add(labels.Labels{{"a", "b"}}, 7, 5)
```
brian-brazil: Shouldn't this be an error?

head_test.go
```diff
@@ -92,72 +92,103 @@ func TestAmendDatapointCausesError(t *testing.T) {
...
+	defer os.RemoveAll(tmpdir)
+
+	hb, err := createHeadBlock(tmpdir+"/hb", 0, nil, 0, 1000)
+	require.NoError(t, err)
+
+	// Append AmendedValue
+	app := hb.Appender()
+	_, err = app.Add(labels.Labels{{"a", "b"}}, 0, 1)
+	require.NoError(t, err)
+	_, err = app.Add(labels.Labels{{"a", "b"}}, 0, 2)
```
brian-brazil: Shouldn't this be an error?

head_test.go
```diff
@@ -92,72 +92,103 @@ func TestAmendDatapointCausesError(t *testing.T) {
...
+}
+
+func TestSkippingInvalidValuesInSameTxn(t *testing.T) {
+	tmpdir, _ := ioutil.TempDir("", "test")
+	defer os.RemoveAll(tmpdir)
+
+	hb, err := createHeadBlock(tmpdir+"/hb", 0, nil, 0, 1000)
+	require.NoError(t, err)
+
+	// Append AmendedValue
```
brian-brazil: commends should end with full stops

head_test.go
Gouthamve: This and the below mentioned error don't error out currently. Instead any wrong values are silently skipped during `Commit()`. Should they error out? 

If they should, then it adds some complexity to the code as we would need a `memSeries` for every series in the txn. cc @fabxc

head_test.go
Gouthamve: Yes indeed! Thanks! Pushed a new commit.

head_test.go
fabxc: Multiple samples for the same series in the same transaction are currently indeed not verified against each other. This would need same changes in how we batch up samples in the transaction object as you said.

For now it's fine to keep it, adding a test that we are not writing anything invalid is certainly good.

fabxc:
👍

PR #75, opened 2017-05-03
E2E test for headBlock

Gouthamve:
This adds the foundation for E2E test of `headBlock` with generated data. Will add more tests and variety if this is ok. @fabxc

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/75)

### DIFF
--- a/head_test.go
+++ b/head_test.go
@@ -226,7 +226,7 @@ func TestHeadBlock_e2e(t *testing.T) {
 
 	for _, l := range lbls {
 		ls := labels.New(l...)
-		series := seriesMap[labels.New(l...).String()]
+		series := []sample{}
 
 		ts := rand.Int63n(300)
 		for i := 0; i < numDatapoints; i++ {
@@ -268,21 +268,15 @@ func TestHeadBlock_e2e(t *testing.T) {
 	}
 
 	for _, qry := range queries {
-		matched := []labels.Labels{}
-	Outer:
+		matched := labels.Slice{}
 		for _, ls := range lbls {
-			for _, m := range qry.ms {
-				if !matchLSet(m, ls) {
-					continue Outer
-				}
+			s := labels.Selector(qry.ms)
+			if s.Matches(ls) {
+				matched = append(matched, ls)
 			}
-
-			matched = append(matched, ls)
 		}
 
-		sort.Slice(matched, func(i, j int) bool {
-			return labels.Compare(matched[i], matched[j]) < 0
-		})
+		sort.Sort(matched)
 
 		for i := 0; i < numRanges; i++ {
 			mint := rand.Int63n(300)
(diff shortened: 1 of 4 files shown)


head_test.go
```diff
@@ -161,3 +163,219 @@ func TestNonDuplicateNaNDatapointsCausesAmendError(t *testing.T) {
...
+
+				// HACK: Skip a series if iterator is empty.
+				if rok {
+					for !ss.At().Iterator().Next() {
+						rok = ss.Next()
+						if !rok {
+							break
+						}
+					}
+				}
```
fabxc: I think the behavior will basically stay as it is not harmful in principal.
Wouldn't it be easier to check sample equality before label equality below and then just `continue` if `len(smplRes) == 0`?

head_test.go
```diff
@@ -161,3 +163,219 @@ func TestNonDuplicateNaNDatapointsCausesAmendError(t *testing.T) {
...
+			break
+		}
+	}
+
+	if end == -1 {
+		end = len(full)
+	}
+
+	return full[start:end]
+}
```
fabxc: This is the culprit of randomized tests. This function extracts what the correct result data is but is so complex in itself, that it would need its own tests again. Many changes for off-by-one errors here and hard to verify by just looking at the code.

If we could make it simpler to the point where it is "obviously correct", that would be good.
Does this roughly implement the same behavior?

```go
// Truncate samples in the front that are before mint.
for len(full) > 0 {
    if full[0] >= mint {
        break
    }
    full = full[1:]
}
for i, s := range full {
    // Terminate on the first sample larger than maxt.
    if s.t > maxt {
        return full[:i]
    }
}
// maxt is after highest sample.
return full
```

querier.go
```diff
@@ -420,14 +420,16 @@ func (s *populatedChunkSeries) Next() bool {
...
+				chks = chks[:i]
 				break
 			}
 			c.Chunk, s.err = s.chunks.Chunk(c.Ref)
 			if s.err != nil {
 				return false
 			}
 		}
+
+		chks = chks[from+1:]
```
fabxc: Can we kind of apply the same simple double for loop to get rid of tracking variables?

head_test.go
Gouthamve: This was labelled a hack as it took me a long to debug why the tests were failing. Will remove the HACK label.

Now as per your suggestion, we would be unrolling weights and moving the set forward if `len(smplRes) == 0`, which is basically what I am doing above, no? We should also move only one set forward.

Gouthamve:
@fabxc Updated, PTAL.

head_test.go
Gouthamve: This is :heart: You have no clue how much time it took to get this bloody function right! Updated it now to use this.

fabxc:
LGTM

PR #77, opened 2017-05-11
Fix missing postings in Merge and Intersect

Gouthamve:
Before we were moving the postings list everytime we create a new
chained `intersectPostings`. That was causing some postings to be
skipped. This test fails on the older version.

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/77)

### DIFF
--- a/postings_test.go
+++ b/postings_test.go
@@ -78,23 +78,35 @@ func TestIntersect(t *testing.T) {
 
 func TestMultiIntersect(t *testing.T) {
 	var cases = []struct {
-		a, b, c []uint32
-		res     []uint32
+		p   [][]uint32
+		res []uint32
 	}{
 		{
-			a:   []uint32{1, 2, 3, 4, 5, 6, 1000, 1001},
-			b:   []uint32{2, 4, 5, 6, 7, 8, 999, 1001},
-			c:   []uint32{1, 2, 5, 6, 7, 8, 1001, 1200},
+			p: [][]uint32{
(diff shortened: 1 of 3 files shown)


Gouthamve:
This PR fixes https://github.com/prometheus/prometheus/issues/2616

@fabxc The tests pass in my local and travis failure is not related to this PR. This is ready for review.

@telendt I rolled back always advancing on Seek as it was causing issues, please take a look at the explanation here: https://github.com/prometheus/tsdb/pull/77/commits/101c751add08c0beb5ee3bdc10c47c81a1776ca3 Would be good to have your review.

postings_test.go
```diff
@@ -78,23 +78,35 @@ func TestIntersect(t *testing.T) {
...
-			c:   []uint32{1, 2, 5, 6, 7, 8, 1001, 1200},
+			p: [][]uint32{
+				{1, 2, 3, 4, 5, 6, 1000, 1001},
+				{2, 4, 5, 6, 7, 8, 999, 1001},
+				{1, 2, 5, 6, 7, 8, 1001, 1200},
+			},
 			res: []uint32{2, 5, 6, 1001},
 		},
+		{
+			p: [][]uint32{
```
fabxc: I suppose that's the test verifying the bug is gone? Probably good to add a comment here which edge case this seemingly trivial test is checking.

Gouthamve: Done

PR #79, opened 2017-05-13
Various type improvements

fabxc:
The `HeadBlock` interface was only used internally, which is not good in general. Interfaces should serve the purpose of composability. The implementation is now public and the interface unexported for internal usage.

For the same reason, the WAL is an interface now and the implementation was renamed to `SegmentWAL`. Head block tests still use the SegmentWAL, but could now inject WALs simulating write failures as well.

The `refdSample` is now `RefSample` as it shows up in an exported method. The WAL interface overall still has some warts, but this should be a fair bit better already.

@Gouthamve

### DIFF
--- a/db.go
+++ b/db.go
@@ -418,7 +418,7 @@ func (db *DB) reloadBlocks() error {
 
 		if meta.Compaction.Generation == 0 {
 			if !ok {
-				b, err = OpenHeadBlock(dirs[i], db.logger)
+				b, err = db.openHeadBlock(dirs[i])
 				if err != nil {
 					return errors.Wrapf(err, "load head at %s", dirs[i])
 				}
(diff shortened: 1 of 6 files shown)


db.go
```diff
@@ -709,16 +709,37 @@ func (db *DB) blocksForInterval(mint, maxt int64) []Block {
...
 
 	dir, seq, err := nextSequenceFile(db.dir, "b-")
 	if err != nil {
 		return nil, err
 	}
-	newHead, err := createHeadBlock(dir, seq, db.logger, mint, maxt)
+	if err := TouchHeadBlock(dir, seq, mint, maxt); err != nil {
+		return nil, errors.Wrapf(err, "touch head block %s", dir)
+	}
+	newHead, err := db.openHeadBlock(dir)
```
Gouthamve: Can we not `touch` inside `openHeadBlock`? Seems logical as it a part of the opening.

fabxc: Touch additionally has the sequence and time range arguments and creates the block dir with its meta.json atomically. We also use `openHeadBlock` on reloading, when we want to open block that already have been created with those fixed arguments.

Maybe `Touch` isn't the best name? But `Create*` generally suggests that it also returns the opened head block, which `Touch*` does not.

Gouthamve: Aah, makes sense! Also, I think `Touch` seems apt.

fabxc: Okay, will merge then :) Thanks

PR #82, opened 2017-05-16
Implement deletions

Gouthamve:
Todo:
- [x] Test the deletion
- [x] Put deletions on `db` (currently only on head and persisted blocks)
- [x] Handle deletions beyond current time.
- [x] Add locking to contend with compaction
- [x] Use WAL for headBlock
- [x] Revisit tombstone file format and add checksums
- [x] Optimise the tombstone lookup by using an Iterator instead of a map.
- [x] Add meta information: numTombstones.
- [x] Revisit all changed and new interfaces again.
  - [x] Rename trange to interval.
  - [x] Make []trange a new type named intervals.

Idea:
The tombstones live inside the index, whenever we look-up the series chunks, we insert the deleted ranges into the chunk.
When this chunk returns an iterator, the iterator returns values purely outside the deleted time-range.

We load the tombstones from a file, we when delete, we replace the file with the new tombstone file and have to execute `reloadBlocks()` to update an index's tombstones.

@fabxc

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/82)


head.go
```diff
@@ -192,11 +210,59 @@ func (h *HeadBlock) Meta() BlockMeta {
...
+
+	ir := h.Index()
+
+	pr := newPostingsReader(ir)
+	p, absent := pr.Select(ms...)
+
+	h.mtx.RUnlock()
+
+	h.mtx.Lock() // We are modifying the tombstones here.
+	defer h.mtx.Unlock()
```
fabxc: Switching from read to write lock is always a bit dangerous as stuff can happen in between. If we always have to the acquire the wlock anyway, it's typically not worth the headache. We can just acquire the write lock from the beginning.

head.go
```diff
@@ -192,11 +210,59 @@ func (h *HeadBlock) Meta() BlockMeta {
...
+			}
+		}
+
+		rs, ok := h.tombstones[ref]
+		if !ok {
+			h.tombstones[ref] = []trange{{mint, maxt}}
+			continue
+		}
+
+		h.tombstones[ref] = addNewInterval(rs, trange{mint, maxt})
```
fabxc: The way `addNewInterval` works, you should be able to simplify the above to

```go
h.tombstones[ref] = addNewInterval(h.tombstones[ref], trange{mint, maxt})
```

head.go
```diff
@@ -192,11 +210,59 @@ func (h *HeadBlock) Meta() BlockMeta {
...
+	}
+
+	if p.Err() != nil {
+		return p.Err()
+	}
+
+	return writeTombstoneFile(h.dir, newMapTombstoneReader(h.tombstones))
+}
+
+// Querier implements Queryable and headBlock
```
fabxc: All comments in their own line should end with a `.`

head.go
```diff
@@ -524,6 +595,9 @@ func (h *headIndexReader) Series(ref uint32) (labels.Labels, []*ChunkMeta, error
 			MinTime: c.minTime,
 			MaxTime: c.maxTime,
 			Ref:     (uint64(ref) << 32) | uint64(i),
+
+			deleted: deleted,
+			dranges: dranges,
```
fabxc: I see how this is getting tricky. Maybe it makes sense to not consider tombstones at this level. It's the low-level interface to access data we have. Given that the data is still there, just extended by tombstones, maybe it should remain queryable.

At higher level abstractions we can then combine that info with our known tombstones to provide the end result. We could just extend `IndexReader` by a `Tombstone(ref)` method. For the disk index this now means  it's multiple files. Or, maybe cleaner, make tombstones its own additional concept on the same level as chunks and index.

index.go
```diff
@@ -165,6 +165,10 @@ func newIndexWriter(dir string) (*indexWriter, error) {
 	if err := iw.writeMeta(); err != nil {
 		return nil, err
 	}
+	// TODO(gouthamve): Figure out where this function goes, index or block.
+	if err := writeTombstoneFile(dir, emptyTombstoneReader); err != nil {
+		return nil, err
+	}
```
fabxc: Ah, I see you were thinking about the same two options I mentioned above.

Gouthamve:
@fabxc So the only thing pending other than (a ton of) tests is the locking to ensure compaction doesn't mess with deletes. Instead of having a lock on each `Block`, I plan to wait for existing compactions to complete and stop compactions until Delete is done.

This functionality is anyways required for backups and deletes need not be very fast, so I think its the best way to go. WDYT?

fabxc:
Yes, exactly, we want an on/off switch for compactions anyway.

Gouthamve:
@fabxc So the changes since the last review:
1. Made tombstones their own thing like index and chunks.
2. Serialisation of Deletes and Compaction is done via a Mutex.
3. Add tests for the tombstone iterators and headblock deletions.

The pending changes are listed in first description. It'd be great if you gave it another review while I work on them. Also, I would like to rebase this on top of #83 once that is merged.

block.go
```diff
@@ -191,17 +215,68 @@ func (pb *persistedBlock) String() string {
...
+			}
+		}
+
+		for _, chk := range chunks {
+			if intervalOverlap(mint, maxt, chk.MinTime, chk.MaxTime) {
+				// Delete only until the current maxtime and not beyond.
+				maxtime := chunks[len(chunks)-1].MaxTime
+				if maxtime > maxt {
+					maxtime = maxt
+				}
```
fabxc: What's the argument for clamping the max time but not the min time?
Currently, the clamping is probably not necessary as we cannot modify the block – but probably safer to do it.

chunks.go
```diff
@@ -54,6 +54,46 @@ func (cm *ChunkMeta) writeHash(h hash.Hash) error {
 	return nil
 }
 
+// deletedIterator wraps an Iterator and makes sure any deleted metrics are not
+// returned.
+type deletedIterator struct {
+	it chunks.Iterator
+
+	intervals intervals
+}
```
fabxc: Is there a reason for applying the tombstones to each chunk iterator rather than wrapping a full series iterator? Dragging it this far down makes it more invasive.

compact.go
```diff
@@ -261,6 +262,15 @@ func (c *compactor) write(uid ulid.ULID, blocks ...Block) (err error) {
...
 	}
 
+	// Create an empty tombstones file.
+	tf, err := os.Create(filepath.Join(tmp, tombstoneFilename))
+	if err != nil {
+		return errors.Wrap(err, "touch tombstones file")
+	}
+	if err := tf.Close(); err != nil {
+		return errors.Wrap(err, "close tombstones file")
+	}
```
fabxc: Does this imply every block always has a tombstone file around?

chunks.go
fabxc: I see, we use this to re-encode chunks in compaction.

db.go
```diff
@@ -660,6 +666,31 @@ func (a *dbAppender) Rollback() error {
 	return g.Wait()
 }
 
+// Delete implements deletion of metrics.
+func (db *DB) Delete(mint, maxt int64, ms ...labels.Matcher) error {
+	db.cmtx.Lock()
+	defer db.cmtx.Unlock()
```
fabxc: As you said, another mutex is not exactly pretty. An alternative may be adding a function channel for coordinating block tasks.

Let's say there was `db.actorc` of type `chan func()`. The DB's backgronud routine would consume these functions in a for loop and execute them. That guarantees synchronization without the mutex. Additionally, if we decide to go the context route in #84  at some point, one can easily time out waiting for ones turns without having `Delete` block forever until it finally gets the mutex.

```go

func (db *DB) delete(mint, maxt int64, ms ...labels.Matcher) error {
    // actual delete logic.
}

func (db *DB) Delete(ctx context.Context, mint, maxt int64, ms ...labels.Matcher) (err error_ {
    select {
    case <-ctx.Done():
        err = ctx.Err()
    case db.actorc <- func() {
        err = db.delete(mint, maxt, ms...)
    }
    return err
}
```

db.go
```diff
@@ -660,6 +666,31 @@ func (a *dbAppender) Rollback() error {
...
+
+	db.headmtx.RLock()
```
fabxc: I think that's the wrong mutex for `db.blocks`.

db.go
```diff
@@ -660,6 +666,31 @@ func (a *dbAppender) Rollback() error {
...
+	blocks := db.blocksForInterval(mint, maxt)
+	db.headmtx.RUnlock()
+
+	var g errgroup.Group
+
+	for _, b := range blocks {
+		f := func() error {
+			return b.Delete(mint, maxt, ms...)
+		}
+		g.Go(f)
```
fabxc: No need to instantiate `f`. Just pass the function directly to `g.Go`. It's a matter of test in the end of course.

db.go
```diff
@@ -673,10 +704,8 @@ func (db *DB) appendable() (r []headBlock) {
 }
 
 func intervalOverlap(amin, amax, bmin, bmax int64) bool {
-	if bmin >= amin && bmin <= amax {
-		return true
-	}
-	if amin >= bmin && amin <= bmax {
+	// Checks Overlap: http://stackoverflow.com/questions/3269434/
+	if amin <= bmax && bmin <= amax {
```
fabxc: Neat!

fabxc: I guess we can just `return amin <= bmax && bmin <= amax` then.

block.go
Gouthamve: Yes, we cannot append before. But yeah, will do it.

compact.go
Gouthamve: Yes, an empty tombstone file atleast. I can handle no tombstone file to be an empty one, if you want.

db.go
Gouthamve: `g.Go` needs a function that accepts no parameters. It needs a `func() error`. Any other way to do it?

fabxc: Yea, I just meant you can just do

```go
g.Go(func() error { return b.Delete(mint, maxt, ms...) })
```

tombstones_test.go
```diff
@@ -0,0 +1,333 @@
...
+				require.True(t, trc.Next())
+				require.Equal(t, pr.At(), trc.At().ref, "refs")
+				require.Equal(t, dranges, trc.At().intervals)
+			}
+
+			require.False(t, trc.Next())
+			require.NoError(t, pr.Err())
+			require.NoError(t, tr.Err())
+		}
+		return
```
fabxc: Is this function different from `testFunc`?

db.go
Gouthamve: You sure: https://github.com/prometheus/tsdb/blob/master/querier.go#L62-L64 ? If yes, I will change there also.

But it also makes sense to use `headmtx` as blocks will change only when headblocks change?

fabxc: A compaction can also only change non-head blocks.
In the querier we hold the entire DB lock regardless, which we don't do here, right? Not 100% sure without looking properly again why we lock headmtx for the brief retrieval of `blocksForRange` – probably there was a reason somewhere

tombstones_test.go
Gouthamve: Yes, in testFunc we are testing against the map, but here we are testing against a static range.

db.go
Gouthamve: I will make this change in later PR and would prefer if this PR went in with the current implementation.

Documentation/format/tombstones.md
```diff
@@ -0,0 +1,55 @@
...
+│ ├──────────────────────────────────────────────┤ │
+│ │                      ...                     │ │
+│ ├──────────────────────────────────────────────┤ │
+│ │                   Ranges N                   │ │
+│ ├──────────────────────────────────────────────┤ │
+│ │                    Stones                    │ │
+│ ├──────────────────────────────────────────────┤ │
+│ │             Ref(stones start)<8b>            │ │
+│ └──────────────────────────────────────────────┘ │
+└──────────────────────────────────────────────────┘
```
fabxc: What is advantage of having a level of indirection from stone to ranges?
Right now (and probably for ever), we are scanning all stones and reading them into memory. In the general case, there's probably just a single deletion range per series (or a few). The total size of ranges for a stone are still tiny, so separating them away from the stone itself just adds complexity to the format and related code.

It seems fine here to just make the file a sequence of `<series_ref>, <#ranges>, <range1>, <range2>, ...`

The reason why we chose indirection elsewhere is usually to apply deduplication (symbol table in the index) or to co-locate frequently accessed, small items better by moving larger contents belonging to them elsewhere (series in the index vs the chunks). Here neither is really a concern.

Probably fine to just CRC all entries once at the end. It will generally be in the single digit MB range even when deleting 200k series or so. At that scale, it will likely be applied via compaction directly because we can save a lot of space by removing data.
If a tombstone file is small enough to leave it around, a CRC32 is definitely sufficient.

block.go
```diff
@@ -167,11 +180,18 @@ func newPersistedBlock(dir string) (*persistedBlock, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
 	pb := &persistedBlock{
 		dir:    dir,
 		meta:   *meta,
 		chunkr: cr,
 		indexr: ir,
+
```
fabxc: No need for newline.

block.go
```diff
@@ -150,6 +160,9 @@ type persistedBlock struct {
 
 	chunkr *chunkReader
 	indexr *indexReader
+
+	// For tombstones.
+	tombstones tombstoneReader
```
fabxc: Obvious comment. I think we can drop that.

tombstones_test.go
```diff
@@ -0,0 +1,110 @@
+package tsdb
```
fabxc: Missing license header

wal.go
```diff
@@ -141,13 +151,40 @@ func (w *SegmentWAL) Reader() WALReader {
...
 	}
+
+	if w.flushInterval <= 0 {
+		return w.Sync()
+	}
+	return nil
+}
+
+// LogDeletes write a batch of new deletes to the log.
+func (w *SegmentWAL) LogDeletes(tr tombstoneReader) error {
```
fabxc: Same. Exported methods must not expose unexported types.

Gouthamve:
@fabxc I think I covered all your feedback, could you give it another review?

Gouthamve:
@fabxc Per our discussion, I have manually tested this and merged it. Will open a PR for the deletes endpoint in Prometheus soon.

PR #83, opened 2017-05-17
Switch append refs to string

fabxc:
@Gouthamve I've to fix tests etc. – but as a first look whether this is sane.

@brian-brazil this kills bug on potential collisions across appends.

### DIFF
--- a/compact.go
+++ b/compact.go
@@ -20,6 +20,8 @@ import (
 	"path/filepath"
 	"time"
 
+	"go4.org/sort"
+
 	"github.com/coreos/etcd/pkg/fileutil"
 	"github.com/go-kit/kit/log"
 	"github.com/oklog/ulid"
@@ -34,10 +36,10 @@ type Compactor interface {
 	// Plan returns a set of non-overlapping directories that can
 	// be compacted concurrently.
 	// Results returned when compactions are in progress are undefined.
-	Plan(dir string) ([][]string, error)
+	Plan() ([][]string, error)
 
 	// Write persists a Block into a directory.
-	Write(dir string, b Block) error
+	Write(b Block) error
 
 	// Compact runs compaction against the provided directories. Must
 	// only be called concurrently with results of Plan().
@@ -46,6 +48,7 @@ type Compactor interface {
 
 // compactor implements the Compactor interface.
 type compactor struct {
+	dir     string
 	metrics *compactorMetrics
 	logger  log.Logger
 	opts    *compactorOptions
@@ -87,8 +90,9 @@ type compactorOptions struct {
 	maxBlockRange uint64
 }
 
-func newCompactor(r prometheus.Registerer, l log.Logger, opts *compactorOptions) *compactor {
+func newCompactor(dir string, r prometheus.Registerer, l log.Logger, opts *compactorOptions) *compactor {
 	return &compactor{
+		dir:     dir,
 		opts:    opts,
 		logger:  l,
 		metrics: newCompactorMetrics(r),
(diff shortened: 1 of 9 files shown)


compact.go
```diff
@@ -20,6 +20,8 @@ import (
 	"path/filepath"
 	"time"
 
+	"go4.org/sort"
```
Gouthamve: I think this is the wrong package that got imported.

compact.go
```diff
@@ -103,23 +107,31 @@ type compactionInfo struct {
...
+}
+
+func (c *compactor) Plan() ([][]string, error) {
+	dirs, err := blockDirs(c.dir)
 	if err != nil {
 		return nil, err
 	}
 
-	var bs []*BlockMeta
+	var bs []dirMeta
```
Gouthamve: nit: Can this be named `ds` or `dirs`?

compact.go
```diff
@@ -143,26 +155,22 @@ func (c *compactor) Plan(dir string) ([][]string, error) {
 	return nil, nil
 }
 
-func (c *compactor) match(bs []*BlockMeta) bool {
-	g := bs[0].Compaction.Generation
+func (c *compactor) match(bs []dirMeta) bool {
```
Gouthamve: nit: rename to `dir` or `dirs`?

fabxc:
Sequential block naming is now gone. Overall I think this brought by a few really nice simplifications in various code paths.
I also removed the `AppendableBlocks` option and everything now assumes 2. Practically, it's irrelevant as all appendable blocks are held in memory and the `MinBlockDuration` has therefore practically equivalent effect.

We should also no longer create a bunch of blocks if there's a large gap between timestamps (c.f. prometheus/prometheus#2725).

fabxc:
No doubt, debugging by looking at files became significantly harder. A tsdb CLI tool would ultimately be very helpful there.

compact.go
```diff
@@ -244,7 +266,7 @@ func (c *compactor) write(dir string, blocks ...Block) (err error) {
 	if err := renameFile(tmp, dir); err != nil {
 		return errors.Wrap(err, "rename block dir")
 	}
-	for _, b := range blocks[1:] {
+	for _, b := range blocks {
```
Gouthamve: Question: What was happening with `blocks[0]` before? Were we writing into it?

head.go
```diff
@@ -71,30 +74,30 @@ type HeadBlock struct {
...
-	tmp := dir + ".tmp"
-
-	if err := os.MkdirAll(tmp, 0777); err != nil {
-		return err
-	}
-
+func TouchHeadBlock(dir string, mint, maxt int64) (string, error) {
 	entropy := rand.New(rand.NewSource(time.Now().UnixNano()))
 
 	ulid, err := ulid.New(ulid.Now(), entropy)
```
Gouthamve: Question: Are ULID clashes possible?

head.go
```diff
@@ -268,74 +278,89 @@ func putHeadAppendBuffer(b []RefSample) {
...
+		labels: lset,
+	})
+	// First bit indicates its a series created in this transaction.
+	ref |= (1 << 63)
+
 	a.newHashes[hash] = ref
+	binary.BigEndian.PutUint64(refb, ref)
 
-	return ref, a.AddFast(ref, t, v)
+	return nullRef, a.AddFast(string(refb), t, v)
```
Gouthamve: Question: What is happening here? Are we using explicit `[]byte` for performance reasons? Can we not do a simple `Itoa`?

compact.go
fabxc: Before blocks [1,2,3,4] got compacted int [1,4] – so only 2 and 3 were deleted while 1 was overwriten with a block spanning a larger range.
As we just use ULIDs now, all blocks have to be removed.

head.go
fabxc: Not realistically, ULID are pretty much like UUIDs but they have the nice property of also encoding time in them.

Certainly no crash possibility in our case here of course as block creations are all synchronous.

head.go
fabxc: We just wrap the byte representation of the ID number into a string. For debugging, you can still hexprint that one but in general, we don't care about human representation of a number, which is what `strconv` does.
While both would work and `stroncv` is probably highly optimized, it's just not necessary and having all refs at a fixed width is probably better for memory alignment overall.

We could also have made the ref type a `[]byte` instead of a string – doesn't really matter.

fabxc: Having the actual int representation also allows us to reliably do stuff like the leading 1 bit for in-transaction series.

This should show it somewhat nicely: https://play.golang.org/p/yJ9u8ZoxeX

PR #85, opened 2017-05-26
Add and fix metrics

fabxc:
@Gouthamve 

`tsdb_active_appenders` is always zero and rarely 1. It probably only has an interesting value if there's a reasonable degree of contention.

### DIFF
--- a/db.go
+++ b/db.go
@@ -122,13 +122,42 @@ type DB struct {
 }
 
 type dbMetrics struct {
+	activeAppenders      prometheus.Gauge
+	loadedBlocks         prometheus.GaugeFunc
+	reloads              prometheus.Counter
+	reloadsFailed        prometheus.Counter
+	reloadDuration       prometheus.Summary
 	samplesAppended      prometheus.Counter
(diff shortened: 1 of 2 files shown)


db.go
```diff
@@ -122,13 +122,42 @@ type DB struct {
...
 
-func newDBMetrics(r prometheus.Registerer) *dbMetrics {
+func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
 	m := &dbMetrics{}
 
+	m.activeAppenders = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "tsdb_active_appenders",
+		Help: "Number of currently active appender transactions",
+	})
+	m.loadedBlocks = prometheus.NewGaugeFunc(prometheus.GaugeOpts{
```
Gouthamve: Do we also want a function for heads? Or is it safe to assume it is always 2?

fabxc: Fair to assume that's always too. If not it's a bug and we cannot really add metrics for all the places where bug X or Y could possibly show.

Issue #4, opened 2017-03-14
Implement backups


fabxc:
Our design generally allows easy backups. Necessary steps for this are:

1. Disable compactor temporarily
2. Create a backup directory that hard-links to all existing block files
3. Snapshot() method that allows compacting current in-memory blocks into fully persistent blocks (not strictly necessary but nicer than backing up write-ahead log segments)
4. Re-enable compactor so we don't ramp up memory during 5. – because all compaction actions replace old files rather than modifying, the inodes the hard links point to remain valid
5. An external tool, e.g. rsync, can copy the backup directory somewhere safely and delete the directory

To disable/enable the compactor and trigger snapshots we can simply rely on the using application to make the method accessible or provide an optional convenience HTTP server in the tsdb directly.

Gouthamve:
Closed via #88

PR #88, opened 2017-06-05
Snapshot support

Gouthamve:
For snapshotting HeadBlock:
Step 1: Lock HeadBlock against new appenders and let existing appenders finish.
Step 2: Write out the HeadBlock to a folder as a persisted block.

Now as we are blocking writes during snapshots, I am not sure how it will behave under heavy ingestion. WDYT @fabxc?

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/88)

### DIFF
--- a/block.go
+++ b/block.go
@@ -1,4 +1,5 @@
 // Copyright 2017 The Prometheus Authors
+
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -52,13 +53,13 @@ type DiskBlock interface {
 type Block interface {
 	DiskBlock
 	Queryable
+	Snapshottable
 }
 
 // headBlock is a regular block that can still be appended to.
 type headBlock interface {
 	Block
 	Appendable
-	Snapshottable
 }
 
 // Snapshottable defines an entity that can be backedup online.
@@ -276,6 +277,42 @@ Outer:
 
 	pb.meta.Stats.NumTombstones = uint64(len(pb.tombstones))
 	return writeMetaFile(pb.dir, &pb.meta)
+}
+
+func (pb *persistedBlock) Snapshot(dir string) error {
+	blockDir := filepath.Join(dir, pb.meta.ULID.String())
+	if err := os.MkdirAll(blockDir, 0777); err != nil {
+		return errors.Wrap(err, "create snapshot block dir")
+	}
+
+	chunksDir := chunkDir(blockDir)
+	if err := os.MkdirAll(chunksDir, 0777); err != nil {
+		return errors.Wrap(err, "create snapshot chunk dir")
+	}
+
+	// Hardlink meta, index and tombstones
+	filenames := []string{metaFilename, indexFilename, tombstoneFilename}
+	for _, fname := range filenames {
+		if err := os.Link(filepath.Join(pb.dir, fname), filepath.Join(blockDir, fname)); err != nil {
+			return errors.Wrapf(err, "create snapshot %s", fname)
+		}
+	}
+
+	// Hardlink the chunks
+	curChunkDir := chunkDir(pb.dir)
+	files, err := ioutil.ReadDir(curChunkDir)
+	if err != nil {
+		return errors.Wrap(err, "ReadDir the current chunk dir")
+	}
+
+	for _, f := range files {
+		err := os.Link(filepath.Join(curChunkDir, f.Name()), filepath.Join(chunksDir, f.Name()))
+		if err != nil {
+			return errors.Wrap(err, "hardlink a chunk")
+		}
+	}
+
+	return nil
 }
 
 func chunkDir(dir string) string { return filepath.Join(dir, "chunks") }
--- a/block.go
+++ b/block.go
@@ -58,6 +58,12 @@ type Block interface {
 type headBlock interface {
 	Block
 	Appendable
+	Snapshottable
+}
+
+// Snapshottable defines an entity that can be backedup online.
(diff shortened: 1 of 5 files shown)


fabxc:
Snapshots should still be reasonably quick. A few seconds I guess?
So some disruption in ingestion is fine there IMO. But we probably want to have a proper way for pending scrapes to timeout in that phase (c.f. #84) – not important initially though I think.

db.go
```diff
@@ -528,6 +528,34 @@ func (db *DB) Close() error {
...
+	return nil
+}
+
+// Snapshot writes the current headBlock snapshots to snapshots directory.
+func (db *DB) Snapshot(dir string) error {
+	db.headmtx.RLock()
+	heads := db.heads[:]
+	db.headmtx.RUnlock()
+
+	for _, h := range heads {
```
fabxc: So additionally to snapshotting head blocks into the directory. We should hard link all persisted blocks into the dir as well.

Did deletions on persisted blocks create a new directory or were they added to existing ones? If the latter, we either have to change that, or require or enforce that not deletions are done while snaphots are being shipped off.
Theoretically, compaction can be resumed right after Snapshot returns (an argument for putting the enable/disbale calls right into this method btw). So this is not guaranteed through `ctmx` alone.

db.go
```diff
@@ -528,6 +528,34 @@ func (db *DB) Close() error {
 	return merr.Err()
 }
 
+// DisableCompactions disables compactions.
+func (db *DB) DisableCompactions() error {
+	db.stopc <- struct{}{} // TODO: Can this block?
```
fabxc: I suppose you want to cancel compactions here? The problem is that this may cancel any background processing forever https://github.com/prometheus/tsdb/blob/master/db.go#L272.

I would say just leave it out for now. That may cause the call to wait for quite a bit if there are lots of compactions in progress. But that's rare. We can figure out a separate approach for interrupting ongoing compactions later.

There are a few concerns here. For example, do we want to terminate and cleanup in-progress compactions, or just prevent that a single compaction trigger to process multiple compactions. We can worry about them later.

Gouthamve:
Fixes #4 This has been tested on a smallish dataset in local. Will be testing and timing it on prombench now with a larger dataset. You can review it in the meantime.

Gouthamve:
It works well with larger datasets too. With `~10sec` (usually around 7s) for a headblock with 15mins of data and negligible amount of time for hardlinking persisted blocks.

Zipped the data, downloaded and verified by running an instance on local and the metrics line up. Will do some more monitoring on performance over longer intervals, but I think this is good to go.

block.go
```diff
@@ -272,6 +279,42 @@ Outer:
...
+		return errors.Wrap(err, "create snapshot chunk dir")
+	}
+
+	// Hardlink meta, index and tombstones
+	filenames := []string{metaFilename, indexFilename, tombstoneFilename}
+	for _, fname := range filenames {
+		if err := os.Link(filepath.Join(pb.dir, fname), filepath.Join(blockDir, fname)); err != nil {
+			return errors.Wrapf(err, "create snapshot %s", fname)
+		}
+	}
```
fabxc: Alternatively without intermediate variable:

```go
for _, fname := range []string{
	metaFilename,
	indexFilename,
	tombstoneFilename,
}{
	if err := os.Link(filepath.Join(pb.dir, fname), filepath.Join(blockDir, fname)); err != nil {
		return errors.Wrapf(err, "create snapshot %s", fname)
	}
}
```

db.go
```diff
@@ -528,6 +530,38 @@ func (db *DB) Close() error {
 	return merr.Err()
 }
 
+// ToggleCompactions toggles compactions and returns if compactions are on or not.
+func (db *DB) ToggleCompactions() bool {
```
fabxc: Mh, we should probably still leave those as separate `EnableCompactions`/`DisableCompactions`.

Especially as we have no way to check the current state, so we don't even know what a call of `Toggle` would get us into :) If we had a getter for the current state, practically, every call to Toggle should probably be preceded by that – so we are not really gaining anything.

db.go
```diff
@@ -528,6 +530,38 @@ func (db *DB) Close() error {
...
+func (db *DB) Snapshot(dir string) error {
+	db.mtx.Lock() // To block any appenders.
+	defer db.mtx.Unlock()
+
+	db.cmtx.Lock()
+	defer db.cmtx.Unlock()
+
+	blocks := db.blocks[:]
+	for _, b := range blocks {
+		db.logger.Log("msg", "compacting block", "block", b.Dir())
```
fabxc: Maybe "snapshotting block" so we can distinguish in our logs from regular compactions.

Gouthamve:
@fabxc Updated PTAL.

db.go
```diff
@@ -528,6 +530,47 @@ func (db *DB) Close() error {
 	return merr.Err()
 }
 
+// DisableCompactions disables compactions.
+func (db *DB) DisableCompactions() error {
```
fabxc: As long as there's no possible error and it's not required by an interface, no need to return an error.

That said, we'll need to have a way for these operations to time out while trying to acquiring a lock (via passing a context) or have them enforce quick acquisition of locks, which is probably not possible due to long running queries.
Again somewhat circles back to #84 – we can look at this down the road.

Gouthamve: Hmm, so I remove the error for now? If we add contexts and we timeout, we would need to return an error then right?

fabxc:
Yes, but then we have to change the method signature again anyway :)

db.go
Gouthamve: Done :)

Issue #87, opened 2017-06-05
Incorrect metadata being populated on deletions


Gouthamve:
The numSeries and numSamples values after a block with tombstones is compacted are wrong.

While I have the fix ready for numSeries, [numSamples](https://github.com/prometheus/tsdb/blob/master/compact.go#L177-L179) is a little tricky. When we are [re-encoding a chunk](https://github.com/prometheus/tsdb/blob/master/compact.go#L329-L341) we donot know the number of samples the old chunk holds.

cc @fabxc

fabxc:
First, maybe we should make the beginning of a chunk always a varint of the
number of samples. The XOR encoding does this anyway. We should probably
just drag it out and make it part of the chunk file format. With that, it's
a lot easier to re-count samples on every compaction without decompressing
everything.

Second, we can just drop the stat for now if you feel that's better than
the inconsistency. It's not really being used right now.

Gouthamve:
Closed in #108

PR #91, opened 2017-06-06
WIP: Recompact blocks with large number of tombstones

Gouthamve:
This is just an RFC.

* We ideally want to recompact old blocks if there are too many tombstones there. But what are old blocks?
* For fixing #87 `NumSamples` is being read from the initial portion of the XOR block. I know it sucks, but can we not make it a part of the chunk format by letting it be only 16 bits instead of a varint as suggested?

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/91)

### DIFF
--- a/compact.go
+++ b/compact.go
@@ -14,6 +14,7 @@
 package tsdb
 
 import (
+	"encoding/binary"
 	"fmt"
 	"math/rand"
 	"os"
@@ -190,9 +191,6 @@ func mergeBlockMetas(blocks ...Block) (res BlockMeta) {
 
 	res.Compaction.Generation = m0.Compaction.Generation + 1
 
-	for _, b := range blocks {
-		res.Stats.NumSamples += b.Meta().Stats.NumSamples
-	}
 	return res
 }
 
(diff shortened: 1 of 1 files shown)


compact.go
```diff
@@ -145,6 +146,22 @@ func (c *compactor) Plan() ([][]string, error) {
...
+				compactDirs = append(compactDirs, []string{dm.dir})
+			}
+			continue
+		}
+		break
+	}
+
+	if len(compactDirs) > 0 {
+		return compactDirs, nil
+	}
```
fabxc: Probably better to place this after the regular compaction planning since we might already apply tombstones as a side effect of proper compactions.

compact.go
```diff
@@ -145,6 +146,22 @@ func (c *compactor) Plan() ([][]string, error) {
...
+	// Check if any "large" blocks need to be recompacted due to deletions.
+	var compactDirs [][]string
+	for _, dm := range dms {
+		if dm.meta.MaxTime-dm.meta.MinTime > int64(c.opts.maxBlockRange/2) {
+			if dm.meta.Stats.NumTombstones > 0 &&
+				(dm.meta.Stats.NumSeries/dm.meta.Stats.NumTombstones) <= 20 { // 5%
+				compactDirs = append(compactDirs, []string{dm.dir})
+			}
+			continue
+		}
```
fabxc: Better to reduce levels of indentation by inverting the first if and putting the `break` inside of it.

compact.go
```diff
@@ -347,22 +361,29 @@ func (c *compactor) populate(blocks []Block, indexw IndexWriter, chunkw ChunkWri
 			return nil, err
 		}
 
-		indexw.AddSeries(i, lset, chks...)
+		// Add a series only if the chunks exist.
+		if len(chks) > 0 {
```
fabxc: Can we again reduce the levels of indentation by flipping the condition?

PR #92, opened 2017-06-07
Add list of source head blocks to meta.json

### LINKED ISSUE #89: List ancestor blocks in bock meta
Currently we track "how old" a block is roughly by the "level" field in the meta.json file. It is generally incremented on every compaction. With compactions happening more frequently triggered by other mechanisms, e.g. for deletes, that field becomes relatively ambiguous/meaningless.

I propose to instead have `compaction.ancestors` field, which contains a list of all block that are contained within the block. On compaction of multiple blocks, those lists are simply merged.
They can obviously get rather long (168 entries for 1 week at 1h min block size), but not beyond reasonable.

It could just be a list of ULIDs. But it's probably a good idea to make them objects, so we can add other fields in the future.

This gives us generally a better idea of how much compaction was done in grand total. It is also really helpful, when reconciling backed up data.

fabxc:
Closes #89

### DIFF
--- a/compact.go
+++ b/compact.go
@@ -166,17 +166,35 @@ func (c *compactor) match(dirs []dirMeta) bool {
 	return uint64(dirs[len(dirs)-1].meta.MaxTime-dirs[0].meta.MinTime) <= c.opts.maxBlockRange
 }
 
-func mergeBlockMetas(blocks ...Block) (res BlockMeta) {
-	m0 := blocks[0].Meta()
+func compactBlockMetas(blocks ...BlockMeta) (res BlockMeta) {
+	res.MinTime = blocks[0].MinTime
+	res.MaxTime = blocks[len(blocks)-1].MaxTime
 
-	res.MinTime = m0.MinTime
-	res.MaxTime = blocks[len(blocks)-1].Meta().MaxTime
-
-	res.Compaction.Generation = m0.Compaction.Generation + 1
+	sources := map[ulid.ULID]struct{}{}
 
 	for _, b := range blocks {
-		res.Stats.NumSamples += b.Meta().Stats.NumSamples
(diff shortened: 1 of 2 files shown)


compact.go
```diff
@@ -166,17 +168,35 @@ func (c *compactor) match(dirs []dirMeta) bool {
...
-	m0 := blocks[0].Meta()
-
-	res.MinTime = m0.MinTime
-	res.MaxTime = blocks[len(blocks)-1].Meta().MaxTime
+func compactBlockMetas(blocks ...BlockMeta) (res BlockMeta) {
+	res.MinTime = blocks[0].MinTime
+	res.MaxTime = blocks[len(blocks)-1].MaxTime
 
-	res.Compaction.Generation = m0.Compaction.Generation + 1
+	sources := map[ulid.ULID]struct{}{}
```
Gouthamve: Arguably, I don't think there will exist 2 blocks with the overlapping sources. Is this here for the case of vertical merging?

compact.go
```diff
@@ -166,17 +168,35 @@ func (c *compactor) match(dirs []dirMeta) bool {
...
+			sources[b.ULID] = struct{}{}
+		}
+	}
+	res.Compaction.Generation++
+
+	for s := range sources {
+		res.Compaction.Sources = append(res.Compaction.Sources, s)
 	}
+	sort.Slice(res.Compaction.Sources, func(i, j int) bool {
+		return bytes.Compare(res.Compaction.Sources[i][:], res.Compaction.Sources[j][:]) < 0
```
Gouthamve: I think we are better off using this function: https://godoc.org/github.com/oklog/ulid#ULID.Compare

fabxc: Oh, neat.

fabxc:
PTAL

PR #94, opened 2017-06-07
Improve heuristic to spread chunks across block

fabxc:
This yields ideally balanced chunks within a block, given that the incoming data has regular timestamps. Any more sophisticated handling seems pointless given that the most frequent cause for unbalanced chunks with this change will be series churn, which we cannot predict.

@Gouthamve

### DIFF
--- a/head.go
+++ b/head.go
@@ -719,12 +719,7 @@ func (h *HeadBlock) get(hash uint64, lset labels.Labels) *memSeries {
 }
 
 func (h *HeadBlock) create(hash uint64, lset labels.Labels) *memSeries {
-	s := &memSeries{
-		lset: lset,
-		ref:  uint32(len(h.series)),
-	}
-	// create the initial chunk and appender
-	s.cut()
+	s := newMemSeries(lset, uint32(len(h.series)), h.meta.MaxTime)
 
 	// Allocate empty space until we can insert at the given index.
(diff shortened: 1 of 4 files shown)


fabxc:
This adds 16 bytes of additional per series overhead. Seems fine to me though.

head.go
```diff
@@ -812,6 +825,17 @@ func (s *memSeries) append(t int64, v float64) bool {
 	return true
 }
 
+// computeEndTime estimates the end timestamp based the beginning of a chunk,
+// its current timestamp and the upper bound up to which we insert data.
+// It assumes that the time range is 1/4 full.
+func computeChunkEndTime(start, cur, max int64) int64 {
+	a := (max - start) / (cur - start + 1) / 4
```
Gouthamve: Umm, maybe its just me, but can we have it as:
```
a := (max - start) / ((cur - start + 1) * 4)
```

Left me a little confused and made me resort to play.golang.org

head.go
```diff
@@ -812,6 +825,17 @@ func (s *memSeries) append(t int64, v float64) bool {
 	return true
 }
 
+// computeEndTime estimates the end timestamp based the beginning of a chunk,
```
Gouthamve: typo, `computeChunkEndTime`

PR #95, opened 2017-06-08
Fix race condition for 2 appenders having same ts

Gouthamve:
Ref: prometheus/promtheus#2795 

Not sure if this fixes the above issue, but it will fix the race condition mentioned.

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/95)

### DIFF
--- a/head.go
+++ b/head.go
@@ -529,8 +529,8 @@ func (a *headAppender) createSeries() error {
 	base0 := len(a.series)
 
 	a.mtx.RUnlock()
-	a.mtx.Lock()
 	defer a.mtx.RLock()
+	a.mtx.Lock()
 	defer a.mtx.Unlock()
 
 	base1 := len(a.series)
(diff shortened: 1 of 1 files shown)


head.go
```diff
@@ -521,15 +521,17 @@ func (a *headAppender) AddFast(ref string, t int64, v float64) error {
...
 	if len(a.newSeries) == 0 {
-		return
+		return nil
 	}
 	a.newLabels = make([]labels.Labels, 0, len(a.newSeries))
 	base0 := len(a.series)
 
 	a.mtx.RUnlock()
 	a.mtx.Lock()
+	defer a.mtx.RLock()
```
fabxc: Can we swap the above two lines for clarity?

fabxc:
Practically this might end up being necessary. Practically, this should be a very unlikely race. We really don't suffer much from lock contention but tsdb's bench tool shows it's very significant when we create a series in bulk, e.g. when a new block starts.  

Probably best to not worry about it yet though as actual Prometheus benchmarks haven't revealed a problem yet.

Gouthamve:
Done

PR #96, opened 2017-06-11
Fix directory file descriptor leak.

fabxc:
@Gouthamve 

I was under the impression leaking FDs would be released once the objects get cleaned up. Maybe my case was just too fast for that to actually happen. Regardless, this should be fixed obviously.

### DIFF
--- a/compact.go
+++ b/compact.go
@@ -300,6 +300,8 @@ func (c *compactor) write(uid ulid.ULID, blocks ...Block) (err error) {
 	if err != nil {
 		return errors.Wrap(err, "sync block dir")
 	}
+	defer df.Close()
+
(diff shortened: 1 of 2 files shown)


Gouthamve:
Hmm, LGTM, but do you think there can be some extra open directories which we need to close after this: https://github.com/prometheus/tsdb/blob/master/db.go#L478-L484

fabxc:
What exactly do you mean? The linked blocks are added to the closers slice and all those blocks get closed as the function returns. Also the blocks do not holden dir file descriptors afaics.

Gouthamve:
Sorry nvm, got confused a little. This is GTG.

PR #97, opened 2017-06-13
Fix bug with Seek() and optimise bounding params.

Gouthamve:
Started out as a minor optimisation but revealed a bug :smile:

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/97)

### DIFF
--- a/querier.go
+++ b/querier.go
@@ -657,10 +657,6 @@ func newChunkSeriesIterator(cs []*ChunkMeta, dranges intervals, mint, maxt int64
 	}
 }
 
-func (it *chunkSeriesIterator) inBounds(t int64) bool {
-	return t >= it.mint && t <= it.maxt
-}
-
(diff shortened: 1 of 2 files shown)


fabxc:
Can you add a regression test?

querier.go
```diff
@@ -708,9 +706,15 @@ func (it *chunkSeriesIterator) At() (t int64, v float64) {
 func (it *chunkSeriesIterator) Next() bool {
 	for it.cur.Next() {
 		t, _ := it.cur.At()
-		if it.inBounds(t) {
-			return true
+		if t < it.mint {
+			return it.Seek(it.mint)
+		}
```
fabxc: Was not doing a seek here before potentially a second bug?

Gouthamve: No it wasn't, it was just inefficient.

Gouthamve:
Added regression test. PTAL.

PR #101, opened 2017-06-30
Fix omitting of chunk on Seek()

fabxc:
We just get rid of the whole binary search complexity. The general use case (and only really allowed usage) is to stream forward through a series. In general, we will touch every chunk in the course of a query. So just iterating through them instead of binary searching is not only a lot simpler but also cheaper.

@Gouthamve

### DIFF
--- a/querier.go
+++ b/querier.go
@@ -15,7 +15,6 @@ package tsdb
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


Gouthamve:
LGTM

Issue #102, opened 2017-07-05
Fix block ranges


fabxc:
Currently block ranges allow fine-grained configuration. Practically, this has little to no benefit for Prometheus. It adds a lot of complexity however when dealing with those blocks in other ways, for example backup procedures.

I propose switching to a time slicing approach where possible block ranges after different compaction levels are fixed, as well as their position in time. This prevents arbitrary overlapping and interleaving of blocks at a global level.

Suppose we want a maximum block size of 9 days, then our valid ranges are as follows:
* 9d
* 3d
* 24h
* 6h
* 2h

The largest range is aligned with the 0 timestamp (epoch in Prometheus), smaller ones are always aligned with their parent range. Thus compaction will always result in larger blocks fitting the slices.

This can still be configurable in TSDB, especially for tests, but should be hard-coded in the using application.
For Prometheus I would fix it to the above values.

Alignment with wall clock weeks would have been nice but not all that important. TSDB having no notion of wall clocks and Unix 0 not aligning with the beginning of a week would just require a bunch of number shifting that just doesn't seem worth it.

@Gouthamve @brian-brazil

brian-brazil:
I'm okay with this change. The fewer configurables there are, the fwer red herrings there are for users to get caught up on.

Gouthamve:
Hmm, one pitfall that needs to be handled is that if we have a retention of 10days, we might end up with 18days worth of data before deleting things.

While we could adjust the things according to the retention, we need to think about how to handle it if someone shuts down and restarts Prometheus with a new retention.

Gouthamve:
Closed in #103

PR #103, opened 2017-07-06
Compaction implementation for block-ranges

Gouthamve:
Super early implementation:

So if you have `20, 60, 240, 720, 2160` as your ranges, you pick the oldest `2160` block you can compact and inside that, you pick the oldest `720` block and so on.

TODO:
- [x] More testing.
- [x] Simplification and more comments.
- [x] Optimise some parts.
- [x] Choose the ranges by retention.
- [x] Align ranges by parent instead of 0 aligning all ranges.
- [x] Cut blocks properly.

/cc @fabxc

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/103)

### DIFF
--- a/compact.go
+++ b/compact.go
@@ -30,6 +30,18 @@ import (
 	"github.com/prometheus/tsdb/labels"
 )
 
+// ExponentialBlockRanges returns the time ranges based on the stepSize
+func ExponentialBlockRanges(minSize int64, steps, stepSize int) []int64 {
+	ranges := make([]int64, 0, steps)
+	curRange := minSize
+	for i := 0; i < steps; i++ {
+		ranges = append(ranges, curRange)
+		curRange = curRange * int64(stepSize)
+	}
+
+	return ranges
+}
+
 // Compactor provides compaction against an underlying storage
 // of time series data.
 type Compactor interface {
@@ -91,15 +103,6 @@ type compactorOptions struct {
 }
 
 func newCompactor(dir string, r prometheus.Registerer, l log.Logger, opts *compactorOptions) *compactor {
-	if opts.blockRanges == nil {
-		opts.blockRanges = []int64{
-			int64(2 * time.Hour),
-			int64(6 * time.Hour),
-			int64(24 * time.Hour),
-			int64(72 * time.Hour),  // 3d
-			int64(216 * time.Hour), // 9d
-		}
-	}
 	return &compactor{
 		dir:     dir,
 		opts:    opts,
@@ -161,6 +164,11 @@ func (c *compactor) Plan() ([][]string, error) {
 }
 
 func (c *compactor) selectDirs(ds []dirMeta) []dirMeta {
+	// The way to skip compaction is to not have blockRanges.
+	if len(c.opts.blockRanges) == 1 {
+		return nil
+	}
+
 	return selectRecurse(ds, c.opts.blockRanges)
 }
 
@@ -184,6 +192,8 @@ func selectRecurse(dms []dirMeta, intervals []int64) []dirMeta {
 	// If there are too many blocks, see if a smaller interval will catch them.
 	// i.e, if we have 0-20, 60-80, 80-100; all fall under 0-240, but we'd rather compact 60-100
 	// than all at once.
+	// Again if have 0-1d, 1d-2d, 3-6d we compact 0-1d, 1d-2d to compact it into the 0-3d block instead of compacting all three
+	// This is to honor the boundaries as much as possible.
 	if len(dirs) > 2 {
 		smallerDirs := selectRecurse(dirs, intervals[:len(intervals)-1])
 		if len(smallerDirs) > 1 {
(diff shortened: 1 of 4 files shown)


Gouthamve:
Fixes #102

Gouthamve:
I just pushed a commit which some very strong defaults:

1. The min-block-size is 2h or lower not more than that.
2. We were thinking about doing 0 aligning of the largest block and every other block aligned to the parent. By making sure that every block-dur cleanly divides the parent, we can 0 align all the blocks and still be safe.
3. We will choose the max block duration so that we don't store data 20% older than retention.


The repercussions for these are the min block duration is 2h or something that divides 2h cleanly which is kinda weird but seems okay.

/cc @fabxc @brian-brazil

compact.go
```diff
@@ -87,10 +95,13 @@ func newCompactorMetrics(r prometheus.Registerer) *compactorMetrics {
...
 
 type compactorOptions struct {
-	maxBlockRange uint64
+	blockRanges []int64
 }
 
 func newCompactor(dir string, r prometheus.Registerer, l log.Logger, opts *compactorOptions) *compactor {
+	if opts.blockRanges == nil {
+		opts.blockRanges = defaultBlockRanges
+	}
```
fabxc: Do we need sanity checks on startup that the ranges actually fit within each other?
What happens if the chosen ranges do not work with already existing blocks?

db.go
```diff
@@ -60,11 +59,9 @@ type Options struct {
 
 	// The timestamp range of head blocks after which they get persisted.
 	// It's the minimum duration of any persisted block.
+	// It should divide 2h cleanly and should be less than or equal to 2h. Default 2h.
 	MinBlockDuration uint64
```
fabxc: As per above suggestion, we probably want to pass in the full ranges as config options instead of this value. The first range is than equivalent to `MinBlockDuration`. As said there, we also shouldn't make assumptions about stuff like 2h.

At least for bench/debugging purposes it probably would be a good idea to still allow passing in short initial ranges via Prometheus. Maybe a hidden `--debug.storage.tsdb...` flag?

db.go
```diff
@@ -227,9 +232,25 @@ func Open(dir string, l log.Logger, r prometheus.Registerer, opts *Options) (db
...
+	for {
+		if len(copts.blockRanges) == 1 {
+			break
+		}
+
+		// Max overflow is restricted to 20%.
+		if float64(copts.blockRanges[len(copts.blockRanges)-1])/float64(opts.RetentionDuration) > 0.2 {
+			copts.blockRanges = copts.blockRanges[:len(copts.blockRanges)-1]
+		}
+	}
```
fabxc: This would just spin if the highest range is less than 20% of the retention, no?

compact.go
Gouthamve: > What happens if the chosen ranges do not work with already existing blocks?

Some blocks which cross the boundaries of the max-block-range would never be compacted. Which is okay, I guess.

db.go
```diff
@@ -227,9 +222,24 @@ func Open(dir string, l log.Logger, r prometheus.Registerer, opts *Options) (db
...
-	})
+	copts := &compactorOptions{
+		blockRanges: opts.BlockRanges,
+	}
+
+	if len(copts.blockRanges) == 0 {
+		return nil, errors.New("atleast one block-range should exist.")
+	}
+
+	for float64(copts.blockRanges[len(copts.blockRanges)-1])/float64(opts.RetentionDuration) > 0.2 {
```
Gouthamve: @fabxc Given we allow passing of full ranges, should we allow for this?

fabxc: Yea, I think so. Imagine we generally have a hard-coded set of ranges. The maximum one we actually compact to should still be ceiled by the dynamically configured retention.

compact.go
```diff
@@ -133,37 +145,89 @@ func (c *compactor) Plan() ([][]string, error) {
...
+			return smallerDirs
+		}
+	}
+
+	return dirs
 }
 
-func (c *compactor) match(dirs []dirMeta) bool {
-	g := dirs[0].meta.Compaction.Generation
+func splitByRange(ds []dirMeta, tr int64) [][]dirMeta {
```
fabxc: This function definitely needs heading and inlined docs.

compact.go
```diff
@@ -133,37 +145,89 @@ func (c *compactor) Plan() ([][]string, error) {
...
+		}
+
+		if dir.meta.MinTime >= t0+tr {
+			splitDirs = append(splitDirs, dirs)
+			dirs = []dirMeta{}
+			t0 = dir.meta.MinTime - dir.meta.MinTime%tr
+			if intervalContains(t0, t0+tr, dir.meta.MinTime) && intervalContains(t0, t0+tr, dir.meta.MaxTime) {
+				dirs = append(dirs, dir)
+			}
 		}
```
fabxc: Repeating initialization and the  condition at different indentation levels makes is hard to grasp.

```go
var splitDirs [][]dirMeta

for i := 0; i < len(ds); {
    var group []dirMeta
    // Compute start of aligned time range of size tr closest to the current block's start.
    t0 := ds[i].meta.MinTime - (ds[i].meta.MinTime % tr)

    // Add all dirs to the current group that are within [t0, t0+tr].
    for ;i < len(ds); i++ {
        if ds[i].meta.MinTime < t0 || ds[i].meta.MaxTime > t0+tr {
            break
        }
        group = append(group, ds[i])
    }

    if len(group) > 0 {
        splitDirs = append(splitDirs, group)
    }
}

return splitDirs
```

Haven't verified this one, but should roughly do the same.

compact.go
```diff
@@ -133,37 +145,89 @@ func (c *compactor) Plan() ([][]string, error) {
...
+	if len(intervals) == 0 {
+		return dms
+	}
+
+	// Get the blocks by the max interval
+	blocks := splitByRange(dms, intervals[len(intervals)-1])
+	dirs := []dirMeta{}
+	for i := len(blocks) - 1; i >= 0; i-- {
+		// We need to choose the oldest blocks to compact. If there are a couple of blocks in
+		// the largest interval, we should compact those first.
```
fabxc: We are reverse-iterating. Doesn't that make us chose the youngest group of compactable blocks?

db.go
```diff
@@ -227,9 +222,24 @@ func Open(dir string, l log.Logger, r prometheus.Registerer, opts *Options) (db
...
 
-	db.compactor = newCompactor(dir, r, l, &compactorOptions{
-		maxBlockRange: opts.MaxBlockDuration,
-	})
+	copts := &compactorOptions{
+		blockRanges: opts.BlockRanges,
+	}
+
+	if len(copts.blockRanges) == 0 {
+		return nil, errors.New("atleast one block-range should exist.")
```
fabxc: at least

must exist

No `.` in errors.

compact.go
```diff
@@ -133,37 +145,89 @@ func (c *compactor) Plan() ([][]string, error) {
...
+			dirs = blocks[i]
+			break
 		}
 	}
 
-	return nil, nil
+	// If there are too many blocks, see if a smaller interval will catch them.
+	// i.e, if we have 0-20, 60-80, 80-100; all fall under 0-240, but we'd rather compact 60-100
+	// than all at once.
+	if len(dirs) > 2 {
```
fabxc: Before we were always compacting groups of 3. Why are we now searching in smaller ranges if we have 3?

compact.go
Gouthamve: Nope. So we actually sort the dirs in ascending order here: https://github.com/Gouthamve/tsdb/blob/block-ranges/compact.go#L144-L146 hence it makes sense to reverse-iterate.

Gouthamve:
@fabxc Incorporated all the changes. PTAL.

fabxc:
Travis is failing.

fabxc:
👍 if manual testing also confirmed this. Probably good to spin up prombench and let it run for a bit after the changes have been pulled there.

We might want to introduce debug flags to manually set the range so we can run tests with frequent compactions.
Or we even keep min block duration around for high frequency cases... they probably still exist and will run out of memory otherwise.

PR #105, opened 2017-07-07
Isolation

brian-brazil:
The core part of https://github.com/prometheus/prometheus/issues/1893

Initial prombench benchmarks indicate a ~5% cpu hit and ~11% RAM hit, which I consider reasonable for this feature. I haven't done any micro-optimisations.

There's some cleanup work to be done here still, but I figured it's best to get some initial review.

### DIFF
--- a/block.go
+++ b/block.go
@@ -90,6 +90,8 @@ type IsolationState struct {
 	// We will ignore all writes above the max, and that are incomplete.
 	maxWriteId       uint64
 	incompleteWrites map[uint64]struct{}
+	readId           uint64
+	db               *DB
 }
 
 // BlockMeta provides meta information about a block.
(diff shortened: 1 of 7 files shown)


mdlayher:
Appears this now has merge conflicts. Mind rebasing?

fabxc:
I'm currently doing some work to make the hot path of creating new series lock free. We are currently seeing rather heavy contention whenever new series are created (first startup, targets added/swapped, new head block created), which causes some notable drops in ingestion for larger servers.

This basically requires isolation as well, so I currently have some pending work doing the same thing. I'm trying to avoid leaking this into our nicely simple APIs.  

Rebasing might be a bit of unnecessary work right now.

brian-brazil:
Superseded by #306

PR #106, opened 2017-07-12
Persist the right MaxTime when snapshotting

Gouthamve:
This is because we cut a new block from where the snapshotted block ends
if we restore from backups and highTimestamp would be where we should be
 starting from.


fabxc:
👍

PR #108, opened 2017-07-12
Recompact blocks with large number of tombstones

Gouthamve:
Carried over #91 as it diverged too much from master.

### DIFF
--- a/chunks/chunk.go
+++ b/chunks/chunk.go
@@ -43,6 +43,7 @@ type Chunk interface {
 	Encoding() Encoding
 	Appender() (Appender, error)
 	Iterator() Iterator
+	NumSamples() int
 }
 
 // FromData returns a chunk from a byte slice of chunk data.
--- a/compact.go
+++ b/compact.go
@@ -14,7 +14,6 @@
 package tsdb
 
 import (
-	"encoding/binary"
 	"fmt"
 	"math/rand"
 	"os"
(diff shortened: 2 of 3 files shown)


compact.go
```diff
@@ -439,6 +455,9 @@ func populateBlock(blocks []Block, indexw IndexWriter, chunkw ChunkWriter) (*Blo
 
 		meta.Stats.NumChunks += uint64(len(chks))
 		meta.Stats.NumSeries++
+		for _, chk := range chks {
+			meta.Stats.NumSamples += uint64(binary.BigEndian.Uint16(chk.Chunk.Bytes()))
```
fabxc: We should eventually make this a first class property of the chunk interface. Strictly speaking not all compression formats would need this. But enforcing it also has negligible cost. In doubt one could have the `Size()` function just iterate through all elements.

fabxc: Actually, maybe we should just do it as part of this PR.

compact.go
```diff
@@ -160,7 +161,19 @@ func (c *compactor) Plan() ([][]string, error) {
...
+	if len(dirs) > 1 {
+		return dirs, nil
+	}
+
+	// Compact any blocks that have >5% tombstones.
+	for i := len(dms) - 1; i >= 0; i-- {
+		if dms[i].meta.Stats.NumSeries/dms[i].meta.Stats.NumTombstones <= 20 { // 5%
+			return [][]string{dms[i].dir}, nil
+		}
+	}
```
fabxc: Didn't you have something in there that only made us consider blocks with a relatively long time range? 
I still that still makes sense to have as short ones will be compacted soon enough.

Gouthamve: Added a rough heuristic now.

chunks/chunk.go
```diff
@@ -43,6 +43,7 @@ type Chunk interface {
 	Encoding() Encoding
 	Appender() (Appender, error)
 	Iterator() Iterator
+	NumSamples() int64
```
fabxc: Just `int` seems fine here.

Gouthamve: Done, PTAL.

fabxc:
👍

PR #109, opened 2017-07-13
Fix compactions

fabxc:
@Gouthamve

### DIFF
--- a/db.go
+++ b/db.go
@@ -99,14 +99,14 @@ type DB struct {
 	metrics *dbMetrics
 	opts    *Options
 
-	// Mutex for that must be held when modifying the general
-	// block layout.
+	// Mutex for that must be held when modifying the general block layout.
+	// cmtx must be held before acquiring it.
 	mtx    sync.RWMutex
 	blocks []Block
 
 	// Mutex that must be held when modifying just the head blocks
 	// or the general layout.
-	// Must never be held when acquiring a blocks's mutex!
+	// mtx must be held before acquiring.
 	headmtx sync.RWMutex
 	heads   []headBlock
 
@@ -341,6 +341,9 @@ func (db *DB) appendableHeads() (r []headBlock) {
 }
 
 func (db *DB) completedHeads() (r []headBlock) {
+	db.mtx.RLock()
+	defer db.mtx.RUnlock()
+
 	db.headmtx.RLock()
 	defer db.headmtx.RUnlock()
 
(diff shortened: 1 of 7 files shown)


db.go
```diff
@@ -677,7 +692,7 @@ func (a *dbAppender) appenderAt(t int64) (*metaAppender, error) {
 	}
 
 	var hb headBlock
-	for _, h := range a.db.appendable() {
+	for _, h := range a.db.appendableHeads() {
```
Gouthamve: Not sure how this works, do this function get called every iteration?

fabxc:
> Compacting only if headFullness > 0.5 seems dangerous. Maybe we should check both highTs and now()

So that was exactly the problem before. We only compacted then and in benchmarks, when our intermediate 30s compaction check didn't hit before a new block was started, we never compacted properly.

The main fix here is to always compact everything that's not the two most recent blocks. That's the old behavior. Now additionally, if we have a chance to catch a 50% full block, we compact this a bit earlier too to distribute our memory spikes a bit.

We cannot reference `now` as TSDB is wall clock agnostic. We could allow passing in a custom `now()` function. But I think what he have here works pretty well in practice and frees us from another config option.

db.go
```diff
@@ -325,37 +325,52 @@ func headFullness(h headBlock) float64 {
...
+	for _, h := range db.heads[:len(db.heads)-2] {
+		if h.ActiveWriters() > 0 {
+			return r
 		}
+		r = append(r, h)
 	}
+	// Add the 2nd last head if the last head is more than 50% filled.
+	// Compacting it early allows us to free its memory before allocating
+	// more for the next block and thus reduces spikes.
+	if h2 := db.heads[len(db.heads)-2]; headFullness(h2) >= 0.5 && h2.ActiveWriters() == 0 {
```
fabxc: If the next value is 40, the next block will be started. Then the 2nd most-recent block immediately moves to position 3 and will be compacted as part of that.

db.go
fabxc: This section of `appenderAt` will only be executed the first time we have to figure out the appropriate head block for timestamp `t`. In the general case, we return within the for loop.

fabxc:
I've no clue why the tests seem to block forever in Travis.

fabxc:
Locking issue fixed. The test was holding an open querier when trying to open a second one. The changes happened to cause timing that exposed a pre-existing issue.
The first querier didn't return its read lock (R1) because the second querier was waiting to acquire a read lock (R2). `reloadBlocks` tried to acquire a write lock (W1), which blocked (R2) but was waiting for (R1) to be released.

Also found two other existing issues around locking order.
Fixed all three and clarified docs around it.

PR #115, opened 2017-08-03
Fix compactions

fabxc:
@Gouthamve I switched to starting with the smallest range. This should guarantee us that the first pick is the one we want instead of correcting afterwards.
Switched from recursive to iterative approach to – recursion is pretty but a for loop seemed easier to grasp in this particular case.

Currently have a prombench running to verify that everything works.

### DIFF
--- a/cmd/tsdb/main.go
+++ b/cmd/tsdb/main.go
@@ -113,7 +113,7 @@ func (b *writeBenchmark) run(cmd *cobra.Command, args []string) {
 	st, err := tsdb.Open(dir, nil, nil, &tsdb.Options{
 		WALFlushInterval:  200 * time.Millisecond,
 		RetentionDuration: 15 * 24 * 60 * 60 * 1000, // 15 days in milliseconds
-		BlockRanges:       tsdb.ExponentialBlockRanges(3*60*60*1000, 3, 5),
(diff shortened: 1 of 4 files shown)


Gouthamve:
This is a much, much more readable version! I over complicated last time. LGTM!

fabxc:
Prombench run so far. Looks good aside from known spikes.

screen shot 2017-08-04 at 09 12 30

Gouthamve:
Hmm, 15+ scenario? So we only pick: `p = parts[0]` and hence we are not looking beyond the first partition. Which could me there could be smaller blocks in further parts that are only picked for a laaarge block size.

Can you try by iterating through the `parts`?

Gouthamve:
Actually no sorry for the noise, we will still pick the smaller blocks because `parts` won't have blocks larger than the interval. 

While this could cause some extra blocks, 15+ is waaay too much. What are block-ranges being used?

fabxc:
Argh, just forgot a for loop. PTAL.

Gouthamve:
Can we add a test case for the part where missing the loop breaks it? Other than that nit, this LGTM

fabxc:
Test case for that is already in there.

PR #116, opened 2017-08-06
Persist series without allocating the full set

fabxc:
Change index persistence for series to not be accumulated in memory
before being written as one large batch. `Labels` and `ChunkMeta`
objects are reused.
This cuts down memory spikes during compaction of multiple blocks
significantly.

As part of the the Index{Reader,Writer} now have an explicit notion of
symbols and series must be inserted in order.

@Gouthamve 

This does not solve all of it yet but has very positive effects. The next effective optimizations are likely around the postings lists.

Max heap usage (using `max_over_time`):

screen shot 2017-08-06 at 12 09 34


head.go
```diff
@@ -332,50 +334,30 @@ func (h *HeadBlock) Snapshot(snapshotDir string) error {
...
 	h.mtx.RLock()
-	defer h.mtx.RUnlock()
-
 	if h.closed {
 		panic(fmt.Sprintf("block %s already closed", h.dir))
 	}
-
-	// Reference on the original slice to use for postings mapping.
-	series := h.series[:]
+	h.mtx.RUnlock()
```
Gouthamve: Defer pattern is better as both are co-located? The instantiation of the blockQuerier should be instantaneous AFAICS.

fabxc: `Index()`, `Chunks()`, and `Tombstones()` do or may in the future lock the mutex too. So defer would call deadlocking here.

fabxc:
Yup, got a bit more complex. But the `SortedPostings` method gets rid of our nasty hacks so far, which is nice.

PR #118, opened 2017-08-09
Avoid chunk allocations and refactor compactions

fabxc:
@Gouthamve 

depends on #116 
Compaction spikes are gone now. Only thing that's causing relevant allocations is rebuilding of postings lists. That can cause spikes with high series churn but TBD whether they are big enough to require optimizations.

I refactored the compactor along the way.

screen shot 2017-08-09 at 08 55 37

### DIFF
--- a/block.go
+++ b/block.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/oklog/ulid"
 	"github.com/pkg/errors"
+	"github.com/prometheus/tsdb/chunks"
 	"github.com/prometheus/tsdb/labels"
 )
 
@@ -181,13 +182,13 @@ type persistedBlock struct {
 	tombstones tombstoneReader
 }
 
-func newPersistedBlock(dir string) (*persistedBlock, error) {
+func newPersistedBlock(dir string, pool chunks.Pool) (*persistedBlock, error) {
 	meta, err := readMetaFile(dir)
 	if err != nil {
 		return nil, err
 	}
 
-	cr, err := newChunkReader(chunkDir(dir))
+	cr, err := newChunkReader(chunkDir(dir), pool)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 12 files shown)


grobie:
:open_mouth: ^5 @fabxc

block.go
```diff
@@ -112,7 +113,7 @@ type BlockStats struct {
 type BlockMetaCompaction struct {
 	// Maximum number of compaction cycles any source block has
 	// gone through.
-	Generation int `json:"generation"`
+	Level int `json:"level"`
```
fabxc: We are beta... it's just a breaking change .__.

chunks.go
```diff
@@ -238,23 +238,22 @@ func (w *chunkWriter) WriteChunks(chks ...*ChunkMeta) error {
...
+		seq = uint64(w.seq()) << 32
+	)
+	for i := range chks {
+		chk := &chks[i]
 
-	for _, chk := range chks {
 		chk.Ref = seq | uint64(w.n)
 
-		n = binary.PutUvarint(b, uint64(len(chk.Chunk.Bytes())))
+		n := binary.PutUvarint(b[:], uint64(len(chk.Chunk.Bytes())))
```
fabxc: We turned `b` into an array instead of a slice, causing it to be allocated on the stack. Escape analysis by the compiler should've done this before too I thought. But it didn't catch it apparently, so I made it explicit.

Anyway, `PutUvarint` wants a slice, not an array, and `[:]` does exactly that.

chunks.go
```diff
@@ -265,7 +264,7 @@ func (w *chunkWriter) WriteChunks(chks ...*ChunkMeta) error {
 		if err := chk.writeHash(w.crc32); err != nil {
 			return err
 		}
-		if err := w.write(w.crc32.Sum(nil)); err != nil {
+		if err := w.write(w.crc32.Sum(b[:0])); err != nil {
```
fabxc: Same as above.

PR #120, opened 2017-08-13
Fix panic caused by 0 division

Gouthamve:
Fixes https://github.com/prometheus/prometheus/issues/3064 but we might have to do another release.

Sorry about this :( I don't know how I missed this check when I introduced it in #108 

What is surprising is how it never turned up in Prombench runs.

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/120)

### DIFF
--- a/compact.go
+++ b/compact.go
@@ -98,11 +98,13 @@ func newCompactorMetrics(r prometheus.Registerer) *compactorMetrics {
 	return m
 }
 
+// LeveledCompactorOptions are the options for a LeveledCompactor.
 type LeveledCompactorOptions struct {
 	blockRanges []int64
 	chunkPool   chunks.Pool
 }
 
+// NewLeveledCompactor returns a LeveledCompactor.
 func NewLeveledCompactor(r prometheus.Registerer, l log.Logger, opts *LeveledCompactorOptions) *LeveledCompactor {
 	if opts == nil {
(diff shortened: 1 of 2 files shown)


fabxc:
Damn – I hit this too before and I thought I had squeezed that in with one of my last PRs.

compact.go
```diff
@@ -170,7 +170,7 @@ func (c *LeveledCompactor) Plan(dir string) ([]string, error) {
 			break
 		}
 
-		if meta.Stats.NumSeries/meta.Stats.NumTombstones <= 20 { // 5%
+		if meta.Stats.NumTombstones > 0 && meta.Stats.NumSeries/meta.Stats.NumTombstones <= 20 { // 5%
```
fabxc: I always find `meta.Stats.NumSeries/(meta.Stats.NumTombstones+1)` cognitively easier

grobie:
> Damn – I hit this too before and I thought I had squeezed that in with one of my last PRs.

Would it be possible to cover that with a test then?

hiscal2015:
I can get prometheus working because of this issue, is it possible to do another release ASAP? Thanks.

Gouthamve:
@fabxc @grobie Added a test-case. It fails pre change and succeeds after.

fabxc:
👍

PR #126, opened 2017-08-21
Log the directory when the WAL cannot be opened

### LINKED ISSUE #125: Missing data for format string 'open WAL %s'
See https://github.com/prometheus/tsdb/blob/e0aca4bee9ae472d1c164c5562eed2a4bf4c700a/db.go#L880: there is no value given for the %s, so it appears literally in the log output.

ankon:
Fixes #125

### DIFF
--- a/db.go
+++ b/db.go
@@ -877,7 +877,7 @@ func (db *DB) openHeadBlock(dir string) (*HeadBlock, error) {
 	)
 	wal, err := OpenSegmentWAL(wdir, l, 5*time.Second)
 	if err != nil {
-		return nil, errors.Wrap(err, "open WAL %s")
+		return nil, errors.Wrapf(err, "open WAL %s", dir)
 	}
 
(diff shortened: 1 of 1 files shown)


db.go
```diff
@@ -877,7 +877,7 @@ func (db *DB) openHeadBlock(dir string) (*HeadBlock, error) {
 	)
 	wal, err := OpenSegmentWAL(wdir, l, 5*time.Second)
 	if err != nil {
-		return nil, errors.Wrap(err, "open WAL %s")
+		return nil, errors.Wrap(err, "open WAL %s", dir)
```
Gouthamve: Thanks, but I think it should be `errors.Wrapf()`. Wrap doesn't take the extra formatting parameters and hence the build is failing, moving to Wrapf should fix it.

ankon: Oops, sorry! That happens when a fix looks too easy. Force-pushed an update now, and ran this through essentially the .travis.yml commands locally.

PR #128, opened 2017-08-25
Expose types for easier interface implementations.

Gouthamve:
TombstoneReader is exposed but Intervals is not.

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/128)

### DIFF
--- a/tombstones.go
+++ b/tombstones.go
@@ -61,8 +61,8 @@ func writeTombstoneFile(dir string, tr tombstoneReader) error {
 		for _, itv := range v {
 			buf.reset()
 			buf.putUvarint32(k)
-			buf.putVarint64(itv.mint)
-			buf.putVarint64(itv.maxt)
+			buf.putVarint64(itv.Mint)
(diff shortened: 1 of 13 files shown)


tombstones.go
```diff
@@ -131,39 +131,40 @@ func readTombstones(dir string) (tombstoneReader, error) {
...
 
-func (t tombstoneReader) add(ref uint32, itv interval) {
+func (t tombstoneReader) add(ref uint32, itv Interval) {
 	t[ref] = t[ref].add(itv)
 }
 
-type interval struct {
+// Interval represents a single time-interval.
+type Interval struct {
 	mint, maxt int64
```
fabxc: Is there a lot of user if we can have an exported TombstoneReader but cannot actually read the values of the intervals it exposes?

Gouthamve:
Updated PTAL.

fabxc:
👍

fabxc:
👍

PR #130, opened 2017-08-26
Add newCRC32 function to simplify hash initialization

mdlayher:
Tidies things up a bit, and makes it easier to change the polynomial in one place later on, if deemed necessary.

### DIFF
--- a/chunks.go
+++ b/chunks.go
@@ -18,7 +18,6 @@ import (
 	"encoding/binary"
 	"fmt"
 	"hash"
-	"hash/crc32"
 	"io"
 	"os"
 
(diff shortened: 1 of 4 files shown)


fabxc:
👍 

Definitely better this way. It doesn't allow us to just change it though as the persisted data depends on the polynomial. Any iteration on this will require adding versioning logic :)

mdlayher:
True. I guess I assume that once 2.0 comes out of beta, it will no longer be changed, and until then, it's easy enough to just wipe the data if large changes are made.

Could be a good idea to open an issue about versioning if that's a concern though.

fabxc:
The disk format already has required versioning fields. Just the code
doesn't care right now since there's just a single one.

PR #134, opened 2017-08-30
Make a DB aware of the first timestamp stored in it

Thib17:
This is needed to fix prometheus/prometheus#3041

### DIFF
--- a/db.go
+++ b/db.go
@@ -520,6 +520,13 @@ func validateBlockSequence(bs []DiskBlock) error {
 	return nil
 }
 
+func (db *DB) Blocks() []DiskBlock {
+	db.mtx.RLock()
+	defer db.mtx.RUnlock()
+
+	return db.blocks
+}
+
 // Close the partition.
 func (db *DB) Close() error {
 	close(db.stopc)


mdlayher:
Seems like you'd probably want to acquire a read-lock on `mtx` since it guards `blocks`?  Also, would it make sense to return a `time.Time` instead of a UNIX timestamp disguised as an int64?

fabxc:
There's no requirement from prometheus/tsdb's side for stored timestamps to
map to wall time in any way and esp. the chosen precision.
So returning `time.Time` would not make sense here.

Thib17:
Thanks @mdlayher and @fabxc for your feedbacks. I added a lock on mtx.

db.go
```diff
@@ -566,6 +567,46 @@ func validateBlockSequence(bs []Block) error {
...
+func (db *DB) StartTime() int64 {
+	db.mtx.RLock()
+	defer db.mtx.RUnlock()
+
+	b := db.blocks[0]
+	startTime := int64(math.MaxInt64)
+
+	indexr := b.Index()
+	joblabel := "job"
+	tpls, err := indexr.LabelValues(joblabel)
```
fabxc: That's leaking Prometheus semantics into TSDB, which we generally want to avoid.
I think it's somewhat questionable how well the overall query has something to do with TSDB.

 It's only accessing public interfaces so it doesn't have to be in here.
Anything speaking against implementing this in `prometheus/storage/tsdb`?

iksaif: Could work.

Would you see another way to get the first time without having to do this hack with a label ? Maybe iterating on the chunks directly ? (I'm not super familiar with the interfaces yet)

Thib17: @fabxc It doesn't only access public interfacess. It need access to blocks list which are not public yet (and I thought it was private on purpose)
@iksaif I didn't find an other way. Especially I couldn't access to a chunks list i could iterate on it. 

I only see two possible solutions :
(1) make blocks public adding a BD.Blocks() func (and move the logic to `prometheus/storage/tsdb`)
(2) iterate on existing labels (using block.Index().LabelIndices()) instead of using "job" labels. It means keeping the same logic without prometheus specific labelname.
What do you think ?

fabxc: Making certain methods public if there's a use case for it is perfectly valid. The exported interfaces so far are largely on a best-guess basis :)

Gouthamve:
Hmm, anything speaking against just returning the MinTime in the oldest block?

db.go
Thib17: Done

Thib17:
@Gouthamve the MinTime in the oldest block  seems to be not very accurate (could be several hours from the MinTime of the first chunk AFAIK)
And adding a public method to access blocks list enable you to just select `db.Blocks()[0].MinTime` if you want.

Gouthamve:
@Thib17 Sorry for the late reply, yes, but it is not possible that there exist a sample in remote-storage that doesn't exist locally.

For example if `MinT` of DB  = 10 and `MinT` of first chunk is 15, then tsdb has seen no sample between 10 and 15. Which means the remote store won't have it. If we ever delete a block for retention at `t=X`, we delete all samples before `X` and the new `MinT` for DB will be `X`.

This doesn't hold in the case of a manual series delete, but I don't think we need to optimise for that.

iksaif:
@Gouthamve , see https://github.com/prometheus/prometheus/issues/10#issuecomment-285690034

See the remote storage as something that's really durable, and local disk as something that can easily fail and get lost. This is even more true when you start running prometheus on marathon/mesos/kube/borg.

Gouthamve:
Yes, I agree to that.

Hmm, it will be very rare that data is corrupted in just the last chunk of a particular series. In that case, this makes sense, but in the general case, I think just looking at the `MinT` of the last block would work.

I am just saying instead of trying to lookup the `MinTime` of each series, it just makes sense to look at the `db.Blocks()[0].MinTime` as the mintime for each series.

In case there is corruption and you lost a block, `db.Blocks()[0].MinTime` would still give you the last **valid** block. But now with the exposure of `Blocks()` method, I think this is an implementation detail on Prometheus' side.

Thib17:
@Gouthamve I think we agree and move the discussion on Prometheus' side : `prometheus/prometheus/pull/3129`

@fabxc Is there anything missing before merging this PR ?

Gouthamve:
Fabian is on a vacation now but this PR looks good (: Looking forward to see what comes up on prometheus/prometheus#3129

PR #135, opened 2017-09-01
Don't do blocking disk I/O under mutex blocking scrape and eval

alin-amana:
As described in [a post on the Prometheus Users list](https://groups.google.com/forum/#!topic/prometheus-users/jPne7PZhw78), I have been having some issues with the 2.0 beta running on a VM with a busy disk. Scraping and rule evaluation would regularly block for tens of seconds at a time on `SegmentWAL.mtx`, which was held by some goroutine waiting on a `fdatasync()` syscall.

This PR attempts to remove that syscall from under the mutex and in about 24 hours of running on my small deployment it has shown significant improvement (zero scrapes or evals missed) and (more importantly) hasn't crashed yet. :o)

### DIFF
--- a/wal.go
+++ b/wal.go
@@ -298,20 +298,39 @@ func (w *SegmentWAL) tail() *os.File {
 
 // Sync flushes the changes to disk.
 func (w *SegmentWAL) Sync() error {
-	w.mtx.Lock()
-	defer w.mtx.Unlock()
+	var tail *os.File
+	var err error
 
-	return w.sync()
+	// Flush the writer and retrieve the reference to the tail segment under mutex lock
+	func() {
+		w.mtx.Lock()
+		defer w.mtx.Unlock()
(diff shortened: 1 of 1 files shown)


fabxc:
Great, if there's no issue with fsyncing asynchronously, we should do that in cases where we are just doing it periodically anyway. Glad that we can still make things work in this case – 10s fsync latency is really rough :)

alin-amana:
`fsync()` and `fdatasync()` are thread safe, per POSIX requirements.

It is somewhat unclear whether they will block concurrent `write()` operations (found a [blog post](http://oldblog.antirez.com/post/fsync-different-thread-useless.html) discussing this, but the empirical results in the comments seem to point in whichever way you want them to). But Prometheus does cached writes and so shouldn't always run into this, only when the buffer fills up.

PR #136, opened 2017-09-05
WIP: use packages from prometheus/common instead of prometheus/prometheus/pkg

seiflotfy:
This commit depends on the change https://github.com/prometheus/common/pull/101 which allows developing and using tsdb without needing prometheus dev branch

### DIFF
--- a/cmd/tsdb/main.go
+++ b/cmd/tsdb/main.go
@@ -29,8 +29,8 @@ import (
 	"unsafe"
 
 	"github.com/pkg/errors"
-	promlabels "github.com/prometheus/prometheus/pkg/labels"
-	"github.com/prometheus/prometheus/pkg/textparse"
+	promlabels "github.com/prometheus/common/labels"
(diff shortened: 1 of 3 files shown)


brian-brazil:
May I ask why you want to change this?
@fabxc

fabxc:
The dependency should only be used for some tests and the benchmark. The main library does not depend on it AFAIK and thus it should work okay with vendoring.

It's a bit annoying right now because you have to checkout the dev-2.0 branch, agreed. But that problem will solve itself soon enough.

I think the prometheus/pkg/textparse should not be in common. It's a very specialized way of parsing, purely done to serve some optimizations in Prometheus. The API is awkward and of little use in the more general case.

seiflotfy:
All good... I'll close the PR in a bit... Sorry about that

seiflotfy:
FWIW since prometheus depends on tsdb, would it maybe make sense to move text parse and labels over to tsdb?

Issue #107, opened 2017-07-12
Ideas for TSDB CLI tool


fabxc:
* A `ls` that gives a tabular view of blocks and info about them
* Scanning tool that checks for corruption and reports on which sections in which file are affected
* Doing compactions on blocks offline
* Getting info about values for labels, series for matchers, sizes of postings lists – anything querying related basically
* Downsampling all or some series in blocks. 
* Tombstoning (and possibly immediately compacting) a set of series in blocks. That's effectively dynamic retention.

Lots of things possible. But not all of them are super useful aside from things being possible in theory.
Some are possible through Prometheus' APIs in one way or another. But it seems fine to have multiple interfaces to the same functionality.

krasi-georgiev:
DIBS

BasPH:
@krasi-georgiev I want to implement some of the other suggestions, are you working on it?

krasi-georgiev:
@BasPH not at the moment so go ahead.

krasi-georgiev:
- [x] [ls](https://github.com/prometheus/tsdb/blob/2aae939d6cbebac9e4ad7324287bb9a6bd135d13/cmd/tsdb/main.go#L48) that gives a tabular view of blocks and info about them
- [x] [Scanning tool](https://github.com/prometheus/tsdb/pull/320) that checks for corruption and reports on which sections in which file are affected
- [ ] Doing compactions on blocks offline
- [ ] Getting info about values for labels, series for matchers, sizes of postings lists – anything querying related basically
- [ ] Downsampling all or some series in blocks. 
- [ ] Tombstoning (and possibly immediately compacting) a set of series in blocks. That's effectively dynamic retention.

krasi-georgiev:
please share use cases and I can implement the remaining ideas that would be useful.

FUSAKLA:
I'd like to take a look at dumping of blocks to human readable data for debugging once I'll get to it.

krasi-georgiev:
sure , just don't forget to discuss how/why would this be useful to get some feedback before starting the design.

codesome:
@krasi-georgiev @FUSAKLA 

I plan to add new features to TSDB CLI in my GSoC (Section 4.5 in my [proposal](https://docs.google.com/document/d/1aB5zGkhrLdWOT6RHhMA2Qd08T-puwlhcnTbjBP3fkx4/edit?usp=sharing)). I will be discussing the design with the team soon. I am surely open for suggestions.

[...]

PR #142, opened 2017-09-09
replaced cobra with kingpin for the tsdb cli tool

krasi-georgiev:
#107 as part of extending the cli tool replacing cobra with kingpin to be consistent with the other packages
Signed-off-by: Krasi Georgiev <krasi.root@gmail.com>

### DIFF
--- a/cmd/tsdb/main.go
+++ b/cmd/tsdb/main.go
@@ -18,8 +18,6 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"net/http"
-	_ "net/http/pprof"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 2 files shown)


krasi-georgiev:
cc @Gouthamve

cmd/tsdb/main.go
```diff
@@ -33,35 +31,30 @@ import (
...
-	root := &cobra.Command{
-		Use:   "tsdb",
-		Short: "CLI tool for tsdb",
-	}
-
-	root.AddCommand(
-		NewBenchCommand(),
-	)
+var (
+	cli = kingpin.New(filepath.Base(os.Args[0]), "CLI tool for tsdb")
```
Gouthamve: Can you move this into main() itself? You could also make benchWriteFile a member of the `writeBenchmark` struct and that way you don't need the global variable.

krasi-georgiev:
@Gouthamve  no problem , have a look now. 
I left in `var()` block as `gofmt` formats the assignment easier to read this way with tabular indentation.

Gouthamve:
LGTM!

PR #147, opened 2017-09-18
Simplify series create logc in head

fabxc:
@Gouthamve


fabxc:
@Gouthamve PTAL. Also added a test for the WAL reading semantics.

Gouthamve:
:+1:

PR #150, opened 2017-09-20
Ensure postings are always sorted

fabxc:
@Gouthamve

### DIFF
--- a/head.go
+++ b/head.go
@@ -207,7 +207,7 @@ func (h *Head) ReadWAL() error {
 			}
 			ms := h.series.getByID(s.Ref)
 			if ms == nil {
-				h.logger.Log("msg", "unknown series reference in WAL", "ref", s.Ref)
+				h.logger.Log("msg", "unknown series reference in WAL", "ref", s.Ref, "ts", s.T, "mint", mint)
 				continue
 			}
 			_, chunkCreated := ms.append(s.T, s.V)
@@ -267,7 +267,7 @@ func (h *Head) Truncate(mint int64) error {
 
 	start = time.Now()
 
-	p, err := h.indexRange(mint, math.MaxInt64).Postings("", "")
+	p, err := h.indexRange(mint, math.MaxInt64).Postings(allPostingsKey.Name, allPostingsKey.Value)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


head.go
```diff
@@ -1038,8 +1038,6 @@ func (s *stripeSeries) getOrSet(hash uint64, series *memSeries) (*memSeries, boo
 		return prev, false
 	}
 	s.hashes[i].set(hash, series)
-
-	s.hashes[i][hash] = append(s.hashes[i][hash], series)
```
fabxc: This was a leftover and we were actually storing series twice – without any negative effects aside from space consumption.

postings.go
```diff
@@ -45,21 +45,44 @@ func (p *memPostings) get(name, value string) Postings {
...
 
+func (p *memPostings) addFor(id uint64, l labels.Label) {
+	list := append(p.m[l], id)
+	p.m[l] = list
+
+	// There is no guarantee that no higher ID was inserted before as they may
+	// be generated independently before adding them to postings.
+	// We repair order violations on insert. The invariant is that the first n-1
+	// items in the list are already sorted.
+	for i := len(list) - 1; i >= 1; i-- {
```
Gouthamve: There is still a race here, but I might be wrong. Also this case is a long shot, but hear me out :)

Let the existing list be: [1, 2, 3]
Now we insert: 4, 5, 6, 7 concurrently in this way:
[1, 2, 3, 6, 4, 5, 7].

Now, if the check for `5` runs before check for `4`, which means in the end, we end up with: `[1, 2, 3, 4, 6, 5, 7]`

Gouthamve: Also, if we have `[1, 2, 3]` and insert 6 and then 4 and then 5.
We get at one point: 
```
list := [1, 2, 3, 6, 4, 5]
p.m[l] = list
```

Even the swapping for `6, 4` that happens is not visible to the new list as it is a copy, AFAICS.

fabxc: The insertion itself is guarded by mutex, so there is no concurrency at play here.
Before each insert all elements so far are sorted. Just the last one is potentially out of order.

Gouthamve: Aah, got it. LGTM!

Gouthamve:
:+1:

mattbostock:
What's the impact if they're not sorted? It might be good to include an explanation in the commit message.

fabxc:
Done.

mattbostock:
Thanks @fabxc.

PR #151, opened 2017-09-21
Use boolean function instead of postings to drop WAL series

fabxc:
There is not guarantee or requirement for WAL writers to only add
series entries in increasing order of IDs. A postings list cannot look
back and thus unordered WAL entries would skip over IDs to not truncate
from the WAL.
We replace it with a simple boolean check function that does not require
order.

@Gouthamve


fabxc:
It's a pretty small change and I've tested it a fair bit, so I take it :)

PR #156, opened 2017-10-01
Add list blocks command to CLI

BasPH:
I've added an `ls` command to the tsdb cli for listing blocks in the DB (https://github.com/prometheus/tsdb/issues/107). It uses [github.com/ryanuber/columnize](url) for printing tabular.

Example output of `tsdb ls`:
```
BLOCK ULID                  MIN TIME  MAX TIME  NUM SAMPLES  NUM CHUNKS  NUM SERIES
01BVCQAJVXZA5DFKWV6DRYE9W3  0         7200000   35900000     300000      100000
01BVCQAT7894QCN38QV8DVGPWJ  7200000   14400000  36000000     300000      100000
01BVCQB11494K3ZS7MDXDPRP4B  14400000  21600000  36000000     300000      100000
```

### DIFF
--- a/cmd/tsdb/main.go
+++ b/cmd/tsdb/main.go
@@ -60,7 +60,7 @@ func main() {
 		if err != nil {
 			exitWithError(err)
 		}
-		db.PrintBlocks()
+		printBlocks(db.Blocks())
 	}
 	flag.CommandLine.Set("log.level", "debug")
 }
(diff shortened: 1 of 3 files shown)


mattbostock:
The Go stdlib has a package for writing tabular columns, did you consider
using that?

https://golang.org/pkg/text/tabwriter/

BasPH:
I looked at the Kubernetes code and they're using the tabwriter. There was a [TODO in the code](https://github.com/kubernetes/kubernetes/blob/628df56b2cbbb74d73dd83da6168b17232292dd3/cmd/kubeadm/app/cmd/upgrade/plan.go#L71) to replace with something like columnize. With columnize printing tables is a one-liner, with the tabwriter it's a little more code. Is there a preference for using stdlib/external packages?

fabxc:
Some preference if possible, since @TheTincho has to package every new dep up for Debian and we are constantly causing him troubles :)
If you can write a 10 line helper function and save a dep in return, that's certainly better.

fabxc:
Thanks for the changes. Having the printing as a direct method to `DB` itself is not ideal. A getter like `DB.Block()` would probably be better. The formatting and printing can then all be handled independently in the main.go.

BasPH:
Thanks. `DB.Blocks()` already exists :) Agree that handling printing is better in main.go, so I'll move it there.

fabxc:
Thanks 👍

TheTincho:
Thanks @fabxc for thinking of me, really appreciate that! <3

PR #158, opened 2017-10-02
Allocate and cache strings for persisted blocks

fabxc:
This change loads the full symbol table when we open a persisted block
and allocates a string for each. This ensures that strings retrieved
through the index can be used after the block was closed.
Before we backed the strings by the mmap'd byte regions which would
segfault in this case.

Also remove an inconsistency in the disk format and move both offset
tables to the end (breaking change).

@grobie @Gouthamve 

This is already tested extensively through existing tests.


fabxc:
It was used in the past but its not anymore.

Gouthamve:
👍

PR #161, opened 2017-10-04
Remove dependency on etcd/pkg/fileutil

fabxc:
@TheTincho I hope this helps a bit.

### DIFF
--- a/compact.go
+++ b/compact.go
@@ -20,12 +20,12 @@ import (
 	"sort"
 	"time"
 
-	"github.com/prometheus/tsdb/fileutil"
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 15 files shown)


fabxc:
@Gouthamve second  commit fsyncs the temporary directory before rename rather than the renamed one afterwards. That might be the cause why the `meta.json` files are gone sometimes.

compact.go
```diff
@@ -753,10 +752,8 @@ func renameFile(from, to string) error {
 	defer pdir.Close()
```
Gouthamve: Shouldn't we remove this defer?

fabxc: Yes

Issue #81, opened 2017-05-15
Improve instrumentation


fabxc:
Instrumentation so far is pretty sparse:

* Make compaction duration metric a summary (no generally applicable histogram buckets possible)
* Fsync duration of the WAL
* WAL corruptions detected
 
There's probably a lot more. Though a lot of performance metrics around querying aren't really possible due to lazy evaluation.

mwitkow:
Can we have a histogram of series per block? This would allow users to see whether the blocks are well distributed or there is massive variance in their sizes.

Also a measure of count of all stored series (in "live blocks") would be fantastic: it would allow comparisons with P1. Additionally, the number of series in WAL-stage and non-commited to blocks would also be incredibly useful.

fabxc:
Instrumentation is pretty good now. Anything related to histograms/summaries around blocks doesn't really work though. The cardinality of observations is so low that the resulting graphs look practically broken.

krasi-georgiev:
@fabxc can you think of any other metric that is needed right now?
Otherwise maybe can close this one  and open new issues for any new metric that would be useful to add.

PR #162, opened 2017-10-04
Instrument WAL fsync

BasPH:
Instrumented the WAL and exposed fsync duration metric as suggested in https://github.com/prometheus/tsdb/issues/81.

### DIFF
--- a/db.go
+++ b/db.go
@@ -204,7 +204,7 @@ func Open(dir string, l log.Logger, r prometheus.Registerer, opts *Options) (db 
 		return nil, errors.Wrap(err, "create leveled compactor")
 	}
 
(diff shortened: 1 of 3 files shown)


fabxc:
Thanks. Change looks good in general.
Just needs callers of the constructor to be adjusted so it compiles :)

BasPH:
Whoops my bad, missed something while committing. I pushed again and now tests fail at `TestWALRestoreCorrupted` but I cannot point out what's failing. Tests pass on my machine, do you happen to know what's wrong?

Gouthamve:
Looks like a flake, restarted the build and it succeeded!

fabxc:
Thanks! 👍

Issue #113, opened 2017-07-26
Export the oldest timestamp in the db


Gouthamve:
This would help us query the remote storage from the correct timestamp and also enable other use-cases such as https://github.com/prometheus/prometheus/issues/2988

This is trivial to implement as it would be a method on `*DB` where we just return the `MinTime` of the oldest block.

Gouthamve:
@ventris Do you want to give this a shot? This is actually quite simple and would help you understand the structure.

The `*DB` object has a list of blocks and each block exposes it `MinTime` and `MaxTime`. So it would be easy to expose that.

ventris:
Yes, I will give this a shot.

krasi-georgiev:
@ventris , @Gouthamve is it ok to take this one ?

Gouthamve:
There is this PR open to address this I think: https://github.com/prometheus/tsdb/pull/134

krasi-georgiev:
@Gouthamve that is against master , will it get merged to dev-2.0 as well?

Gouthamve:
Yes, this is the storage engine used by dev-2.0.

So the master of this repo (prometheus/tsdb) is vendored into the dev-2.0 of the prometheus/prometheus repo. If you are looking for issues to pick, I am happy to provide pointers for #107

krasi-georgiev:
@Gouthamve great , I will read the info and ping you if I need any pointers.

ventris:
@krasi-georgiev yes please go ahead, I haven't had the time to begin.

Gouthamve:
Closed in #134 and #152

PR #165, opened 2017-10-06
Expose mintime in the DB

BasPH:
Expose the oldest timestamp in the DB for https://github.com/prometheus/tsdb/issues/113. It loops over all blocks and exposes the smallest MinTime as metric `tsdb_oldest_timestamp`.


Gouthamve:
Woops! Thanks a lot for the contrib, but we have #134 and #152 for the same :( I forgot to close the issue after the merge. Sorry for the trouble!

BasPH:
Whoops no worries!

PR #166, opened 2017-10-06
Load postings in batch on startup

fabxc:
This allows to insert IDs to postings out of order until
a trigger function is called. This avoids the insertion sort we usually
do which can be very costly since WAL entries are more out of order than
regular adds.

@grobie @Gouthamve


grobie:
So who is going to call `ensureOrder`?

PR #167, opened 2017-10-06
Instrument WAL corruptions

simonpasquier:
This is one of the improvements described in #81. Not totally sure if this is the correct place where to add the metric though.


fabxc:
Thanks! This is the right place!

Basically, we stop after the first detected corruption so this value can only ever be 0 or 1 in the general case. We could just make it a true/false gauge. But given tsdb is a library, people may use it differently and a counter could actually be the better choice in some cases.

👍

PR #168, opened 2017-10-06
wal: decode and process in separate threads.

fabxc:
@grobie this is merely a constant improvement getting us from 1 to 1.15 cores being used.

By my measurements the initial 4.x million series from the data you sent me are processed after ~20 seconds. So most of the time will then actually will be spent on replaying the samples themselves, given I had only 5 out of a few dozen files with sample data.


fabxc:
Are we good for merging this aggregate PR?

Gouthamve:
@fabxc Yep :+1:, sorry for the delay.

PR #169, opened 2017-10-09
wal: parallelize sample processing

fabxc:
@grobie @Gouthamve 

It's not perfect but seems to reliably get us about 80-85% core utilization. At least for my 8 cores.

Retaining the order ends us up with a loop of channels. That just introduces a fundamental inefficiency due to the circular dependency.

### DIFF
--- a/head.go
+++ b/head.go
@@ -15,6 +15,7 @@ package tsdb
 
 import (
 	"math"
+	"runtime"
 	"sort"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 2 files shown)


brancz:
Go runtime/go-routine work distribution parts seem sane, but for the remainder another review would be good, as I'm not familiar enough (yet) with the surrounding tsdb code. :slightly_smiling_face:

head.go
```diff
@@ -195,8 +227,30 @@ func (h *Head) ReadWAL() error {
...
 	// Track number of samples that referenced a series we don't know about
 	// for error reporting.
-	var unknownRefs int
+	var unknownRefs uint64
+
+	// Start workers that each process samples for a partition of the series ID space.
+	// They are connected through a ring of channels which ensures that all sample batches
+	// read from the WAL are processed in order.
+	var (
+		n          = runtime.GOMAXPROCS(0)
```
Gouthamve: Can we use `NumCPU`? https://golang.org/pkg/runtime/#GOMAXPROCS

brancz: I think this was done intentionally to use the configured go procs.

fabxc: By default `GOMAXPROCS` will be set to `NumCPU()`. However, if it is set manually to a different number, we should respect that. The runtime will only schedule onto `GOMAXPROCS` threads at a time so anything above that only introduces more active goroutines requiring coordination overhead even though they cannot actually run in parallel.

head.go
```diff
@@ -195,8 +227,30 @@ func (h *Head) ReadWAL() error {
...
+	)
+	for i := 0; i < n; i++ {
+		output := make(chan []RefSample, 300)
+
+		go func(i int, input <-chan []RefSample, output chan<- []RefSample) {
+			unknown := h.processWALSamples(mint, uint64(i), uint64(n), input, output)
+			atomic.AddUint64(&unknownRefs, unknown)
+		}(i, input, output)
+
+		input = output
```
Gouthamve: Can we have a comment here or somewhere that this is being done to reuse the sample buffer? (Is it what it is for?)

Gouthamve:
LGTM

head.go
fabxc: Yes, done.

PR #170, opened 2017-10-09
Fix innocuous typo in variable names

simonpasquier:
This change fixes the variable names holding the tsdb_head_max_time and
tsdb_head_min_time metrics. It is a cosmetic change to improve the
code readability as the metric values are taken from the correct
variables.

This PR is pure nit-picking from my side and doesn't solve any concrete issue except my own perplexity when I've read these lines.


fabxc:
Thanks 👍

PR #175, opened 2017-10-11
Clarify postings index semantics, handle staleness

fabxc:
The postings list index may point to series that no longer
exist during garbage collection. This clarifies that this is valid
behavior.
It would be possible, though more complex, to always keep them in sync.
However, series existance means nothing in itself as the queried time
range defines whether there's actual data. Thus our definition is sane
overall as long as drift is kept small.

@grobie @Gouthamve

### DIFF
--- a/cmd/tsdb/main.go
+++ b/cmd/tsdb/main.go
@@ -332,20 +332,22 @@ func exitWithError(err error) {
 	os.Exit(1)
 }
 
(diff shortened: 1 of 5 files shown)


grobie:
Looks good. What about a counter metric for missing timeseries so that I can verify they only happen during GC?

PR #176, opened 2017-10-12
Races

fabxc:
@grobie This addresses 2 out 3 of the races you reported. The last one is on a file descriptor and attempts a sync after close.

 However, the file already got synced before closed so this has no consequences at this point. Skipping it for now as 'fixing' it causes unnecessary complexity right now. I'd like to address in a better way but will likely defer that PR past the release for safety.


fabxc:
The push check doesn't pick up the change merged in master which fixes the compile error. PR check is passing.

PR #181, opened 2017-10-20
Fix dangling chunk reference panic

fabxc:
This is just an intermediate fix. It's not perfect and query artefacts may happen. However, this is much better than the panics and their consequences we are seeing currently.

This being pretty straightforward, I think it is worth merging and getting into Prometheus.
The proper fix of the underlying problem is more complex and I'd like to think more about a simple solution rather than bringing in complication that may cause other regressions right now.

@grobie @beorn7 could you take this for a spin on the servers where you saw the issue?

### DIFF
--- a/chunks/chunk.go
+++ b/chunks/chunk.go
@@ -69,6 +69,17 @@ type Iterator interface {
 	Next() bool
 }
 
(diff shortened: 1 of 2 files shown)


fabxc:
Filed https://github.com/prometheus/tsdb/issues/182

PR #183, opened 2017-10-20
Truncate segments on broken header

fabxc:
Addresses parts of https://github.com/prometheus/prometheus/issues/3283 so we at least get a clean restart.


Gouthamve:
Test was a flake. LGTM :+1:

PR #184, opened 2017-10-20
Prefix all metrics with `prometheus_*`

fabxc:


brancz:
Passing a `registry.WithPrefix("prometheus")` would be great, but this is as good as we can do for now I guess. :+1:

beorn7:
> Passing a `registry.WithPrefix("prometheus")` would be great,

Working on it.

PR #187, opened 2017-10-25
Ensure near-empty chunks end at correct boundary

fabxc:
We were determining a chunk's end time once it was one quarter full to
compute it so all chunks have uniform number of samples.
This accidentally skipped the case where series started near the end of
a chunk range/block and never reached that threshold. As a result they
got persisted but were continued across the range.

This resulted in corrupted persisted data.

@smarterclayton @EdSchouten I believe this is the source of the issue you are seeing. At the very least this seems to reliably fix the unordered and repeated samples that were reported.
It also fixes artefacts of rate() calculations for me. Though mine looked a bit different – but given that a very fundamental invariant is violated in the raw data, I'd expect different shapes of artefacts anyway.

Would you be open to run a version with this and some other recent patches to confirm whether this fixes things? I'd do another RC afterwards.

### DIFF
--- a/head.go
+++ b/head.go
@@ -1162,6 +1162,10 @@ func (s *memSeries) cut(mint int64) *memChunk {
 	}
 	s.chunks = append(s.chunks, c)
 
+	// Set upper bound on when the next chunk must be started. An earlier timestamp
+	// may be chosen dynamically at a later point.
+	_, s.nextAt = rangeForTimestamp(mint, s.chunkRange)
+
 	app, err := c.chunk.Appender()
 	if err != nil {
 		panic(err)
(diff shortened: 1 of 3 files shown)


EdSchouten:
Hey! Sorry for not being able to test this for you today; I'm OOO on Wednesdays. Will test master first thing tomorrow morning.

fabxc:
That would be great, thanks. Actually trying to get rc.2 out with those changes since I could reliably reproduce the issue and this here fixed it.

EdSchouten:
Oh, good! I'll then patiently wait until RC2 comes out. :+1:

Issue #163, opened 2017-10-05
TestWALRestoreCorrupted is flaking


Gouthamve:
That test is flaking for some reason, needs investigation.

Ref: https://github.com/prometheus/tsdb/pull/162

Gouthamve:
Not sure if I am just (un) lucky but not able to replicate the flake locally. Are people seeing this locally too? 

And how can I look into travis' environment?

nipuntalukdar:
TestWALRestoreCorrupted sometime fails as the test routine and wal.cut method may updating the same segment file in parallel resulting in error.  Putting a sleep of around 2 seconds (hoepfully cut is done by 2 seconds) will fix the issue.

PR #190, opened 2017-10-27
wal: synchronize background operations

fabxc:
This adds an actor channel and thereby serializes all background
operations through the run() method.
Fixes an existing race.

@beorn7 @grobie

### DIFF
--- a/wal.go
+++ b/wal.go
@@ -190,6 +190,7 @@ type SegmentWAL struct {
 
 	stopc   chan struct{}
 	donec   chan struct{}
+	actorc  chan func() error // sequantilized background operations
 	buffers sync.Pool
 }
 
@@ -213,6 +214,7 @@ func OpenSegmentWAL(dir string, logger log.Logger, flushInterval time.Duration, 
 		flushInterval: flushInterval,
 		donec:         make(chan struct{}),
 		stopc:         make(chan struct{}),
+		actorc:        make(chan func() error, 1),
 		segmentSize:   walSegmentSizeBytes,
 		crc32:         newCRC32(),
 	}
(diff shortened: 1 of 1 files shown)


beorn7:
I'm currently not in a state where I could do a qualified review, but I'll happily try out this patch on our test servers.

beorn7:
I have built a Prom2 binary with this change (see commit 0de912895ff9973f032a20f6900ec7bcc9a6a906 in the prometheus/prometheus repo). At first, everything looked well, but after a day or two, two of our three servers went into some kind of stall. HTTP requests are still responded to, but there seem to be no ingestion or compaction taking place, almost no CPU usage. Nothing suspicious in the logs.

I haven't done any detailed investigation, and I don't know if this behavior has been introduced by this change, or we are just running into it now by bad luck. 

I have attached the goroutine dump of the two servers that appear to have stalled.


[prom2-collapsed-0de912895ff9973f032a20f6900ec7bcc9a6a906-stalled.20171029.txt](https://github.com/prometheus/tsdb/files/1425570/prom2-collapsed-0de912895ff9973f032a20f6900ec7bcc9a6a906-stalled.20171029.txt)

[prom2-collapsed-0de912895ff9973f032a20f6900ec7bcc9a6a906-stalled-2.20171029.txt](https://github.com/prometheus/tsdb/files/1425571/prom2-collapsed-0de912895ff9973f032a20f6900ec7bcc9a6a906-stalled-2.20171029.txt)

fabxc:
Thanks. Yea, same deadlock in either coming from this PR. Basically we attempt to send on `actorc` while holding the WALs mutex. The processing loop started calling `Sync` though which waits for the mutex to be released. So `Sync` doesn't complete, and sending on `actorc` keeps blocking which means the mutex doesn't get released.

fabxc:
Wrapped the channel sends in a goroutine again. This es essentially the same working behavior as before. But we still use the actor channel to serialize the operations in general.

beorn7:
Running now on our test servers.
BTW: The test that's failing in Travis is passing for me locally.

fabxc:
Yes, it's a flaky test (issue #163).

beorn7:
FYI: This has now been running without any problem for the last 2d. Also, no races detected in the one server that ran with race detection enabled.

wal.go
```diff
@@ -190,6 +190,7 @@ type SegmentWAL struct {
 
 	stopc   chan struct{}
 	donec   chan struct{}
+	actorc  chan func() error // sequantilized background operations
```
juliusv: sequantilized -> sequentialized

juliusv:
Looks good to me, but tests are still failing. Clicked on Retry in Travis just in case it's a flake...

Issue #191, opened 2017-10-30
Compact failed error in windows environment


ranbochen:
tsdb did not close file before rename it case "compaction failed" on windows.

https://github.com/prometheus/tsdb/blob/f347eac33d3e0e9a8ab67cd12d7e99b58dd84155/tombstones.go#L43

see:
[https://github.com/prometheus/prometheus/issues/3253](https://github.com/prometheus/prometheus/issues/3253)

PR #192, opened 2017-10-31
fix bugs on platform windows to pass all test case.

ranbochen:
#191 

[prometheus issues#3253](https://github.com/prometheus/prometheus/issues/3253)

### DIFF
--- a/chunks.go
+++ b/chunks.go
@@ -171,12 +171,7 @@ func (w *chunkWriter) finalizeTail() error {
 		return err
 	}
 
-	if err := tf.Close(); err != nil {
-		return err
-	}
-
-	// close dir file (if not windows platform will fail on rename)
-	return w.dirFile.Close()
+	return tf.Close()
 }
 
 func (w *chunkWriter) cut() error {
@@ -282,7 +277,12 @@ func (w *chunkWriter) seq() int {
 }
 
 func (w *chunkWriter) Close() error {
-	return w.finalizeTail()
+	if err := w.finalizeTail(); err != nil {
+		return err
+	}
+
+	// close dir file (if not windows platform will fail on rename)
+	return w.dirFile.Close()
 }
 
 // ChunkReader provides reading access of serialized time series data.
--- a/chunks.go
+++ b/chunks.go
@@ -170,7 +170,13 @@ func (w *chunkWriter) finalizeTail() error {
 	if err := tf.Truncate(off); err != nil {
 		return err
 	}
-	return tf.Close()
+
+	if err := tf.Close(); err != nil {
+		return err
+	}
+
+	// close dir file (if not windows platform will fail on rename)
+	return w.dirFile.Close()
 }
 
 func (w *chunkWriter) cut() error {
--- a/db_windows.go
+++ b/db_windows.go
@@ -21,8 +21,7 @@ import (
 
 func mmap(f *os.File, sz int) ([]byte, error) {
 	low, high := uint32(sz), uint32(sz>>32)
-
-	h, errno := syscall.CreateFileMapping(syscall.Handle(f.Fd()), nil, syscall.PAGE_READONLY, low, high, nil)
+	h, errno := syscall.CreateFileMapping(syscall.Handle(f.Fd()), nil, syscall.PAGE_READONLY, high, low, nil)
 	if h == 0 {
 		return nil, os.NewSyscallError("CreateFileMapping", errno)
 	}
--- a/wal.go
+++ b/wal.go
@@ -384,7 +384,7 @@ func (w *SegmentWAL) Truncate(mint int64, keep func(uint64) bool) error {
 		w.putBuffer(buf)
 
 		if err != nil {
-			return err
+			return errors.Wrap(err, "write to compaction segment")
 		}
 	}
 	if r.Err() != nil {
@@ -401,14 +401,15 @@ func (w *SegmentWAL) Truncate(mint int64, keep func(uint64) bool) error {
 	csf.Sync()
 	csf.Close()
 
+	candidates[0].Close() // need close before remove on platform windows
 	if err := renameFile(csf.Name(), candidates[0].Name()); err != nil {
-		return err
+		return errors.Wrap(err, "rename compaction segment")
 	}
 	for _, f := range candidates[1:] {
+		f.Close() // need close before remove on platform windows
 		if err := os.RemoveAll(f.Name()); err != nil {
 			return errors.Wrap(err, "delete WAL segment file")
 		}
-		f.Close()
 	}
 	if err := w.dirFile.Sync(); err != nil {
 		return err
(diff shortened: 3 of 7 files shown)


ranbochen:
@Gouthamve
@fabxc

chunks.go
```diff
@@ -170,7 +170,13 @@ func (w *chunkWriter) finalizeTail() error {
...
 		return err
 	}
-	return tf.Close()
+
+	if err := tf.Close(); err != nil {
+		return err
+	}
+
+	// close dir file (if not windows platform will fail on rename)
+	return w.dirFile.Close()
```
fabxc: I don't believe this is correct. `finalizeTail` is called whenever we move on to the next chunk file. Thus after the first chunk file is completed we'd end up with a closed file handle for `dirFile`.

Which rename does this attempt to fix?

ranbochen: oh！
How about  move it too chunks.Close?

fabxc: Yes, that should be good.

db_windows.go
```diff
@@ -21,8 +21,7 @@ import (
 
 func mmap(f *os.File, sz int) ([]byte, error) {
 	low, high := uint32(sz), uint32(sz>>32)
-
-	h, errno := syscall.CreateFileMapping(syscall.Handle(f.Fd()), nil, syscall.PAGE_READONLY, low, high, nil)
+	h, errno := syscall.CreateFileMapping(syscall.Handle(f.Fd()), nil, syscall.PAGE_READONLY, high, low, nil)
```
fabxc: That's a pretty significant change, did I get the order wrong?
Seems so from https://msdn.microsoft.com/en-us/library/windows/desktop/aa366537(v=vs.85).aspx I suppose.

wal_test.go
```diff
@@ -271,8 +271,9 @@ func TestWALRestoreCorrupted_invalidSegment(t *testing.T) {
 	Ok(t, err)
 	f, err := wal.createSegmentFile(dir + "/000001")
 	Ok(t, err)
-	_, err = wal.createSegmentFile(dir + "/000002")
+	f2, err := wal.createSegmentFile(dir + "/000002")
 	Ok(t, err)
+	Ok(t, f2.Close()) // close for windows!
```
fabxc: No need for extra comment. That close should've been there regardless of OS.

db_windows.go
ranbochen: yes, that make it create a huge mmap. so  CreateFileMapping failed.

wal.go
```diff
@@ -702,6 +713,10 @@ func (w *SegmentWAL) Close() error {
 	if hf := w.head(); hf != nil {
 		return errors.Wrapf(hf.Close(), "closing WAL head %s", hf.Name())
 	}
+
+	if w.dirFile != nil {
```
fabxc: Is there any case that would cause `dirFile` to be nil? I don't think so in general and just letting it panic seems ok given it would be a contract violation.

wal.go
```diff
@@ -401,14 +401,15 @@ func (w *SegmentWAL) Truncate(mint int64, keep func(uint64) bool) error {
 	csf.Sync()
 	csf.Close()
 
+	candidates[0].Close() // need close before remove on platform windows
```
fabxc: Is not closing really causing issues on windows or are you just assuming that it is more strict about requiring things to be closed explicitly? I'd like to avoid platform/OS specific comments unless there's really a tricky caveat.

ranbochen: yes there are issues!
every change is to make sure, it pass test case.

ranbochen: you can comment some change, and run test on windows. 
it will fail.

ranbochen:
There are still some issues with file OP, need to dig deeper!
****
level=info ts=2017-11-01T01:00:12.1912693Z caller=compact.go:361 component=tsdb msg="compact blocks" count=3 mint=1509472800000 maxt=1509494400000
level=error ts=2017-11-01T01:00:12.3639651Z caller=db.go:260 component=tsdb msg="compaction failed" err="delete compacted block: remove data\\01BXT3R1RX8Q0TVX7E6NZCEM5F\\chunks\\000001: The process cannot access the file because it is being used by another process."
level=error ts=2017-11-01T01:00:13.3727503Z caller=db.go:260 component=tsdb msg="compaction failed" err="plan compaction: open data\\01BXT3R1RX8Q0TVX7E6NZCEM5F\\meta.json: The system cannot find the file specified."
level=error ts=2017-11-01T01:01:17.3786091Z caller=db.go:260 component=tsdb msg="compaction failed" err="plan compaction: open data\\01BXT3R1RX8Q0TVX7E6NZCEM5F\\meta.json: The system cannot find the file specified."
level=error ts=2017-11-01T01:02:25.3913068Z caller=db.go:260 component=tsdb msg="compaction failed" err="plan compaction: open data\\01BXT3R1RX8Q0TVX7E6NZCEM5F\\meta.json: The system cannot find the file specified."
level=error ts=2017-11-01T01:03:41.401283Z caller=db.go:260 component=tsdb msg="compaction failed" err="plan compaction: open data\\01BXT3R1RX8Q0TVX7E6NZCEM5F\\meta.json: The system cannot find the file specified."
level=error ts=2017-11-01T01:05:13.4098578Z caller=db.go:260 component=tsdb msg="compaction failed" err="plan compaction: open data\\01BXT3R1RX8Q0TVX7E6NZCEM5F\\meta.json: The system cannot find the file specified."
level=error ts=2017-11-01T01:07:17.4176678Z caller=db.go:260 component=tsdb msg="compaction failed" err="plan compaction: open data\\01BXT3R1RX8Q0TVX7E6NZCEM5F\\meta.json: The system cannot find the file specified."
level=error ts=2017-11-01T01:10:17.4254843Z caller=db.go:260 component=tsdb msg="compaction failed" err="plan compaction: open data\\01BXT3R1RX8Q0TVX7E6NZCEM5F\\meta.json: The system cannot find the file specified."
level=error ts=2017-11-01T01:13:17.4332892Z caller=db.go:260 component=tsdb msg="compaction failed" err="plan compaction: open data\\01BXT3R1RX8Q0TVX7E6NZCEM5F\\meta.json: The system cannot find the file specified."
level=error ts=2017-11-01T01:16:17.440245Z caller=db.go:260 component=tsdb msg="compaction failed" err="plan compaction: open data\\01BXT3R1RX8Q0TVX7E6NZCEM5F\\meta.json: The system cannot find the file specified."
level=error ts=2017-11-01T01:19:17.447945Z caller=db.go:260 component=tsdb msg="compaction failed" err="plan compaction: open data\\01BXT3R1RX8Q0TVX7E6NZCEM5F\\meta.json: The system cannot find the file specified."
level=error ts=2017-11-01T01:22:17.4557646Z caller=db.go:260 component=tsdb msg="compaction failed" err="plan compaction: open data\\01BXT3R1RX8Q0TVX7E6NZCEM5F\\meta.json: The system cannot find the file specified."
level=error ts=2017-11-01T01:25:17.4664968Z caller=db.go:260 component=tsdb msg="compaction failed" err="plan compaction: open data\\01BXT3R1RX8Q0TVX7E6NZCEM5F\\meta.json: The system cannot find the file specified."

****

ranbochen:
DB holds all blocks fd, so the RemoveAll OP failed here.
https://github.com/prometheus/tsdb/blob/a27cf34a36c0357cbd9225a658019030c4b490ea/db.go#L382

Can I  close  and clear all db.blocks before Remove? (db will call reload just after Remove OP)

Need find a better way to handle windows file operations！

@fabxc

PR #193, opened 2017-11-01
Fix #191

### LINKED ISSUE #191: Compact failed error in windows environment

ranbochen:
#191

Fix remove file issue on windows, when compact blocks(block count > 1)

Need find a better way to handle windows file operations! It's hard to find which file must be closed!

### DIFF
--- a/db.go
+++ b/db.go
@@ -379,6 +379,27 @@ func (db *DB) compact() (changes bool, err error) {
 		}
 		changes = true
 
+		// close blocks in plan so that we can remove files.
+		var blocks []*Block
+		db.mtx.Lock()
+		oldBlocks := db.blocks
+		for _, b := range oldBlocks {
+			keep := true
+			for _, pd := range plan {
+				if pd == b.Dir() {
+					keep = false
+					break
+				}
+			}
+			if keep {
+				blocks = append(blocks, b)
+			} else {
+				b.Close()
+			}
+		}
+		db.blocks = blocks
+		db.mtx.Unlock()
+
 		for _, pd := range plan {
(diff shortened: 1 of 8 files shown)


ranbochen:
after this patch, it work on my windows server now.

@fabxc

grobie:
@beorn7

beorn7:
Do we have a qualified reviewer for this? @fabxc @Gouthamve ?
This looks like it should go into rc3 . (I assume we don't want to release something that doesn't work on MS Windows?)

db.go
```diff
@@ -379,6 +379,27 @@ func (db *DB) compact() (changes bool, err error) {
...
+				}
+			}
+			if keep {
+				blocks = append(blocks, b)
+			} else {
+				b.Close()
+			}
+		}
+		db.blocks = blocks
+		db.mtx.Unlock()
```
fabxc: This will cause data for the given time range to become unavailable until the reload, which is called later gets completed.

Also, introspecting `db.blocks` in this code path kinda throws a few things together that were separated before.

But I suppose windows not allowing us to delete open files literally forces us to have this small blip where data is not accessible. Anyone sees a way around this?

fabxc:
I'll merge this now – maybe we can find an alternative way for the described problem in the future.

PR #194, opened 2017-11-03
Fixed a problem of adding padding of 4 zero bytes in some cases

nipuntalukdar:
Fixed a small problem of adding a padding of 4 zero bytes when the index file-size is already multiple of 4.

### DIFF
--- a/index.go
+++ b/index.go
@@ -35,6 +35,10 @@ const (
 	MagicIndex = 0xBAAAD700
 
 	indexFormatV1 = 1
+
+	size_unit = 4 // should be power of 2
+
+	padding_and = size_unit - 1
 )
 
 const indexFilename = "index"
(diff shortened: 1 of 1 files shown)


index.go
```diff
@@ -203,10 +207,10 @@ func (w *indexWriter) write(bufs ...[]byte) error {
...
 }
 
-// addPadding adds zero byte padding until the file size is a multiple of n.
-func (w *indexWriter) addPadding(n int) error {
-	p := n - (int(w.pos) % n)
-	if p == 0 {
+// addPadding adds zero byte padding until the file size is a multiple size_unit.
+func (w *indexWriter) addPadding() error {
+	p := size_unit - (int(w.pos) & padding_and)
+	if p == size_unit {
```
fabxc: Would this be potentially simpler?

```go
p := w.pos % size_unit
if p == 0 {
    return nil
}
p = size_unit - p
```

That would skip the extra `padding_and` constant – `&` always need a few looks when reading it without context.

fabxc:
Thanks, good catch! Added a comment that could potentially make it a bit simpler.

index.go
nipuntalukdar: Okay. Will do the change. % is sometimes expensive, but anyway 4 is a very small number.

fabxc: It will probably be optimized into `&` by the compiler – but we are not calling too often in the grand scheme of things anyway :)

fabxc:
Thanks 👍

PR #195, opened 2017-11-03
db: delete old blocks during reload

fabxc:
Windows requires blocks to be closed before deleting their directories.
This adds a set of deleteable blocks to reload(), which then deletes
them without causing disruption to querying.

@beorn7 @grobie thought of this after merging #193.

### DIFF
--- a/db.go
+++ b/db.go
@@ -378,36 +378,9 @@ func (db *DB) compact() (changes bool, err error) {
 			return changes, errors.Wrapf(err, "compact %s", plan)
 		}
 		changes = true
-
-		// close blocks in plan so that we can remove files.
-		var blocks []*Block
-		db.mtx.Lock()
-		oldBlocks := db.blocks
-		for _, b := range oldBlocks {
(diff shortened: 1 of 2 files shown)


fabxc:
@ranbochen could you give this  a look as well?

db.go
```diff
@@ -461,7 +434,18 @@ func (db *DB) getBlock(id ulid.ULID) (*Block, bool) {
 	return nil, false
 }
 
-func (db *DB) reload() (err error) {
+func stringsContain(set []string, elem string) bool {
```
grobie: Why not strings.Contain?

fabxc: That finds substrings in strings. We need to find a string in a slice.

grobie: Facepalm. My first thought was "I wonder how often this function was written in go codebases", but then got distracted by the function name.

ranbochen:
Good idea to change **reload** func.
I tested it on my laptop with random-exporter. It works.

PR #198, opened 2017-11-08
Use Go 1.9 math/bits funcs for counting leading / trailing zeros

dswarbrick:


fabxc:
Thanks 👍 

I'm personally cool with bumping the minimum Go version. We also have pending PRs in prometheus/prometheus waiting for that. Any objections @brian-brazil?

PR #202, opened 2017-11-09
index: simplify checksum validation

fabxc:
We can just take a CRC32 over everything we want to decode beforehand. This way we don't have to track offsets and reconstruct the decoded byte range in checkCRC.


fabxc:
Superseded by #203

PR #203, opened 2017-11-09
Generalize index and chunk readers

fabxc:


chunks.go
```diff
@@ -298,44 +298,60 @@ type ChunkReader interface {
...
 
 	// Closers for resources behind the byte slices.
 	cs []io.Closer
 
 	pool chunks.Pool
 }
 
-// newChunkReader returns a new chunkReader based on mmaped files found in dir.
-func newChunkReader(dir string, pool chunks.Pool) (*chunkReader, error) {
+func newChunkReader(bs []ByteSlice, cs []io.Closer, pool chunks.Pool) (*chunkReader, error) {
```
Bplotka: Is the `newChunkReader` used anywhere beside the `NewDirChunkReader` func? If no, I would inline this in `NewDirChunkReader`. Otherwise that magic validation part could be confusing (:

index.go
```diff
@@ -585,27 +585,52 @@ var (
...
+
+func (b realByteSlice) Range(start, end int) []byte {
+	return b[start:end]
+}
+
+func (b realByteSlice) Sub(start, end int) ByteSlice {
+	return b[start:end]
+}
+
 // NewIndexReader returns a new IndexReader on the given directory.
```
Bplotka: Is that comment up-to-date? I think it returns IndexReader from the given byte slice

index.go
```diff
@@ -585,27 +585,52 @@ var (
...
-	f, err := openMmapFile(filepath.Join(dir, "index"))
+func NewFileIndexReader(path string) (IndexReader, error) {
+	f, err := openMmapFile(path)
 	if err != nil {
 		return nil, err
 	}
+	return newIndexReader(realByteSlice(f.b), f)
+}
+
+// newIndexReader returns a new indexReader on the given directory.
```
Bplotka: Another obsolete comment

index.go
```diff
@@ -585,27 +585,52 @@ var (
...
 // NewIndexReader returns a new IndexReader on the given directory.
-func NewIndexReader(dir string) (IndexReader, error) { return newIndexReader(dir) }
+func NewIndexReader(b ByteSlice) (IndexReader, error) {
+	return newIndexReader(b, nil)
+}
 
-// newIndexReader returns a new indexReader on the given directory.
-func newIndexReader(dir string) (*indexReader, error) {
-	f, err := openMmapFile(filepath.Join(dir, "index"))
+func NewFileIndexReader(path string) (IndexReader, error) {
```
Bplotka: No comment

index.go
```diff
@@ -625,7 +651,17 @@ func newIndexReader(dir string) (*indexReader, error) {
 }
 
 func (r *indexReader) readTOC() error {
-	d := r.decbufAt(len(r.b) - indexTOCLen)
+	if r.b.Len() < indexTOCLen {
+		return errInvalidSize
```
Bplotka: No wrap? To indicate the exact section?

index.go
```diff
@@ -625,7 +651,17 @@ func newIndexReader(dir string) (*indexReader, error) {
...
+	}
+	b := r.b.Range(r.b.Len()-indexTOCLen, r.b.Len())
+
+	expCRC := binary.BigEndian.Uint32(b[len(b)-4:])
+	d := decbuf{b: b[:len(b)-4]}
+
+	if d.crc32() != expCRC {
+		return (errInvalidChecksum)
```
Bplotka: Again, some wrap would be useful to indicate the exact section

index.go
```diff
@@ -634,32 +670,61 @@ func (r *indexReader) readTOC() error {
...
-
 	return d.err()
 }
 
+// decbufAt returns a new decoding buffer. It expects the first 4 bytes
+// after offset to hold the big endian encoded content length, followed by the contents and the expected
+// checksum.
 func (r *indexReader) decbufAt(off int) decbuf {
-	if len(r.b) < off {
+	if r.b.Len() < off+4 {
```
Bplotka: Not 100% sure if it is worth it, but maybe worth to put this 4 number to `bigEndLen` constant? Maybe will be more clear what the `4` indicates.

index.go
```diff
@@ -634,32 +670,61 @@ func (r *indexReader) readTOC() error {
...
+		return decbuf{e: errInvalidChecksum}
+	}
+	return dec
 }
 
-func (r *indexReader) checkCRC(crc uint32, off, cnt int) (bool, error) {
-	r.crc32.Reset()
-	if len(r.b) < off+cnt {
-		return false, errInvalidSize
+// decbufAt returns a new decoding buffer. It expects the first bytes
```
Bplotka: `s/decbufAt/decbufUvarintAt`

index.go
```diff
@@ -634,32 +670,61 @@ func (r *indexReader) readTOC() error {
...
+// after offset to hold the uvarint-encoded buffers length, followed by the contents and the expected
+// checksum.
+func (r *indexReader) decbufUvarintAt(off int) decbuf {
+	// We never have to access this method at the far end of the byte slice. Thus just checking
+	// against the MaxVarintLen32 is sufficient.
+	if r.b.Len() < off+binary.MaxVarintLen32 {
```
Bplotka: To be honest, `decbufUvarintAt` and `decbufAt` are totally the same, just differ in one argument.. do you think it would be worth to reuse the code? (I know - only two places, but quite important ones) (:

index.go
```diff
@@ -881,12 +904,7 @@ func (r *indexReader) Series(ref uint64, lbls *labels.Labels, chks *[]ChunkMeta)
 			MaxTime: maxt,
 		})
 	}
-
-	if valid, err := r.checkCRC(d1.be32(), int(ref)+sl, l); !valid {
```
Bplotka: Are we checking it anywhere in Series? Or we just don't want to do it, because it is already checked somewhere in  `decbuf`?

index.go
fabxc: Yea, but the differences are spread across those mostly common lines – so extracting shared code and calling it would probably be a lot more LOC and harder to follow I felt.

index.go
fabxc: The checksums are now validated as part of `decbufAt`/`decbufUvarintAt`

chunks.go
fabxc: Ops, no, `NewChunkReader` should actually call it too.

index.go
fabxc: Mh, I think that line was overwritten by subsequent comments.

PR #206, opened 2017-11-13
Add explicit error to Querier.Select

fabxc:
This has been a frequent source of debugging pain since errors are
potentially delayed to a much later point. They bubble up in an
unrelated execution path.

Originally seemed like a neat idea – it was not :)

### DIFF
--- a/block.go
+++ b/block.go
@@ -285,7 +285,10 @@ func (pb *Block) Delete(mint, maxt int64, ms ...labels.Matcher) error {
 	}
 
 	pr := newPostingsReader(pb.indexr)
-	p, absent := pr.Select(ms...)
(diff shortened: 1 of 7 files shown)


grobie:
Yes please. This will make the error handling of remote read endpoints easier.

I assume we'll want to change the interface everywhere, also in Prometheus. If we have a better name for the `Querier()` function (not the interface) this might be a good moment to think about other changes to the storage/querier interface. @tomwilkie @beorn7

fabxc:
Yes, that has to propagate to Prometheus.

fabxc:
@Gouthamve is this good to go?

Gouthamve:
Yep 👍

PR #209, opened 2017-11-22
Partial chunk deletions are not compacted away

Gouthamve:
This is a bug when a partial chunk is deleted. We are creating a new chunk without the deletions but assigned it to a copied struct, essentially dropping it to the floor.

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/209)

### DIFF
--- a/compact.go
+++ b/compact.go
@@ -525,22 +525,24 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta, 
 
 		if len(dranges) > 0 {
 			// Re-encode the chunk to not have deleted values.
-			for _, chk := range chks {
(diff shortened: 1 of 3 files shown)


compact.go
```diff
@@ -553,22 +553,24 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
...
+					return err
+				}
+
+				it := &deletedIterator{it: chk.Chunk.Iterator(), intervals: dranges}
+				for it.Next() {
+					ts, v := it.At()
+					app.Append(ts, v)
+				}
+
+				chks[i].Chunk = newChunk
```
fabxc: Do we need to update the time range of the ChunkMeta as well?

Gouthamve: Shouldn't make a huge difference. Sometimes the range can be in-between the start and end too. 
Overall given, how rare a partial chunk delete is, there should not be any visible impact.

PR #210, opened 2017-11-22
Add a function to cleanup tombstones.

Gouthamve:
Still not happy with the interfaces. But let me know if a function like this makes sense.

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/210)

### DIFF
--- a/db.go
+++ b/db.go
@@ -122,6 +122,7 @@ type dbMetrics struct {
 	reloads              prometheus.Counter
 	reloadsFailed        prometheus.Counter
 	compactionsTriggered prometheus.Counter
+	tombCleanTimer       prometheus.Histogram
 }
 
 func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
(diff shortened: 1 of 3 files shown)


Gouthamve:
Tests will fail until #209 is merged

db.go
```diff
@@ -686,6 +692,37 @@ func (db *DB) Delete(mint, maxt int64, ms ...labels.Matcher) error {
...
+	deleted := []string{}
+	for _, b := range blocks {
+		ok, err := b.CleanTombstones(db.Dir(), db.compactor)
+		if err != nil {
+			return errors.Wrapf(err, "clean tombstones: %s", b.Dir())
+		}
+
+		if ok {
+			deleted = append(deleted, b.Dir())
+		}
```
fabxc: It looks like we just delete any block the we applied tombstones to?

Gouthamve: Yes, thats because we are compacting the data into a new block.

fabxc: Oh right, ULID changes – makes sense

fabxc:
👍

PR #211, opened 2017-11-23
A small optimization

nipuntalukdar:
A small optimization when updating the highest timestamp of a segment file while adding samples to log.


nipuntalukdar:
Sometimes if we query using the same db object that was used for adding samples, doesn't return the complete data. If we use fresh db object to query, it returns the entire data. It can be checked from the example here : https://gist.github.com/nipuntalukdar/d19e7374db26e54dbc4ef7fe00082b80

I made a small change in head index reader to honour the time range passed for the query, which fixes the problem.

nipuntalukdar:
Made a small change that will prevent creation of block dirs with zero samples during compaction.

Gouthamve:
Hi, thanks for the PR and the efforts and sorry we did not get back to you earlier. 

Can you make separate PRs for each of the following:
1. Added fatal error handling when writes to wal file fails
2. Fix for `TestWALRestoreCorrupted`
3. prevent creation of block dirs with zero samples

This will make it much easier to review and I will make comments on each. Right now everything is mixed up and its not easy to review.

nipuntalukdar:
@Gouthamve I will do that.

PR #218, opened 2017-11-30
Fdatasync on read to flush any unflushed data.

Gouthamve:
This is to handle partial writes from a previous crash.

Fixes prometheus/prometheus#3487

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/218)

### DIFF
--- a/wal.go
+++ b/wal.go
@@ -523,6 +523,12 @@ func (w *SegmentWAL) openSegmentFile(name string) (*os.File, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	// Sync the file first to see any unflushed buffers.
+	if err := fileutil.Fdatasync(f); err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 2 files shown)


brian-brazil:
This change doesn't make sense, according to POSIX this should only help if the machine crashes.

This would only help if we were doing direct hardware access to the disk.

wal.go
```diff
@@ -523,6 +523,12 @@ func (w *SegmentWAL) openSegmentFile(name string) (*os.File, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	//Sync the file first to see any unflushed buffers.
+	if err := fileutil.Fdatasync(f); err != nil {
+		return nil, err
+	}
+
```
fabxc: I really believe this should[tm] not be necessary but doesn't seem like the worst idea for safety.

Space missing in front of comment.

PR #219, opened 2017-11-30
Move index and chunk encoders to own packages

fabxc:
So this is kind of a big change. 

It overall seemed like the number of implementations we have for various interfaces is growing and we are kind of running out of sane names for them.
Also all kinds of types are used whenever convenient, e.g. for certain tests, which makes it a bit messy and causes lots of internal dependencies.

I think it is time to break things into multiple packages to address those problems. This is obviously just a first step to keep it somewhat short. We'll want to migrate some types around, change some names, and do some cleanups.

A secondary purpose is that I'd like to export more stuff in general without polluting the top level package's API. We have a really nice public interfaces for `{Chunk,Index}{Reader,Writer}` we don't want to extend.  
But to work with those files in a flexible well direct access to their lower level concerns is really necessary.

@Gouthamve


fabxc:
It's obviously mostly renamings, exporting types, switching to those types. But in some places I had to make actual changes to make things work out.

One place where extra review care should be taken is the cleanup of postings lists in the head garbage collection: https://github.com/prometheus/tsdb/pull/219/files#diff-0ced3454d447442a96a77b33ebe87300R630
It could even have performance implications but it also should be a bit easier to understand now.

block.go
```diff
@@ -23,10 +23,101 @@ import (
...
+	Close() error
+}
+
+// IndexReader provides reading access of serialized index data.
+type IndexReader interface {
+	// Symbols returns a set of string symbols that may occur in series' labels
+	// and indices.
+	Symbols() (map[string]struct{}, error)
+
+	// LabelValues returns the possible label values
```
Bplotka: Missing trailing period

block.go
```diff
@@ -23,10 +23,101 @@ import (
...
+
+	// Series populates the given labels and chunk metas for the series identified
+	// by the reference.
+	// Returns ErrNotFound if the ref does not resolve to a known series.
+	Series(ref uint64, lset *labels.Labels, chks *[]chunks.Meta) error
+
+	// LabelIndices returns the label pairs for which indices exist.
+	LabelIndices() ([][]string, error)
+
+	// Close released the underlying resources of the reader.
```
Bplotka: `s/released/releases`

block.go
```diff
@@ -23,10 +23,101 @@ import (
...
+	// of chunks that the index can reference.
+	// Implementations may require series to be insert in increasing order by
+	// their labels.
+	// The reference numbers are used to resolve entries in postings lists that
+	// are added later.
+	AddSeries(ref uint64, l labels.Labels, chunks ...chunks.Meta) error
+
+	// WriteLabelIndex serializes an index from label names to values.
+	// The passed in values chained tuples of strings of the length of names.
+	WriteLabelIndex(names []string, values []string) error
```
Bplotka: Can we use `StringTuples` interface you defined later on?

chunks/chunks.go
```diff
@@ -11,18 +11,22 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package tsdb
+package chunks
```
Bplotka: just `chunk`? e.g `chunk.Meta`, `chunk.Writer`?

chunks/chunks.go
```diff
@@ -53,62 +57,30 @@ func (cm *ChunkMeta) writeHash(h hash.Hash) error {
...
-	}
+var (
+	errInvalidSize     = fmt.Errorf("invalid size")
+	errInvalidFlag     = fmt.Errorf("invalid flag")
+	errInvalidChecksum = fmt.Errorf("invalid checksum")
+)
 
-	return false
-}
+// The table gets initialized with sync.Once but may still cause a race
```
Bplotka: This comment relates rather to init body, right? Can we move it there?

compact.go
```diff
@@ -514,7 +516,7 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
 			allSymbols[s] = struct{}{}
 		}
 
-		all, err := indexr.Postings(allPostingsKey.Name, allPostingsKey.Value)
+		all, err := indexr.Postings("", "")
```
Bplotka: Maybe would be nice to add some comment that this means fetching postings for all labels?

block.go
```diff
@@ -23,10 +23,101 @@ import (
...
+	Symbols() (map[string]struct{}, error)
+
+	// LabelValues returns the possible label values
+	LabelValues(names ...string) (index.StringTuples, error)
+
+	// Postings returns the postings list iterator for the label pair.
+	// The Postings here contain the offsets to the series inside the index.
+	// Found IDs are not strictly required to point to a valid Series, e.g. during
+	// background garbage collections.
+	Postings(name, value string) (index.Postings, error)
```
Bplotka: You could comment also that you can put empty name and value too

index/index.go
```diff
@@ -206,18 +192,18 @@ func (w *indexWriter) write(bufs ...[]byte) error {
 }
 
 // addPadding adds zero byte padding until the file size is a multiple size_unit.
-func (w *indexWriter) addPadding() error {
-	p := w.pos % size_unit
+func (w *Writer) addPadding(size int) error {
```
Bplotka: why not `size uint64`

index/postings.go
```diff
@@ -117,9 +122,61 @@ func (p *memPostings) ensureOrder() {
...
+func (p *MemPostings) Delete(deleted map[uint64]struct{}) {
+	var keys []labels.Label
+
+	p.mtx.RLock()
+	for l := range p.m {
+		keys = append(keys, l)
+	}
+	p.mtx.RUnlock()
+
+	for _, l := range keys {
```
Bplotka: I am pretty sure these keys can go outdated pretty quick in race conditions, between this loop iterations and the initial keys append.

The only result would be new key in `p.m` being added in the meantime and not covered by this Delete method.. I guess this is fine?

querier.go
```diff
@@ -202,9 +204,9 @@ func (q *blockQuerier) Close() error {
 // PostingsForMatchers assembles a single postings iterator against the index reader
 // based on the given matchers. It returns a list of label names that must be manually
 // checked to not exist in series the postings list points to.
-func PostingsForMatchers(index IndexReader, ms ...labels.Matcher) (Postings, []string, error) {
+func PostingsForMatchers(ix IndexReader, ms ...labels.Matcher) (index.Postings, []string, error) {
```
Bplotka: `ix`? Maybe `ir` is more relevant to the name if you want to make it short

index/postings.go
```diff
@@ -117,9 +122,61 @@ func (p *memPostings) ensureOrder() {
...
+		p.mtx.Lock()
+
+		found := false
+		for _, id := range p.m[l] {
+			if _, ok := deleted[id]; ok {
+				found = true
```
Bplotka: You are not doing `repl = append(repl, id)` here to lock as short time as possible, right?

block.go
fabxc: I'm indeed not that happy with that area. I was actually wondering to move the reader to this format instead as it is less complex to use. Will have to think about it. But probably something for a later PR. I want to keep this one to the minimum amount of moving existing stuff around.

index/postings.go
fabxc: Anything added after we got the keys should by definition be something relevant to the new block range and thus not being relevant for GC.

index/postings.go
fabxc: Yea and to not reallocate every single postings list on GC even if it wasn't changed.

block.go
```diff
@@ -23,10 +23,101 @@ import (
...
+	// SortedPostings returns a postings list that is reordered to be sorted
+	// by the label set of the underlying series.
+	SortedPostings(index.Postings) index.Postings
+
+	// Series populates the given labels and chunk metas for the series identified
+	// by the reference.
+	// Returns ErrNotFound if the ref does not resolve to a known series.
+	Series(ref uint64, lset *labels.Labels, chks *[]chunks.Meta) error
+
+	// LabelIndices returns the label pairs for which indices exist.
```
Bplotka: Can we add more commentary on this? The `label pair` could be misleading, since it indicates label-value pairs, but it seems there are just label names here with the `[][]string` type prepared for composite indices. Can we explain that shortly here?

fabxc:
Rebased and threw in the rest I had in the pipeline. It's a big review anyway, so we might just get it over with in one go.

chunks/chunks.go
Gouthamve: Comment needs to be updated.

index/index.go
```diff
@@ -642,15 +601,56 @@ func newIndexReader(b ByteSlice, c io.Closer) (*indexReader, error) {
...
+}
+
+// Range marks a byte range.
+type Range struct {
+	Start, End int64
+}
+
+// PostingsRanges returns a new map of byte range in the underlying index file
+// for all postings lists.
+func (r *Reader) PostingsRanges() (map[labels.Label]Range, error) {
```
Gouthamve: This is a weird addition, is it being used in thanos?

index/postings.go
```diff
@@ -117,9 +122,61 @@ func (p *memPostings) ensureOrder() {
...
+	var keys []labels.Label
+
+	p.mtx.RLock()
+	for l := range p.m {
+		keys = append(keys, l)
+	}
+	p.mtx.RUnlock()
+
+	for _, l := range keys {
+		p.mtx.Lock()
```
Gouthamve: We are locking and unlocking repeatedly here, any reason why we are not doing a lock before and after the loop?

index/postings.go
Gouthamve: Can you add that as a comment? I did not understand why you were selecting the keys first until I read this.

compact.go
fabxc: Yea, there was some kind of awkwardness/conflict during refactoring. In absence of a better idea I just reverted to this. Will think again on how to solve it.

index/index.go
fabxc: Yea, I agree – mostly weird because it is _just_ for postings and nothing else? Or weird as data to expose in general.

For context, if we were to redo the format (or do a v2) I'd probably strictly move all references (except for symbols) to be of shape `(offset,length,crc32)` rather than just `offset`. It's really  painful to not know how much you have to fetch from `offset` – and that's not just true for object storages but even for a `File.Read` based implementation.

index/postings.go
fabxc: Simply for locking granularity. Walking every single posting and processing deletes on it may well take a bit of time. If we locked until we have processed each, we'd potentially block any kind of read for multiple seconds.

index/index.go
Gouthamve: +1 to include length! I thought about the exact same thing several times now!

index/postings.go
Gouthamve: Could you also include this as a comment? Will help anyone reading the code.

Issue #208, opened 2017-11-21
Move away from testify/require


Gouthamve:
Current we have both our own utils and `testify/require` for our test assertions. Move away from require to only the util functions.

Gouthamve:
Closed by #220. Thanks a ton @cstyan!

PR #220, opened 2017-12-07
Use internal test utils rather than testify package

cstyan:
For #208 

WIP, I still need to go through and add useful messages to a lot of the tests. Could probably use some help there. All the tests pass locally.

I moved util_test.go to a separate package.

cc: @Gouthamve @fabxc


fabxc:
Thanks, this was probably quite tedious. Much appreciated :)

PR #222, opened 2017-12-10
Close db so that file space is reclaimed immediately after delete

Gouthamve:
@TheTincho Can you check if this fixes the space issues?

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/222)


TheTincho:
While it is difficult to trace if every test is cleaning up properly, I can confirm this is no longer filling up my disk during testing. Thanks!

PR #226, opened 2017-12-14
decouple tsdb main.go (and tests) from prometheus/pkg

cstyan:
Does the thing! Let me know if anything I'm doing is not sane, just made the changes and made sure the tests still passed :)

@Gouthamve @fabxc @TheTincho

### DIFF
--- a/cmd/tsdb/main.go
+++ b/cmd/tsdb/main.go
@@ -14,6 +14,7 @@
 package main
 
 import (
+	"bufio"
 	"flag"
 	"fmt"
 	"io"
@@ -22,15 +23,13 @@ import (
 	"path/filepath"
 	"runtime"
 	"runtime/pprof"
+	"strings"
 	"sync"
 	"text/tabwriter"
 	"time"
-	"unsafe"
 
 	"github.com/go-kit/kit/log"
 	"github.com/pkg/errors"
-	promlabels "github.com/prometheus/prometheus/pkg/labels"
-	"github.com/prometheus/prometheus/pkg/textparse"
 	"github.com/prometheus/tsdb"
 	"github.com/prometheus/tsdb/labels"
 	"gopkg.in/alecthomas/kingpin.v2"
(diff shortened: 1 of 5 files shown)


head_test.go
```diff
@@ -47,26 +45,39 @@ func BenchmarkCreateSeries(b *testing.B) {
 	}
 }
 
+func mapToLabels(m map[string]interface{}, l *labels.Labels) {
+	for k, v := range m {
+		*l = append(*l, labels.Label{Name: k, Value: v.(string)})
+	}
+}
```
fabxc: This function doesn't seem to be used anywhere.

cmd/tsdb/main.go
```diff
@@ -302,20 +301,32 @@ func measureTime(stage string, f func()) time.Duration {
...
 
-func readPrometheusLabels(r io.Reader, n int) ([]labels.Labels, error) {
-	b, err := ioutil.ReadAll(r)
-	if err != nil {
-		return nil, err
+func mapToLabels(m map[string]interface{}, l *labels.Labels) {
+	for k, v := range m {
+		*l = append(*l, labels.Label{Name: k, Value: v.(string)})
 	}
+}
```
fabxc: This function doesn't seem to be used anywhere.

TheTincho:
Thanks!!

cstyan:
@fabxc the reason I'm scanning then parsing the text and using `strings.Replace` is because initially I had used the json package. However, because decoding arbitrary json into `interface{}` results in a map, the labels end up in random orders otherwise some of the WAL tests fail. It was only later that @Gouthamve pointed out that the labels just need to be in sorted order.

So I went with the current implementation. There are however options for using the json package to decode and then sorting. We can decode the json, and then created a sorted slice of the keys from the map, and then generate the `Labels` slice in the correct order. Or we can just create the `Labels` slice and then sort that. In either case it takes about the same amount of time. However both solutions using sorting take about twice as long as the current implementation. It's about 300ms the sorting method for 20k series vs 150ms for the implementation in this PR.

Goutham mentioned that we never really need to process much more than 20k series at once via this function? So I guess it's then just a question of how often this function is run in most normal operating cases.

Up to you guys which implementation you prefer, I'm not particular attached to any :)

cmd/tsdb/main.go
cstyan: oops, this is leftover from my first implementation, see comment below.

fabxc:
It's never run for regular ops and just relevant for testing and benchmarking. The latter is where we have larger files sometimes. Overall speed is nice but not crazy important.
I suppose since we just have a few k/v pairs, it's fine to do it with a custom decoder – but sorting of the final label set needs to be added regardless.

cstyan:
Okay, that sounds like we can't assume the stored series have labels that are already sorted? The test files had series with all sorted labels.

fabxc:
Thanks 👍

PR #245, opened 2018-01-06
Prevent creation of block dirs with zero samples

nipuntalukdar:


nipuntalukdar:
Made a small change that will prevent creation of block dirs with zero samples during compaction.

Gouthamve:
Instead of this can we return super early? Like right after the defer?

nipuntalukdar:
I think we cannot return early because before deferred function is called, the tmp directories might have been created. Also, the error will be very infrequent and hence shouldn't have any impact by allowing proper cleanup flow.

Gouthamve:
Oh no. https://github.com/prometheus/tsdb/pull/245/files#diff-43297cfd6441b52c54c50d646a564339R419 this is where we create the tmp directory. We removeAll before just to make sure but there should be no such temp dir.

krasi-georgiev:
superseded by: https://github.com/prometheus/tsdb/pull/374

PR #246, opened 2018-01-07
fixed an issue in TestWALRestoreCorrupted

nipuntalukdar:
Fixed a small issue with test case "TestWALRestoreCorrupted" as the test routine and cut method may update the same segment file concurrently and resulting in error sometime.

### DIFF
--- a/wal_test.go
+++ b/wal_test.go
@@ -371,9 +371,9 @@ func TestWALRestoreCorrupted(t *testing.T) {
 
 			testutil.Ok(t, w.cut())
 
-			//Sleep 2 seconds to avoid error where cut and test "cases" function may write or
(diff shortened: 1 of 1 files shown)


wal_test.go
```diff
@@ -370,6 +371,11 @@ func TestWALRestoreCorrupted(t *testing.T) {
 
 			testutil.Ok(t, w.cut())
 
+			//Sleep 2 seconds to avoid error where cut and test "cases" function may write or
```
Gouthamve: Space after `//` And the last sentence needs a `.`

nipuntalukdar:
Formatted comments.

fabxc:
👍

PR #248, opened 2018-01-08
Align series to 8/16 byte padding to increase addressable space

### LINKED ISSUE #238: Align series to 8/16byte padding to increase addressable space
https://github.com/prometheus/prometheus/issues/3190 is because we store the offset to series data in the index and we store in a `uint32` giving us only 4GB of addressable space.

To increase it, we can use an `uint64` but that would double the postings space required and compression would mean a good chunking mechanism.

Alternately, @fabxc came up with this brilliant idea of aligning the offsets to 8/16bytes and that would mean we just store the multiple of 8/16 bytes, giving us upto 64GB of addressable space.

shubheksha:
Fixes #238


index/index.go
```diff
@@ -955,15 +965,117 @@ func (t *serializedStringTuples) At(i int) ([]string, error) {
...
+// Decoder provides decoding methods for the v1 and v2 index file format.
 //
 // It currently does not contain decoding methods for all entry types but can be extended
 // by them if there's demand.
+type Decoder interface {
+	lookupSymbol(o uint32) (string, error)
+	SetSymbolTable(t map[uint32]string)
+	Postings(b []byte) (int, Postings, error)
+	Series(b []byte, lbls *labels.Labels, chks *[]chunks.Meta) error
+}
```
fabxc: We can probably skip making this an interface if `v2` and thus `DecoderV1` vs `DecoderV2` is not necessary.
But generally, exported interfaces should not have any unexported methods.

index/index.go
```diff
@@ -250,6 +250,7 @@ func (w *Writer) writeMeta() error {
 	return w.write(w.buf1.get())
 }
 
+//AddSeries adds the series one at a time along with its chunks.
```
fabxc: Missing space after `//`

block.go
```diff
@@ -151,6 +151,9 @@ type BlockMeta struct {
 
 	// Information on compactions the block was created from.
 	Compaction BlockMetaCompaction `json:"compaction"`
+
+	// Version of the index format
+	Version int `json:"version"`
```
fabxc: Generally 👍  for making this explicit here. But I'm wondering whether we actually need to bump the format. The serialisation is certainly different with the padding, but it seems like `DecoderV2` below doesn't do anything differently than `DecoderV1`.

shubheksha: Yup, fair enough. I'll just rename it to `Decoder` instead.

shubheksha: However, the main issue with having two different formats/versions was how to make them configurable. Currently, the value is being hardcoded as `1`. However, picking it up from `BlockMeta` itself makes the test fail as there is no value specified in the test files and the default value is `0`. How can that be fixed?

index/index.go
```diff
@@ -261,7 +266,9 @@ func (w *Writer) AddSeries(ref uint64, lset labels.Labels, chunks ...chunks.Meta
 	if _, ok := w.seriesOffsets[ref]; ok {
 		return errors.Errorf("series with reference %d already added", ref)
 	}
-	w.seriesOffsets[ref] = w.pos
+	w.addPadding(16)
+	w.seriesOffsets[ref] = w.pos / 16
+	w.Version = 2
```
Gouthamve: Redundant

index/index.go
```diff
@@ -852,9 +866,13 @@ func (r *Reader) LabelIndices() ([][]string, error) {
 	return res, nil
 }
 
-// Series the series with the given ID and writes its labels and chunks into lbls and chks.
+// Reads the series with the given ID and writes its labels and chunks into lbls and chks.
```
Gouthamve: In go the comments need to start with the function name. `Series reads the.....`

block.go
```diff
@@ -151,6 +151,9 @@ type BlockMeta struct {
 
 	// Information on compactions the block was created from.
 	Compaction BlockMetaCompaction `json:"compaction"`
+
+	// Version of the index format
```
Gouthamve: Fullstop after comment.

index/index.go
```diff
@@ -563,27 +572,32 @@ func (b realByteSlice) Sub(start, end int) ByteSlice {
 }
 
 // NewReader returns a new IndexReader on the given byte slice.
-func NewReader(b ByteSlice) (*Reader, error) {
-	return newReader(b, nil)
+func NewReader(b ByteSlice, v int) (*Reader, error) {
```
Gouthamve: Name accepted argument `version` everywhere. Will become much clearer.

Gouthamve:
Also, the docs need to be updated about the format.

index/index.go
```diff
@@ -563,27 +571,32 @@ func (b realByteSlice) Sub(start, end int) ByteSlice {
 }
 
 // NewReader returns a new IndexReader on the given byte slice.
-func NewReader(b ByteSlice) (*Reader, error) {
-	return newReader(b, nil)
+func NewReader(b ByteSlice, version int) (*Reader, error) {
+	return newReader(b, nil, version)
+}
+
+func NewReaderV1(b ByteSlice, c io.Closer, version int) (*Reader, error) {
```
Gouthamve: Do we need this function with the above definition already doing the same thing? I know I was leaning towards a `NewReaderV1` and `NewReaderV2`, but that doesn't look necessary right now.

Gouthamve: Also, if its `V1` why is it taking a version ;)?

index/index.go
```diff
@@ -563,27 +571,32 @@ func (b realByteSlice) Sub(start, end int) ByteSlice {
...
 	if err != nil {
 		return nil, err
 	}
-	return newReader(realByteSlice(f.Bytes()), f)
+	return newReader(realByteSlice(f.Bytes()), f, version)
 }
 
-func newReader(b ByteSlice, c io.Closer) (*Reader, error) {
+func newReader(b ByteSlice, c io.Closer, version int) (*Reader, error) {
 	r := &Reader{
```
Gouthamve: Check if the version is `1` or `2` and return an error if its neither?

fabxc:
Per my comment in the first review, do we really need to bump the version?

index/index.go
```diff
@@ -563,27 +571,32 @@ func (b realByteSlice) Sub(start, end int) ByteSlice {
...
-func NewReader(b ByteSlice) (*Reader, error) {
-	return newReader(b, nil)
+func NewReader(b ByteSlice, version int) (*Reader, error) {
+	return newReader(b, nil, version)
 }
 
 // NewFileReader returns a new index reader against the given index file.
-func NewFileReader(path string) (*Reader, error) {
+func NewFileReader(path string, version int) (*Reader, error) {
+	if version != 1 && version != 2 {
```
Gouthamve: Can we move this check into `newReader`? Because there is another caller above for which we are not doing any checks :)

PR #252, opened 2018-01-11
Count the total & failed numbers of block cutoffs

simonpasquier:
It happened on IRC that someone had a Prometheus instance not deleting blocks being past the retention period. Eventually it was a NFS ("cough cough") issue but the metrics could have at least confirmed that the retention wasn't working.

### DIFF
--- a/db.go
+++ b/db.go
@@ -122,6 +122,8 @@ type dbMetrics struct {
 	reloads              prometheus.Counter
 	reloadsFailed        prometheus.Counter
 	compactionsTriggered prometheus.Counter
+	cutoffs              prometheus.Counter
+	cutoffsFailed        prometheus.Counter
 	tombCleanTimer       prometheus.Histogram
 }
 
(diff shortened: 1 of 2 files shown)


db.go
```diff
@@ -148,6 +150,14 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
 		Name: "prometheus_tsdb_compactions_triggered_total",
 		Help: "Total number of triggered compactions for the partition.",
 	})
+	m.cutoffs = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "prometheus_tsdb_cutoffs_total",
```
Gouthamve: Can you name it `retention_cutoffs`?

db.go
```diff
@@ -277,7 +289,17 @@ func (db *DB) run() {
 	}
 }
 
-func (db *DB) retentionCutoff() (bool, error) {
+func (db *DB) retentionCutoff() (b bool, err error) {
+	defer func() {
+		if !b && err == nil {
+			// no data had to be cut off
```
brian-brazil: No data had to be cut off.

simonpasquier:
@Gouthamve @brian-brazil PTAL

fabxc:
👍

PR #254, opened 2018-01-13
Fix typos in comments

mattbostock:
a the -> the
timestmap -> timestamp
badded -> padded
its -> it is
callers -> caller's


brian-brazil:
Thanks!

PR #258, opened 2018-01-13
Add test for out-of-order datapoints

mattbostock:
Similar to the existing `TestAmendDatapointCausesError` test, assert
that `ErrOutOfOrderSample` is returned if we try to append a sample
older than the most recent sample for a given series.


brian-brazil:
That should return an error I'd say.

krasi-georgiev:
superseded in : https://github.com/prometheus/tsdb/pull/387

PR #261, opened 2018-01-16
Add block ULID to index reader errors

Gouthamve:
Not happy with this PR and is unlikely to be merged but this patch would unblock people who have issues with corrupt indexes.

@SuperQ

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/261)

### DIFF
--- a/block.go
+++ b/block.go
@@ -319,7 +319,7 @@ func (pb *Block) Index() (IndexReader, error) {
 	if err := pb.startRead(); err != nil {
 		return nil, err
 	}
-	return blockIndexReader{IndexReader: pb.indexr, b: pb}, nil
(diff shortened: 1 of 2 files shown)


fabxc:
We don't have the ULID available at the call sites?

Maybe we could at least exchange this for a wrapper `IndexReader` in the root package that attaches these rather than leaking this all across the code?

Gouthamve:
Yeah, that's better. I was thinking of making the querier do it, but couldn't find the right abstraction, I guess :) Let me do this.

Gouthamve:
@fabxc How about now?

fabxc:
👍

PR #262, opened 2018-01-17
Reduce index file size caused by symbol references by using indicies rather than offsets

### LINKED ISSUE #249: Reduce index file size caused by symbol references
As blocks get compacted, index files currently converge towards a size 5-8% compared to a block's chunk files.
This mostly true for relatively low series churn levels. As churn increaeses, there will be fewer chunks per series and per-series overheads for postings lists and series entries become more pronounced.

Postings compression is a known topic, but improving compression of series entries has not been looked into so far. A few strategies I've thought about:

Series entries hold label pairs in the form of offsets into the symbol table. The offsets are stored as varints. This was deemed important when we still accessed the symbol table through mmap on demand. Today however, we load all symbols into memory when loading a block for other reasons.

1. We could not store offsets but just incremental IDs for strings, which makes the numbers we reference a string with significantly shorter and they will therefore compress better. This requires storing. On startup we read the symbol table and just store a `<index>: <string>` mapping in memory rather than an `<offset>: <string>` mapping.
2. Currently the symbol table is sorted lexicographically. I believe there's no strict requirement for that. Instead it would be more efficient to order the table by frequency, i.e. how often a string appears in series. This way more frequent strings have a smaller index and references to them take up less space.
3. Lastly, we could think about having an additional table for label pairs, i.e. name/value combinations that make up a label. This would allow us in series to just store one reference per label instead of two. The extra table for label pairs should have roughly the same amount of entries like the symbol table, as each label value mostly is tied to a particular pair. It could generally be stored as a pair of references into the string table of course and therefore be pretty cheap.

Those are all possible to implement independently but probably roughly in this order. The first two should be very straightforward. 3 is a bit more involved.

cstyan:
fixes #249 

Left comments in sections where depending on whether we're using v1 or v2 all that would change is the var name (`offset` vs `index`) but not any of the functionality. 

Wasn't entirely sure if we want to change this struct: 
```
type hashEntry struct {
	keys   []string
	offset uint64
}
```

cc: @Gouthamve @fabxc

### DIFF
--- a/docs/format/index.md
+++ b/docs/format/index.md
@@ -43,7 +43,7 @@ Most of the sections described below start with a `len` field. It always specifi
 The symbol table holds a sorted list of deduplicated strings that occurred in label pairs of the stored series. They can be referenced from subsequent sections and significantly reduce the total index size.
 
 The section contains a sequence of the string entries, each prefixed with the string's length in raw bytes. All strings are utf-8 encoded.
-Strings are referenced by pointing to the beginning of their length field. The strings are sorted in lexicographically ascending order.
+Strings are referenced by sequential indexing. The strings are sorted in lexicographically ascending order.
 
 ```
 ┌────────────────────┬─────────────────────┐
--- a/index/index.go
+++ b/index/index.go
@@ -343,10 +343,7 @@ func (w *Writer) AddSymbols(sym map[string]struct{}) error {
 	w.symbols = make(map[string]uint32, len(symbols))
 
 	for index, s := range symbols {
-		w.symbols[s] = uint32(w.pos) + headerSize + uint32(w.buf2.len())
-		if w.Version == 2 {
-			w.symbols[s] = uint32(index)
-		}
+		w.symbols[s] = uint32(index)
 		w.buf2.putUvarintStr(s)
 	}
 
--- a/index/index.go
+++ b/index/index.go
@@ -273,17 +273,17 @@ func (w *Writer) AddSeries(ref uint64, lset labels.Labels, chunks ...chunks.Meta
 	w.buf2.putUvarint(len(lset))
 
 	for _, l := range lset {
-		offset, ok := w.symbols[l.Name]
+		index, ok := w.symbols[l.Name]
 		if !ok {
 			return errors.Errorf("symbol entry for %q does not exist", l.Name)
 		}
-		w.buf2.putUvarint32(offset)
+		w.buf2.putUvarint32(index)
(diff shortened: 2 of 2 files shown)


docs/format/index.md
```diff
@@ -43,19 +43,19 @@ Most of the sections described below start with a `len` field. It always specifi
...
-│ ├──────────────────────┴───────────────┤ │
+│ ┌──────────────────┬───────────────────┐ │
+│ │ 0 <uvarint>      │ str_1 <bytes>     │ │
+│ ├──────────────────┴───────────────────┤ │
 │ │                . . .                 │ │
-│ ├──────────────────────┬───────────────┤ │
-│ │ len(str_n) <uvarint> │ str_1 <bytes> │ │
-│ └──────────────────────┴───────────────┘ │
+│ ├──────────────────┬───────────────────┤ │
+│ │ n <uvarint>      │ str_1 <bytes>     │ │
```
fabxc: `len(str)` was  actually the number of bytes we have to subsequently read to know where the string ends. The field was not the offset. 

Looking at the code actually does not serilize the index (correctly) so the update of the docs here seems to be revertible?

index/index.go
```diff
@@ -341,8 +342,11 @@ func (w *Writer) AddSymbols(sym map[string]struct{}) error {
 
 	w.symbols = make(map[string]uint32, len(symbols))
 
-	for _, s := range symbols {
+	for index, s := range symbols {
 		w.symbols[s] = uint32(w.pos) + headerSize + uint32(w.buf2.len())
+		if w.Version == 2 {
```
fabxc: Unrelated to this change I suppose but do we really want the writer to be subject to versioning. I think we are fine with always have the writer write the most recent format version and just having the reader supporting multiple.

docs/format/index.md
cstyan: Ah okay I misread this then. We don't want to include a field for the index in this diagram?

index/index.go
cstyan: I guess this is dependent on how we intend to support different index file versions in the future. Is the intention to only allow older versions for read purposed with the intention that they can load in an old tsdb database with a newer version of the code and it would then be written to an index file again with the new format?

If we don't intend to allow writing of index files with old versions then I'll remove the version checks from the writer.

Gouthamve: We don't want to write older versions.

Gouthamve:
Thanks!

PR #269, opened 2018-01-26
Fix extra reloads.

Gouthamve:
Previously we reloaded every 1 minute irrespective of whether there are deleted blocks or not.

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/269)


brian-brazil:
:+1:

Issue #266, opened 2018-01-25
Clean Tombstones doesn't remove data from HeadBlock


gouthamve:
https://github.com/prometheus/prometheus/issues/3728 is because of this.

PR #270, opened 2018-01-28
Head: don't create stones, delete samples directly

codesome:
This should close [#266](https://github.com/prometheus/tsdb/issues/266) and [https://github.com/prometheus/prometheus/issues/3728](https://github.com/prometheus/prometheus/issues/3728)

/cc @Gouthamve


Gouthamve:
Can you add a test for this?

codesome:
Sorry, forgot to run the test before commit. Anyway, all tests assume that samples are not actually deleted after calling Delete(). I guess those tests need to be updated?

codesome:
@Gouthamve @fabxc 
So this is what I have done to take care of `sampleBuf` while clearing tombstones. (failed test for commit 61cc663 was because of not handling this)

As far as I understand `sampleBuf` holds the last 4 samples in that series. So after clearing tombstones I just update the `sampleBuf` using the last 4 samples in the series.

Example for visualisation:
(the numbers are only t from sample and not showing v)

Current samples: `0 1 2 3 4 5 6 7 8 9`
sampleBuf: `6 7 8 9`
representing as => mem: `0 1 2 3 4 5 6 7 8 9 | 6 7 8 9`

Tombstones: `{{2,3}, {4,7}, {0,1}}`

Deleting:

`{2,3}`
mem: `0 1 4 5 6 7 8 9 | 6 7 8 9`

`{4,7}`
mem: `0 1 8 9 | 0 1 8 9`

`{0,1}`
mem: `8 9 | 0 0 8 9` (first 2 samples in `sampleBuf` are empty)

Please let me know if what I have done is correct.

PS: Test for `head.CleanTombstones()` will be added soon.

head.go
```diff
@@ -601,9 +601,112 @@ func (h *Head) Delete(mint, maxt int64, ms ...labels.Matcher) error {
...
+			continue
+		}
+
+		cleaned := false // flag for current tombstone
+		stripeSeriesIndex := sid%stripeSize
+
+		h.series.locks[stripeSeriesIndex].Lock()
+		s := h.series.series[stripeSeriesIndex] // series with tombstones
+
+		if ms, ok := s[sid]; ok {
```
Gouthamve: Need to lock `ms` too.

head.go
```diff
@@ -601,9 +601,112 @@ func (h *Head) Delete(mint, maxt int64, ms ...labels.Matcher) error {
...
+	}
+
 	return nil
 }
 
+// CleanTombstones re-writes the chunks with tombstones
+// and deletes chunks if there is no samples after removing tombstones 
+func (h *Head) CleanTombstones() (bool, error) {
+
+	modified := false // flag for entire block
```
Gouthamve: Comments need to start with a capital and end with `.`

head.go
```diff
@@ -601,9 +601,112 @@ func (h *Head) Delete(mint, maxt int64, ms ...labels.Matcher) error {
...
+				it := &deletedIterator{it: chk.chunk.Iterator(), intervals: dranges}
+				for it.Next() {
+					ts, v := it.At()
+					app.Append(ts, v)
+				}
+
+				if newChunk.NumSamples() > 0 {
+					ms.chunks[i].chunk = newChunk
+				} else { // empty chunk, no required to store
+					ms.chunks = append(ms.chunks[:i], ms.chunks[i+1:]...)
```
Gouthamve: There is a race here? TODO(gouthamve): Explain with golang playground.

head.go
```diff
@@ -601,9 +601,112 @@ func (h *Head) Delete(mint, maxt int64, ms ...labels.Matcher) error {
...
+				}
+			}
+
+			// Getting the last 4 samples in the series
+			// to update the sampleBuf
+			nchks := len(ms.chunks)
+			tofill := 4 // samples left to fill in buffer
+			last4 := []sample{} // last 4 samples
+			for i := nchks-1; i>=0 && tofill>0; i-- {
```
Gouthamve: Why are you doing this for every chunk? Not just the last chunk?

head_test.go
```diff
@@ -299,6 +299,7 @@ Outer:
 
 		if len(expSamples) == 0 {
 			testutil.Assert(t, res.Next() == false, "")
+			head.Close()
```
Gouthamve: Check error here.

head_test.go
```diff
@@ -307,6 +308,7 @@ Outer:
 			testutil.Equals(t, eok, rok)
 
 			if !eok {
+				head.Close()
```
Gouthamve: Here too.

head_test.go
```diff
@@ -320,6 +322,122 @@ Outer:
...
+						}
+					}
+				}
+				return [4]sample{{0,0},{0,0},{0,0},{0,0}}
+			} ()
+
+
+		testutil.Equals(t, expSampleBuf, actSampleBuf)
+
+		head.Close()
```
Gouthamve: Need another test to check if the series is deleted if everything is delete. If the index doesn't have the label values...

head.go
codesome: Line 627 locks the series, and `ms` is obtained from that series. So why lock `ms` again?

head.go
codesome: Series is locked [here at line 627](https://github.com/prometheus/tsdb/pull/270/files#diff-0ced3454d447442a96a77b33ebe87300R627). Those operations are done before unlocking. So I think there is no race here. Or am I missing something?

head.go
codesome: I am iterating from last chunk till I get 4 samples or exhaust the chunks (and not checking all chunks all the time). I had currently taken as `sampleBuf` will have last 4 samples from the series and not just the last chunk. Looking into it now.

codesome: Seems like updating buffer only the last chunk is enough. I will make the change.

codesome:
@Gouthamve, need a review on this. The last test on 092c35a looks falky, need to restart the test again.

codesome:
@Gouthamve need a review. Test on 78b8cc8 looks flaky.

codesome:
@Gouthamve need a review. Test on 09ad2ed looks flaky.

codesome:
@Gouthamve as discussed offline, I have removed the 'tombstones' field from Head.

db.go
```diff
@@ -702,6 +702,7 @@ func (db *DB) Delete(mint, maxt int64, ms ...labels.Matcher) error {
 }
 
 // CleanTombstones re-writes any blocks with tombstones.
+// It also clears tombstones from in-mem block (head).
```
Gouthamve: Function needs updating.

head.go
```diff
@@ -598,9 +602,86 @@ func (h *Head) Delete(mint, maxt int64, ms ...labels.Matcher) error {
...
-	for _, s := range stones {
-		h.tombstones.add(s.ref, s.intervals[0])
+	if dirty {
+		h.gc()
+	}
+
+	return nil
+}
+
+// cleanTombstones re-writes the chunks with tombstones (deleted range)
```
Gouthamve: Should be named `delete()` instead.

head.go
```diff
@@ -598,9 +602,86 @@ func (h *Head) Delete(mint, maxt int64, ms ...labels.Matcher) error {
...
+		}
+
+		if newChunk.NumSamples() > 0 {
+			ms.chunks[i].chunk = newChunk
+		} else { // Empty chunk, not required to store.
+			chnksToDel = append(chnksToDel, i)
+		}
+	}
+	// Deleting empty chunks.
+	for i, idx := range chnksToDel {
```
Gouthamve: TODO(gouthamve): Review once again.

head_test.go
```diff
@@ -307,6 +308,7 @@ Outer:
 			testutil.Equals(t, eok, rok)
 
 			if !eok {
+				testutil.Assert(t, head.Close() == nil, "")
```
Gouthamve: You can use `testutil.Equals`

head_test.go
```diff
@@ -320,6 +322,129 @@ Outer:
...
+						}
+					}
+				}
+				return [4]sample{{0,0},{0,0},{0,0},{0,0}}
+			} ()
+
+		testutil.Equals(t, expSampleBuf, actSampleBuf)
+
+		if len(expSamples) == 0 {
+			testutil.Assert(t, len(head.values) == 0, "")
```
Gouthamve: `Equals` here too.

Gouthamve:
LGTM 👍 !

@fabxc Essentially this deletes data from memory instantly and removes any series with no chunks. This is required to tackle https://github.com/prometheus/prometheus/issues/3728.

I like removing the data from memory instantly rather than calling a seperate `CleanTombstones` to do it. I'll test this under load later tonight but this is good for a review now.

fabxc:
Mh, is this fixing an immediate bug? Otherwise I'd like to put it on hold until prometheus/tsdb is stable again.

Is prometheus/prometheus#3728 really an issue? On compaction the stale job labels should dropped eventually no?

codesome:
This can be an enhancement, rather than just fixing the issue of labels. As its cheaper to delete samples when in memory, we will avoid some disk writes while cleaning tombstones from persistent blocks (in some cases).

Anyway, better to hold this till 2.2.0 is released.

krasi-georgiev:
@fabxc do you predict  any problems with this change or should we move on with the reviews?

krasi-georgiev:
@codesome would you mind resolving the conflicts and I will review it.

bare in mind the changes in https://github.com/prometheus/tsdb/pull/378

codesome:
This happens to be my first PR to tsdb (also when I was just getting introduced to Prometheus), and I already found a small bug while loading WAL with this PR.

Resolving conflicts took some time, many things had changed. I will push once I clean up the code and verify once more.

krasi-georgiev:
ok ping me when ready for a review.

codesome:
ping @krasi-georgiev

krasi-georgiev:
thanks , I will review once we clear the other few pending PRs that were before this one.

head.go
```diff
@@ -367,6 +370,14 @@ func (h *Head) loadWAL(r *wal.Reader) error {
 	}
 	wg.Wait()
 
+	// Though we don't store tombstones anymore, we need this when user
+	// is migrating from the version where we used to store tombstones.
```
krasi-georgiev: This doesn't look right:

Imagine the following:
1. Delete some series from the head.
1. Stop Prometheus before the head is persisted into a block
1. Start Prometheus.

At this point Prometheus reads the wal so all data that was deleted in the head would be loaded again from the wal.

head.go
```diff
@@ -752,22 +762,89 @@ func (h *Head) Delete(mint, maxt int64, ms ...labels.Matcher) error {
...
 	}
-	for _, s := range stones {
-		h.tombstones.addInterval(s.ref, s.intervals[0])
+	// Deleting empty chunks.
+	for i, idx := range chnksToDel {
+		ms.chunks = append(ms.chunks[:idx-i], ms.chunks[idx-i+1:]...)
+	}
+
+	// Getting the last 4 samples in the last chunk
+	// to update the sampleBuf.
```
krasi-georgiev: I couldn't fully understand the logic behind `sampleBuf`(we need some better comments somewhere about this), but shouldn't this update happen only when we have some modification in the last chunk?
Track if the delete happened in the last chunk and only than update the buffer?

Ideally this should be in a separate func `updateSampleBuf`  as well.

codesome:
@krasi-georgiev I will address the reviews in couple of days, caught up with university work.

head.go
```diff
@@ -752,22 +762,89 @@ func (h *Head) Delete(mint, maxt int64, ms ...labels.Matcher) error {
...
+		return fmt.Errorf("Mem series not found for reference id %d", ref)
+	}
+	ms.Lock()
+	defer ms.Unlock()
 
-	if h.wal != nil {
-		if err := h.wal.Log(enc.Tombstones(stones, nil)); err != nil {
+	// Deleting samples for intervals in tombstones.
+	var chnksToDel []int
+	for i, chk := range ms.chunks {
```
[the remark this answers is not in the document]
codesome: That would call for another approach. Currently what I was doing was replacing the chunk in-place by removing deleted samples. Which would require access to the chunk slice and its index in the slice.

With `newChunkSeriesIterator` we can totally discard the old slice and create a new slice out of it. I will make the change.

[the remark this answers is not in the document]
codesome: There is also `append()` for `memSeries`. Should have know this and the above before, it would have saved lot of effort before. Thanks!

head.go
```diff
@@ -752,22 +762,89 @@ func (h *Head) Delete(mint, maxt int64, ms ...labels.Matcher) error {
...
+		dirty = true
 	}
-
 	if p.Err() != nil {
 		return p.Err()
 	}
-	var enc RecordEncoder
+	if dirty {
+		h.gc()
+	}
```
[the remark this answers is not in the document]
codesome: The symbols and values gets updated. So if some labels no longer exist after deleting, then it would be removed here.
It's tested here: https://github.com/prometheus/tsdb/blob/158f3c19bb5c7671c3ff18c225dd1d9bbb82c0e6/head_test.go#L400-L404

head.go
```diff
@@ -752,22 +762,89 @@ func (h *Head) Delete(mint, maxt int64, ms ...labels.Matcher) error {
...
+
+	stripeSeriesIndex := ref % stripeSize
+
+	h.series.locks[stripeSeriesIndex].Lock()
+	defer h.series.locks[stripeSeriesIndex].Unlock()
+	ms, ok := h.series.series[stripeSeriesIndex][ref]
+	if !ok {
+		return fmt.Errorf("Mem series not found for reference id %d", ref)
+	}
```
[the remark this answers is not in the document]
codesome: Now when I see it I think I can use that by removing lock on `h.series.locks[stripeSeriesIndex]`.

`h.series.getByID` acquires Read Lock for the series. But we have already acquired Write Lock here, so calling that would be a deadlock.

I previously thought that as we are modifying that memSeries, I need to use `h.series.locks` too. Now I see it is not required, lock on `ms` is enough.

head.go
codesome: Didn't think of that. I had removed writing of stones into the WAL after `Delete()`, I will bring it back.

head.go
codesome: I was just checking last 4 samples of the last chunk and updating it, that's what `sampleBuf` stores.

But now after the latest change we wont require this.

krasi-georgiev:
sure , ping me when ready for another review.

codesome:
@krasi-georgiev ping

head_test.go
```diff
@@ -295,92 +296,172 @@ func TestHeadDeleteSeriesWithoutSamples(t *testing.T) {
...
+			actSampleBuf := func() [4]sample {
+				for _, msmap := range head.series.series {
+					if len(msmap) > 0 {
+						seriesExists = true
+						for _, ms := range msmap {
+							return ms.sampleBuf
+						}
+					}
+				}
+				return [4]sample{{0, 0}, {0, 0}, {0, 0}, {0, 0}}
```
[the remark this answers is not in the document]
codesome: I did not understand this, what part can be simplified?

head_test.go
```diff
@@ -295,92 +296,172 @@ func TestHeadDeleteSeriesWithoutSamples(t *testing.T) {
...
-		remaint   []int64
+		intervals     Intervals
+		remaint       []int64
+		remainSampbuf []int64 // Sample buffer that should remain after deletion.
 	}{
 		{
-			intervals: Intervals{{0, 3}},
-			remaint:   []int64{4, 5, 6, 7, 8, 9},
+			intervals:     Intervals{{0, 3}},
+			remaint:       []int64{4, 5, 6, 7, 8, 9},
```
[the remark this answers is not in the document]
codesome: I think the current test is more understandable than adding a helper function and passing samples in the test cases, because currently the expected timestamps are clearly visible in the test cases.

codesome:
@krasi-georgiev as discussed over chat, we want to test the behaviour of the exposed APIs and not all unexposed API. 

So basically we want to remove `TestMemSeries_chunkMetas` and `TestMemSeriesReset`, and test the Querier and other exposed API to see if these don't cause any side affect.

Please correct me if I have missed something.

krasi-georgiev:
correct, the last behaviour we want to test is that when we delete samples from the head something that uses the buffer should behave as expected and in this case following the stack up it is the (db *DB) Querier.

krasi-georgiev:
@codesome lets do one final push to get this merged :smiley_cat:

codesome:
Was busy with some other PRs. Will get on this soon.

codesome:
@krasi-georgiev I would like to keep `TestMemSeriesReset`. Reason: When I merged this PR with master to remove conflicts, almost all delete tests failed (including `TestMemSeriesReset`), and apparently it was a bug in `reset()` itself. `TestMemSeriesReset` helped me debug it (just had to add `s.headChunk = nil`). 

So for the easy of debugging, I will keep that test and add more tests as discussed above.

krasi-georgiev:
ok

head_test.go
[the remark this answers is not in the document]
krasi-georgiev: that was a comment to myself :)

codesome:
ping @krasi-georgiev

codesome:
@krasi-georgiev do you think we should run benchmark on this in prometheus once to test some deletes manually on real setting too?

krasi-georgiev:
time to squash for the DCO.

> @krasi-georgiev do you think we should run benchmark on this in prometheus once to test some deletes manually on real setting too?

I would say do all the manual tests you can think of locally and that should be good enough.

codesome:
Instead of squashing, I reset the branch to before my first commit, synced with master, and manually applied the patch.

krasi-georgiev:
doesn't include the changes I proposed for the head test.

codesome:
No idea how that happened. Will add that patch too.

krasi-georgiev:
LGTM

ping @gouthamve for a final look.

krasi-georgiev:
It seems @gouthamve  won't be able to have a look any time soon. 
I suggest that you fix the conflict and will merge.

krasi-georgiev:
Thanks Ganesh!

if @gouthamve 	or @bwplotka 	find the time to look into this PR we will open a new PR for any suggested fixes.

Issue #145, opened 2017-09-13
add vendor to this project


timchenxiaoyu:
some package need to download ,please add vendor

fabxc:
In general this repository is a library and not a full application. A command line tool exists but in a very early state. In general, libraries should not have `vendor/` directories in Go.

Gouthamve:
Closing this as won't fix. As said above, this is a lib and its highly unlikely we will add a vendor folder. But if you have good reasons for a vendor folder, please feel free to re-open.

PR #271, opened 2018-01-29
Use `dep` to manage dependencies, but don't commit `vendor/` directory.

cofyc:
## What this PR does

Use [dep](https://github.com/golang/dep) to manage dependencies, but don't commit `vendor/` directory.

## Why we need it

1, With `dep` tool, it's easier to download all dependencies this library needs for newbies (like me) to learn `tsdb`. And using `vendor/` directory will not cause any possible conflicts in go workspace. Currently `Gopkg.toml` tracks master branch for dependencies which don't have stable version, or stable tag if they have a stable version. 

```
$ go get github.com/golang/dep/cmd/dep
$ dep ensure -v -vendor-only # populate `vendor/` directory from Gopkg.lock
$ go test github.com/prometheus/tsdb/...
```

2, I agreed that [it should not contains `vendor` directory](https://github.com/prometheus/tsdb/issues/145), as it may [cause problems](https://groups.google.com/forum/#!msg/golang-nuts/AnMr9NL6dtc/UnyUUKcMCAAJ). But we can commit dependency requirements (Gopkg.lock and Gopkg.toml). For dependers, they can know exact version requirements `tsdb` requires in case some dependencies introduced a bug or breaking change in new version.


fabxc:
👍 for using dep. I've used it elsewhere for a while now and it works out very nicely. Pinning dependencies to certain refs can be hugely helpful in practice too (e.g. prometheus/tsdb).

cofyc:
@fabxc 

FYI. I pushed a new commit (In https://github.com/prometheus/tsdb/pull/271/commits/6d9e1cc09e7f012c0a6af595429cc84a24196a56 ) to populate `vendor/` directory from Gopkg.lock before testing.

fabxc:
Oh, I didn't realise before that this was on prometheus/tsdb rather than prometheus/prometheus. With `dep` I suppose it's a bit less questionable for a library repository though.

krasi-georgiev:
closing as even if end up needing this it would be with go modules so will revisit if needed

Issue #244, opened 2018-01-05
Add metric for the space taken of the symbol tables


brian-brazil:
If a user has way too many label values this could get large, so having a metric indicating the space taken would be useful.

Gouthamve:
Should this be a prometheus metric or should it be part of the tsdb/promtool cli?

brian-brazil:
A metric of the TSDB itself, as Prometheus has no knowledge of this.

sipian:
If there is no one working on this, I would like to take this issue up.

sipian:
In index/index.go there is a function,  `func (r *Reader) SymbolTable() map[uint32]string`

Based on my understanding, to get the symbol table size we are supposed to call this function and get the total size of the `map[uint32]string`. For map total size, I found [this](https://stackoverflow.com/a/31855111/5730790) reference

Is this the correct direction in solving this issue?

brian-brazil:
Either in-memory or disk works, they're both basically the same size.

PR #272, opened 2018-01-31
add function to retreive index's section memory size and use this to get symbol table size

sipian:
[#244](https://github.com/prometheus/tsdb/issues/244)
I have added a function in index/index.go to get a section's size. 
Acc to documentation of [index.md](https://github.com/prometheus/tsdb/blob/master/docs/format/index.md) & [this](https://github.com/prometheus/tsdb/blob/master/index/index.go#L693) comment, a sections 1st 4 bytes holds the big endian encoded content length .

For symbol table size I retrieve these 4 bytes from (r.toc.symbols offsets)


index/index.go
```diff
@@ -743,6 +743,27 @@ func (r *Reader) decbufUvarintAt(off int) decbuf {
 	return dec
 }
 
+// getSectionSize returns the first 4 bytes in a section starting at offset off
+// to get the content length bytes
+func (r *Reader) getSectionSize(off int) int {
+	if off == 0 {
```
Gouthamve: You can remove this check.

index/index.go
```diff
@@ -743,6 +743,27 @@ func (r *Reader) decbufUvarintAt(off int) decbuf {
...
+		return 0
+	}
+	b := r.b.Range(off, off+4)
+	var l int
+	if l <= 0 {
```
Gouthamve: Umm, `l` has just been declared hence will always be 0.

index/index.go
```diff
@@ -743,6 +743,27 @@ func (r *Reader) decbufUvarintAt(off int) decbuf {
...
+		l = -1
+	} else {
+		l  = int(binary.BigEndian.Uint32(b))
+	}
+	return l
+}
+
+// getSymbolTableSize returns the bytes taken by the symbol table of Reader object
```
Gouthamve: end comments with `.`

sipian:
@fabxc @brian-brazil 

I have facing some difficulty in exposing symbol-table len to IndexReader interface.
The way I am doing is in **index.go** I've added a function GetSymbolTableSize() in struct Reader.
I need to call this function from each Block in db.block. Hence, I get IndexReader for each block(using Index()) and have added GetSymbolTableSize() in IndexReader interface in **block.go**.
But then we need to add implementation of GetSymbolTableSize() in head.go & querier_test.go() also.
Could you suggest a way to expose  GetSymbolTableSize() to IndexReader interface?

block.go
```diff
@@ -84,6 +84,9 @@ type IndexReader interface {
 	// LabelIndices returns a list of string tuples for which a label value index exists.
 	LabelIndices() ([][]string, error)
 
+	// GetSymbolTableSize returns the size occupied by the symbol table of Reader object.
+	GetSymbolTableSize() uint32
+
```
fabxc: This has nothing to do with functionality of an `IndexReader` and makes strong assumptions about how it is implemented. However we want to expose it, this interface isn't it.

sipian: I had the same question in this https://github.com/prometheus/tsdb/pull/297 PR.
There I have added a new interface `IndexExtra` for exposing LabelName() function.
Does this way work or is there any other way to expose functions from index.go ?

codesome:
@sipian, do you plan to work on this?

sipian:
@codesome 
You can go ahead

krasi-georgiev:
Superseded by https://github.com/prometheus/tsdb/pull/375

PR #273, opened 2018-02-02
Fix reader for large index files.

Gouthamve:
Currently the offsets are cast into uint32 even though the index can
grow larger than 4GiB.

Test is currently OOMing, will fix.

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/273)

### DIFF
--- a/index/index.go
+++ b/index/index.go
@@ -143,7 +143,11 @@ func NewWriter(fn string) (*Writer, error) {
 	if err != nil {
 		return nil, err
 	}
-	defer df.Close() // close for flatform windows
+	defer df.Close() // Close for platform windows.
+
+	if err := os.RemoveAll(fn); err != nil {
+		return nil, errors.Wrap(err, "remove any existing index at path")
+	}
 
 	f, err := os.OpenFile(fn, os.O_CREATE|os.O_WRONLY, 0666)
 	if err != nil {
@@ -530,8 +534,8 @@ type Reader struct {
 	c io.Closer
 
 	// Cached hashmaps of section offsets.
-	labels   map[string]uint32
-	postings map[labels.Label]uint32
+	labels   map[string]uint64
+	postings map[labels.Label]uint64
 	// Cache of read symbols. Strings that are returned when reading from the
 	// block are always backed by true strings held in here rather than
 	// strings that are backed by byte slices from the mmap'd index file. This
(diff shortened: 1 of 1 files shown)


Gouthamve:
Removed test as it OOMs. @fabxc take a look?

tzz:
Is there a workaround for users currently affected by this bug? Removing or splitting files maybe?

fabxc:
👍

SuperQ:
LGTM

index/index.go
```diff
@@ -143,7 +143,11 @@ func NewWriter(fn string) (*Writer, error) {
 	if err != nil {
 		return nil, err
 	}
-	defer df.Close() // close for flatform windows
+	defer df.Close() // Close for platform windows.
+
+	if err := os.RemoveAll(fn); err != nil {
+		return nil, errors.Wrap(err, "remove any existing index at path")
+	}
```
fabxc: Mh, do we need this? Just opening a new file in create/write mode on it should just overwrite it as expected.

deejay1:
This doesn't seem to make it into 2.2.0-rc.0, is there a chance for it to make it for final?

index/index.go
Gouthamve: Yep. So if there happened to be an index that is larger than the one being written it'll cause issues. Like one of my tests, I wrote a bigger one and then a smaller one, but couldn't figure out what's wrong.

Very rare edge case, but I guess it doesn't hurt to add a check.

fabxc:
👍

deejay1:
@fabxc Will this make it into 2.2.0 as it didn't make the cut to rc1?

fabxc:
Mh, what makes you think this didn't make it into rc.1? Is the underlying issue not resolved.

It _should_ be in: 
* https://github.com/prometheus/prometheus/blob/v2.2.0-rc.1/vendor/vendor.json#L805
* https://github.com/prometheus/tsdb/commit/494acd307058387ced7646f9996b0f7372eaa558

deejay1:
@fabxc Ah, stupid me, I was looking at the release/2.2 branch not the tag :/

tzz:
After trying Prometheus 2.2.0-rc.1 which has this fix, I can confirm that large index files are handled correctly so far (1 full day of testing). I have multiple 9-15GB index files and no errors. Thank you for making this fix available.

PR #276, opened 2018-02-07
Functions to get MinTime of DB and Block.

codesome:
As proposed/asked in #prometheus IRC, 
screenshot from 2018-02-07 11-43-02
I have added the function to get `MinTime` of the DB - MinTime of the oldest block.

This can be used in some endpoint.

### DIFF
--- a/block.go
+++ b/block.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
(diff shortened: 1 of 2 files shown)


Gouthamve:
See: https://github.com/prometheus/tsdb/pull/134

Still need this?

PR #277, opened 2018-02-07
Fix crash when a series has no block

simonpasquier:
Fixes https://github.com/prometheus/prometheus/issues/3788.

@Gouthamve I'm not sure how this can happen in practice but from a formal standpoint, some safety checks are missing. There's some overlap with @codesome's work on #270.

### DIFF
--- a/head.go
+++ b/head.go
@@ -587,8 +587,12 @@ func (h *Head) Delete(mint, maxt int64, ms ...labels.Matcher) error {
 	for p.Next() {
 		series := h.series.getByID(p.At())
 
+		t0, t1 := series.minTime(), series.maxTime()
+		if t0 == math.MinInt64 || t1 == math.MinInt64 {
+			continue
+		}
 		// Delete only until the current values and not beyond.
-		t0, t1 := clampInterval(mint, maxt, series.minTime(), series.maxTime())
+		t0, t1 = clampInterval(mint, maxt, t0, t1)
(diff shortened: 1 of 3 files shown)


Gouthamve:
👍 

This can happen when `head.gc()` is called and it removes all the chunks from a series. @fabxc

Gouthamve:
Closed by mistake, sorry.

head.go
```diff
@@ -587,8 +587,12 @@ func (h *Head) Delete(mint, maxt int64, ms ...labels.Matcher) error {
 	for p.Next() {
 		series := h.series.getByID(p.At())
 
+		t0, t1 := series.minTime(), series.maxTime()
+		if t0 < 0 || t1 < 0 {
```
fabxc: Why do we check `< 0` here but set `math.MinInt64` below. If we want to have the smallest number as a sentinel value checks should be precise for it as well.

That said, while we theoretically support timestamps below `0` practically it won't end well.

simonpasquier: Done

simonpasquier:
@fabxc @Gouthamve PTAL

kerneljack:
I was hoping this might have been merged by now. Any idea how long that might take? I raised the original issue in https://github.com/prometheus/prometheus/issues/3788

gouthamve:
LGTM! Thanks!

PR #283, opened 2018-02-14
don't bail for missing meta.js files

krasi-georgiev:
this change ensures that the db can be opened even when one or more folders are missing the meta.js file.
These are skipped when opening the db and deleted when db.reload is called.

### DIFF
--- a/compact.go
+++ b/compact.go
@@ -157,7 +157,8 @@ func (c *LeveledCompactor) Plan(dir string) ([]string, error) {
 	for _, dir := range dirs {
 		meta, err := readMetaFile(dir)
 		if err != nil {
-			return nil, err
+			// No need to return an error as this will be deleted when reloading the db.
+			continue
 		}
 		dms = append(dms, dirMeta{dir, meta})
 	}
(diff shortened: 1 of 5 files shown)


krasi-georgiev:
@fabxc is this what you had i mind?

krasi-georgiev:
:clapper: @fabxc , @Gouthamve just need a quick look and will do the tests. 
Also not sure if deleting folders with missing meta.js at retention time is ok?

compact.go
```diff
@@ -313,13 +314,15 @@ func (c *LeveledCompactor) Compact(dest string, dirs ...string) (uid ulid.ULID,
 	for _, d := range dirs {
 		b, err := OpenBlock(d, c.chunkPool)
 		if err != nil {
-			return uid, err
+			level.Error(c.logger).Log("msg", "couldn't open a block", "dir", d, "err", err.Error())
+			continue
```
fabxc: Did you check how deletion of blocks after compaction is handled? For example, if the caller just removes all input dirs if we return with no error, that would be kinda bad.

krasi-georgiev: yes I mentioned in the PR description that wasn't sure what should we do here.
Is the folder without a meta file unusable so just delete with some warning log?
I haven't looked at the code for the retention yet, I guess should do the same there, delete with some warning?

krasi-georgiev:
@fabxc PTAL

krasi-georgiev:
ping @Gouthamve

gouthamve:
Hmm this looks good to me but I'd be very cautious about adding this. Like a missing `meta.js` file means corrupted data-dir.

With this, we'll get missing data-ranges and a lot of user-queries as to why the storage is dropping data. I still think error-ing out makes sense rather than deleting data. We could add a command to tsdbtool/promtool that does this cleanup.

@krasi-georgiev If this is being accepted, could you add a metric for the number of blocks deleted due to corruption?

krasi-georgiev:
I think @fabxc mentioned that this was the behaviour at some point without any bad  consequences.

@gouthamve if we decide to go this road will add the metrics to the same PR.

krasi-georgiev:
closing in favour of #320  best to have this as a separate interface and be run explicitly in case of a data corruption.

PR #286, opened 2018-02-21
head: Rename highTimestamp to maxt

mattbostock:
`maxt` seems more consistent with `mint` and other uses of `maxt`
elsewhere in the code, if I've understand the intent correctly.


gouthamve:
@mattbostock Thanks! This makes sense!

PR #289, opened 2018-02-24
unlock too early, make other routine get an "unknown series" error

work-chausat:
s, created := a.head.getOrCreate(lset.Hash(), lset) in (a *headAppender) Add(lset labels.Labels, t int64, v float64) gives a not-nil s, but s := a.head.series.getByID(ref) in addfast says s is nil and gives an "unknown series" error

### DIFF
--- a/db.go
+++ b/db.go
@@ -765,7 +765,13 @@ func (db *DB) Snapshot(dir string, withHead bool) error {
 	if !withHead {
 		return nil
 	}
-	_, err := db.compactor.Write(dir, db.head, db.head.MinTime(), db.head.MaxTime())
+
+	head := &rangeHead{
+		head: db.head,
+		mint: db.head.MinTime(),
+		maxt: db.head.MaxTime(),
+	}
+	_, err := db.compactor.Write(dir, head, head.mint, head.maxt)
 	return errors.Wrap(err, "snapshot head block")
 }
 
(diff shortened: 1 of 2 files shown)


fabxc:
Thanks! Are you running into this often? This _should_ be a rare scenario, but I might have totally miscalculated this.

While the fix is generally right, I think it can lead to a deadlock. Currently `gc()` is the only thing actually locking stripes for ID and hash and it is never executed concurrently.
However, if another goroutine starts locking both, they might try to lock the same two stripes but in different order and consequentially block each other.

So to fix this, we probably need to think of some general change to how we are handling this.

fabxc:
It might well be we just want to "fail silentely" if the series is no longer there since it generally means it just got GCd because it didn't have data anymore anyway.

work-chausat:
Yes, you are right, it can really lead to a deadlock!!!
But if there are many goroutine call add, "unknown series" errors occur often!!!

work-chausat:
Yes, you are right, it can really lead to a deadlock!!!
But if there are many goroutines call add, "unknown series" errors occur often!!!
Do you think the way of locks1 [stripeSize]stripeLock for series and locks2 [stripeSize]stripeLock for hashes is an appropriate solution? And two locks give a better performance.

work-chausat:
I edit the code again, you can review it

PR #290, opened 2018-02-28
Improve usability of snapshots

fabxc:
@brian-brazil


brian-brazil:
:+1: 

I'll be interested to see how you word this in the docs, as we'll still want some freedom to change internals like this in the future.

PR #292, opened 2018-03-02
Make WAL no-op if flush interval is zero

mattbostock:
Allow users of the TSDB library to use the NopWAL by setting
WALFlushInterval to zero.

This can be desirable in situations where data integrity is safeguarded
by other mechanisms (e.g. by reading from a message bus), or when the
data being written is not valuable enough to warrant the I/O overhead
and storage used by the WAL.


brian-brazil:
I would take a value of 0 to mean that Sync wasn't being called, not that there was no WAL.

mattbostock:
@brian-brazil: You're right. Even then, overloading the flush duration seems too implicit. I'll rework it.

mattbostock:
Amended to add a new option, `NopWAL`, that defaults to false (so existing users do not need to make any changes).

I considered allowing users to pass in a custom WAL, but I think that's too low-level for this high-level interface.

fabxc:
Mh, WAL is an interface. Adding this switch allows us to chose between 2 predefined versions. That seems to defeat the purpose.
The field should probably be a WAL object.

mattbostock:
@fabxc: Amended, PTAL.

krasi-georgiev:
ping @mattbostock did you end up needing this after all?

PR #293, opened 2018-03-06
implement atomic mutex and use it.

krasi-georgiev:
before
>> completed stage=ingestScrapes duration=10.715800978s
 > total samples: 60000000
 > samples/sec: 5.599203370354241e+06

after

>> completed stage=ingestScrapes duration=9.638972787s
 > total samples: 60000000
 > samples/sec: 6.224723718283271e+06

a simplified version of the sync.Mutex which can e used across other places that we would need locking.

### DIFF
--- a/cmd/tsdb/main.go
+++ b/cmd/tsdb/main.go
@@ -29,6 +29,8 @@ import (
 	"text/tabwriter"
 	"time"
 
+	"github.com/prometheus/tsdb/mutex"
+
 	"github.com/go-kit/kit/log"
 	"github.com/pkg/errors"
 	"github.com/prometheus/tsdb"
@@ -150,7 +152,7 @@ func (b *writeBenchmark) run() {
 const timeDelta = 30000
 
 func (b *writeBenchmark) ingestScrapes(lbls []labels.Labels, scrapeCount int) (uint64, error) {
-	var mu sync.Mutex
+	var mu mutex.Atomic
 	var total uint64
 
 	for i := 0; i < scrapeCount; i += 100 {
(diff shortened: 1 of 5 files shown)


krasi-georgiev:
ping @Gouthamve

brian-brazil:
None of the code you're touching would affect ingestion rate, so I suspect the change you see is random deviation. A spinlock would also eat CPU when there are multiple API calls waiting on the lock.

fabxc:
Unfortunately I have to agree with Brian here though I believe the performance increase you are seeing is real.
It's just that the metric of samples/sec isn't relevant anymore because it's virtually impossible to hit the maximum in Prometheus with any realistic workload. The number of active series is the limiting factor across various other dimensions way before that.

OTOH, the benchmark is a very one-sided testing scenario and we don't know whether this mutex will have good general purpose performance in various real-world cases.

Lastly, the stdlib mutex is actually a hybrid mutex and has quite elaborate logic to try spinning first before actually pausing the goroutine: https://golang.org/src/sync/mutex.go#L72
Even if it's a bit slower in some cases, it seems hard to beat it on average across all scenarios.

Given the stability issues we had recently, this is probably not something we want to get into right now. Though I admittedly thought about it for `memSeries` – I'd certainly avoid just replacing all our mutexes (esp. long-held ones) with it.

krasi-georgiev:
Thanks for the feedback.

fabxc:
You probably want to keep that branch around – as I said it may still be interesting for the `memSeries` part once we are past our stability issues.

PR #298, opened 2018-03-13
compact: Additional test for issue prometheus#3943

Bplotka:
Really in debugging form just to figure out the problem now. [Not prepared to be merged]

TL;DR: Found bug - got repro of case when compaction is merging `36h` + `6h` block into `54h` with `12h` block gap DESPITE the fact 12h block was there! This is because we ignore last block in compaction - and it results of 2 block being compacted and USED without one 12h laying around and not being used. Nothing got's deleted. 

Fix in progress in separate PR

I tried to mimick the compaction flow from the logs provided by @hgranillo in https://github.com/prometheus/prometheus/issues/3943 

Further weirdness caused by the bug:
`level=info ts=2018-03-09T15:05:31.944625635Z caller=compact.go:394 component=tsdb msg="compact blocks" count=3 mint=1520402400000 maxt=1520596800000`
and couple of logs after:
`level=info ts=2018-03-09T17:02:51.604621507Z caller=compact.go:394 component=tsdb msg="compact blocks" count=2 mint=1520402400000 maxt=1520575200000`

How compaction can REDUCE block?

### DIFF
--- a/compaction_issue3943_test.go
+++ b/compaction_issue3943_test.go
@@ -1,0 +1,203 @@
+package tsdb
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/go-kit/kit/log"
+	"github.com/oklog/ulid"
+	"github.com/prometheus/tsdb/testutil"
+	"io/ioutil"
+	"math/rand"
+	"os"
+	"path"
+	"testing"
+	"time"
+)
+
+var (
+	testMinBlockSize = 2 * time.Hour
+	entropy          = rand.New(rand.NewSource(time.Now().UnixNano()))
+)
+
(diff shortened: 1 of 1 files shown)


Bplotka:
Ok so to 10th iteration - so far so good

fabxc:
Thanks. It's quite hard to say if it's not entirely known what blocks existed at what point.

But the 12h thing can happen if the 18h slice (modulo aligned by unix seconds) doesn't have any blocks for the first 6h, e.g. after startup. Then we'll just create the 12h block for the last 2/3 of the slice.

Bplotka:
OK so moving forward, until I can repro mentioned logs

Bplotka:
Closing in favor of https://github.com/prometheus/tsdb/pull/299

PR #299, opened 2018-03-13
compact: Excluded last block with shortest range instead of newest one by ULID.

Bplotka:
Fixes https://github.com/prometheus/prometheus/issues/3943 issue.

Added test cases.

### DIFF
--- a/compact.go
+++ b/compact.go
@@ -151,16 +151,11 @@ func (c *LeveledCompactor) Plan(dir string) ([]string, error) {
 	if err != nil {
 		return nil, err
 	}
-	// We do not include the most recently created block. This gives users a window
-	// of a full block size to piece-wise backup new data without having to care
-	// about data overlap.
 	if len(dirs) < 1 {
 		return nil, nil
 	}
-	dirs = dirs[:len(dirs)-1]
 
 	var dms []dirMeta
-
 	for _, dir := range dirs {
 		meta, err := readMetaFile(dir)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


compact.go
```diff
@@ -199,6 +198,20 @@ func (c *LeveledCompactor) plan(dms []dirMeta) ([]string, error) {
 	return nil, nil
 }
 
+// excludeFreshMeta removes newest block with the shortest range from given meta files.
+func excludeFreshMeta(dms []dirMeta) []dirMeta {
+	sort.Slice(dms, func(i, j int) bool {
```
Gouthamve: Why are we not comparing just the MinTimes?

Bplotka: Hm.. Yea - you might be right. Wanted to be too explicit here and not miss any edge case. This makes sense.

Bplotka: so actually it changes the logic completely - it's not the same. But maybe it is something we want. If we sort only by MinTimes in case:
```
		{
			// We have 20, 20, 20, 60, 60 range blocks.
			metas: []dirMeta{
				metaRange("1", 0, 20, nil),
				metaRange("2", 20, 40, nil),
				metaRange("3", 40, 60, nil),
				metaRange("4", 60, 120, nil),
				metaRange("5", 120, 180, nil),
			},
		},
```
We compact 1, 2, 3. It might be the case that 3 is generated from WAL dump (4 and 5 might be generated from compaction), but we still include 3 in the process. That's wrong.

The question is now, if such case make any sense. Maybe there is no possibility to have such situation.

Bplotka: I would actually compare MaxTimes here.. isn't that a true "Last block"?

Gouthamve:
:+1:

Bplotka:
Who's approval we need more to merge it? I guess some green flag from @fabxc ? We talked about this change set offline, but I guess his quick look might be useful here (:

fabxc:
I'll try to get to it after lunch over here

Bplotka:
@fabxc wonder if we should check if compaction level `==` 0 just in case ^^

PR #300, opened 2018-03-14
Close file for windows platform

Gouthamve:
Windows needs files that are being deleted to be closed.

Fixes: https://github.com/prometheus/prometheus/issues/3956

@fabxc @brian-brazil Should fix the mentioned issue, but we also need to enable tests on windows if we plan to catch bugs like these earlier.

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/300)


brian-brazil:
:+1:

PR #301, opened 2018-03-14
Use the right sized byte array for large indexes

Gouthamve:
Fixes: https://github.com/prometheus/prometheus/issues/3957

Not happy with this, suggestions welcome. @Bplotka

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/tsdb/301)

### DIFF
--- a/fileutil/mmap_windows.go
+++ b/fileutil/mmap_windows.go
@@ -19,14 +19,16 @@ import (
 	"unsafe"
 )
 
-func mmap(f *os.File, sz int) ([]byte, error) {
-	low, high := uint32(sz), uint32(sz>>32)
+const maxMapSize = 0xFFFFFFFFFFFF // 256TB
+
+func mmap(f *os.File, size int) ([]byte, error) {
+	low, high := uint32(size), uint32(size>>32)
 	h, errno := syscall.CreateFileMapping(syscall.Handle(f.Fd()), nil, syscall.PAGE_READONLY, high, low, nil)
 	if h == 0 {
 		return nil, os.NewSyscallError("CreateFileMapping", errno)
 	}
 
-	addr, errno := syscall.MapViewOfFile(h, syscall.FILE_MAP_READ, 0, 0, uintptr(sz))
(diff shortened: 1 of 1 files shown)


fileutil/mmap_windows.go
```diff
@@ -35,7 +35,17 @@ func mmap(f *os.File, sz int) ([]byte, error) {
...
-	return (*[1 << 30]byte)(unsafe.Pointer(addr))[:sz], nil
+	if sz < 1<<30 {
+		return (*[1 << 30]byte)(unsafe.Pointer(addr))[:sz], nil
+	}
+
+	if sz < 1<<32 {
+		return (*[1 << 32]byte)(unsafe.Pointer(addr))[:sz], nil
+	}
+
+	if sz <= 1<<34 {
```
brian-brazil: Why not always use this?

Gouthamve: You are allocating 16G everytime? Whether or not we are using it.

brian-brazil: Are you sure that's an allocation?

fileutil/mmap_windows.go
```diff
@@ -35,7 +35,7 @@ func mmap(f *os.File, sz int) ([]byte, error) {
 		return nil, os.NewSyscallError("CloseHandle", err)
 	}
 
-	return (*[1 << 30]byte)(unsafe.Pointer(addr))[:sz], nil
```
Bplotka: `sz` is just `size`? Can we name it like that? (maybe just for master)

Bplotka: Also I have stupid question. I have never used `mmap` before, but... why don't we cast to just `[sz]byte` here?

krasi-georgiev: I was thinking the same thing and tried it but casting to an array requires a constant for the array size as this is evaluated at compile time

-	return (*[sz]byte)(unsafe.Pointer(addr))[:sz], nil	=  non-constant array bound

fabxc:
👍 

tsdb needs another revendor then

PR #303, opened 2018-03-15
repair + compact: Improved logging for easier future debug purposes.

Bplotka:
This is based on my experience while debugging https://github.com/prometheus/prometheus/issues/3943.

I needed to deduct few things, and all that would be just bit easier with these two logs:
- new block's ULID on each compaction.
- actual list of Blocks (ulid + time range) on Prometheus startup (easy to log that while repairing blocks).

We don't really need blocks that takes part in compaction - that can be deducted easily based on time ranges of blocks we have currently in system.

Additionally, fixed comments.

What do you think @fabxc  @Gouthamve ?

### DIFF
--- a/compact.go
+++ b/compact.go
@@ -14,6 +14,7 @@
 package tsdb
 
 import (
+	"fmt"
 	"io"
 	"math/rand"
 	"os"
@@ -33,7 +34,7 @@ import (
 	"github.com/prometheus/tsdb/labels"
 )
 
-// ExponentialBlockRanges returns the time ranges based on the stepSize
+// ExponentialBlockRanges returns the time ranges based on the stepSize.
 func ExponentialBlockRanges(minSize int64, steps, stepSize int) []int64 {
 	ranges := make([]int64, 0, steps)
 	curRange := minSize
@@ -215,7 +216,7 @@ func (c *LeveledCompactor) selectDirs(ds []dirMeta) []dirMeta {
 
 	Outer:
 		for _, p := range parts {
-			// Donot select the range if it has a block whose compaction failed.
+			// Do not select the range if it has a block whose compaction failed.
 			for _, dm := range p {
 				if dm.meta.Compaction.Failed {
 					continue Outer
@@ -312,9 +313,12 @@ func compactBlockMetas(uid ulid.ULID, blocks ...*BlockMeta) *BlockMeta {
 // Compact creates a new block in the compactor's directory from the blocks in the
 // provided directories.
 func (c *LeveledCompactor) Compact(dest string, dirs ...string) (uid ulid.ULID, err error) {
-	var blocks []BlockReader
-	var bs []*Block
-	var metas []*BlockMeta
+	var (
+		blocks []BlockReader
+		bs     []*Block
+		metas  []*BlockMeta
+		uids   []string
+	)
 
 	for _, d := range dirs {
 		b, err := OpenBlock(d, c.chunkPool)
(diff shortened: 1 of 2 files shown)


compact.go
```diff
@@ -482,6 +480,14 @@ func (c *LeveledCompactor) write(dest string, meta *BlockMeta, blocks ...BlockRe
 	if err := renameFile(tmp, dir); err != nil {
 		return errors.Wrap(err, "rename block dir")
 	}
+
+	level.Info(c.logger).Log(
```
Bplotka: IMO we need to log on successful compaction only, right?

Gouthamve: Hmm, if it fails, it is good to know which blocks were involved. Is it worth printing out the blocks being compacted as well? And print out before the compaction?

Bplotka: Yea, my worry is that we log this and assumed everything is fine and the error is for some other compaction.. what do you think?

Shouldn't we make sure our `errors.Wrap` have all required info (what blocks are involved) in case of error?

I will end up passing here compacted block metas as well for this, but maybe that's ok.

compact.go
```diff
@@ -482,6 +480,14 @@ func (c *LeveledCompactor) write(dest string, meta *BlockMeta, blocks ...BlockRe
...
+		"msg", "compact blocks",
+		"count", len(blocks),
+		"mint", meta.MinTime,
+		"maxt", meta.MaxTime,
+		"ulid", meta.ULID.String(),
```
fabxc: You can leave out the `.String()` I think.

compact.go
```diff
@@ -482,6 +480,14 @@ func (c *LeveledCompactor) write(dest string, meta *BlockMeta, blocks ...BlockRe
 	if err := renameFile(tmp, dir); err != nil {
 		return errors.Wrap(err, "rename block dir")
 	}
+
+	level.Info(c.logger).Log(
+		"msg", "compact blocks",
+		"count", len(blocks),
```
fabxc: Maybe we should also just log all input blocks. A bit noisy but at least we can deduce the exact history in debugging cases.

compact.go
Gouthamve: Fair enough, yep :) Need to fix our errors.

compact.go
Bplotka: As mentioned in the description - I can deduct that easily given that I know what is the state on startup. But I guess  there is no harm in printing these here too.

fabxc: Yea, seems like a lot of mental work though compared to just having it printed right there.

Bplotka:
@Gouthamve  @fabxc Addressed comments. 

@Gouthamve I double checked and our wrap for `Compact` already include `planned`  dirs so they will be printed in case of error (:

Gouthamve:
```
level=info msg="write block" mint=1518912000000 maxt=1518919200000 ulid=01C8N2CB9RTXYTZ9JTWNK8DAD6
level=info msg="head GC completed" duration=1.34756ms
level=info msg="WAL truncation completed" duration=960ns
.
.
.
level=info msg="write block" mint=1518933600000 maxt=1518940800000 ulid=01C8N2CDSHRNX0C32QZTGFMSPM
level=info msg="head GC completed" duration=774.55µs
level=info msg="WAL truncation completed" duration=831ns
level=info msg="compact blocks" count=3 mint=1518912000000 maxt=1518933600000 ulid=01C8N2CDXQ77D06H5WGV3KD3G1 sources="unsupported value type"
```

Arrays are not automatically printed I guess.

Bplotka:
thanks for checking ): Was too sure about it!

Bplotka:
@Gouthamve  Sorry for that - I fixed it and tested log line manually.

Gouthamve:
np, Works perfectly now! :+1: :+1: :+1:

PR #304, opened 2018-03-15
Reuse xorAppender for appenders where possible.

brian-brazil:
Also don't allocate an xorIterator to iterate over an empty chunk.

This uses about the same CPU, but cuts allocs by 25% and garbage by 12%.

Before:
BenchmarkRangeQuery-4                  1        10925389973 ns/op 153998008 B/op   1768900 allocs/op
After:
BenchmarkRangeQuery-4                  1        10917168504 ns/op 135992712 B/op   1320118 allocs/op

@fabxc 
I was trying to speed up the setup steps of my PromQL benchmarks, and noticed this low hanging fruit.
This particular hacked-up benchmark includes some compactions.


fabxc:
Do you have some more background on that benchmark?
From an interface perspective, things become a bit more convoluted of course, which is generally not avoidable for memory reuse. So I'd like to understand whether this improves the numbers on a narrow set or whether the apply to effectively the entire write path?

brian-brazil:
The benchmark was only inserting data into the database, and I'd expect it to improve memory usage for writes generally - which includes the writes in compaction.

fabxc:
This will not affect compactions directly. In those we merely write completed chunk byte slices to disk, i.e. it will never touch the appender. Except for series that have deletions to be applied, which is rare one-off case and will usually be a tiny fraction.

brian-brazil:
Ah, I'd missed that. It's still a 25% reduction on allocs for tsdb writes.

fabxc:
Yes, definitely worth taking to a real world benchmark. It will help if we flip to a new chunk, which can happen at roughly the same time for many series. And at block boundaries (which are _not_ aligned with compaction), it will in fact happen for all series at once.
Surprisingly though, the usual memory pattern we have today does not show any intense memory spikes at block boundaries AFAICT.

brian-brazil:
Okay, I'll drop it for now then. This was one of the simpler alloc reductions in the tsdb.

Issue #260, opened 2018-01-16
Isolation for writes


Gouthamve:
Currently, when we call `app.Commit()`, we go ahead and start adding samples to the database. But now if there is a concurrent reader that is reading data, it might read some samples from the commit but not all.

This causes issues in Prometheus and has been known to make histograms falter: https://github.com/prometheus/prometheus/issues/1169

This PR is an attempt to fix it: https://github.com/prometheus/tsdb/pull/105

fabxc:
The issue with that PR was mostly that it broke apart our APIs relatively
forcefully to make it fit. I once had an attempt around that avoided this
and should in theory be a bit more memory friendly. However, that is more
than stale as well by now.

brian-brazil:
Isolation is a fairly core thing for a database, I'd be surprised if it didn't require a similar level of changes as what I proposed.

fabxc:
Note that the structure changed a fair bit and some things got simplified
into a single place since. That alone makes some leakage across the some
main interfaces no longer necessary.

PR #306, opened 2018-03-19
WIP: Isolation

Gouthamve:
A rebase of #105 

Tests are broken and cleanup pending.

### DIFF
--- a/block.go
+++ b/block.go
@@ -133,14 +133,14 @@ type BlockReader interface {
 // Appendable defines an entity to which data can be appended.
 type Appendable interface {
 	// Appender returns a new Appender against an underlying store.
-	Appender(writeId uint64) Appender
+	Appender(writeId, cleanupWriteIdsBelow uint64) Appender
 
 	// Busy returns whether there are any currently active appenders.
 	Busy() bool
 }
 
 type IsolationState struct {
-	// We will ignore all writes above the max, and that are incomplete.
+	// We will ignore all writes above the max, or that are incomplete.
 	maxWriteId       uint64
 	incompleteWrites map[uint64]struct{}
 	readId           uint64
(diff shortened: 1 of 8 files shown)


Gouthamve:
Prombench results under heavy query load:
[samples-s](https://user-images.githubusercontent.com/7354143/37877072-1300c1ac-3073-11e8-9b94-98d3721b767b.jpg)

[cpu-s](https://user-images.githubusercontent.com/7354143/37877068-0c076e6e-3073-11e8-98be-351036963d77.jpg)
[mem-s](https://user-images.githubusercontent.com/7354143/37877070-0f94c400-3073-11e8-9268-41d7e49a97bc.jpg)

Queries are marginally slower and almost unaffected.

Gouthamve:
@fabxc @brian-brazil DONE!

This is ready for review. Having a single Head and not multiple made it possible to have 0 public interface changes! I'll run this in prombench overnight for any surprises :)

block.go
```diff
@@ -130,10 +130,17 @@ type BlockReader interface {
 	Tombstones() (TombstoneReader, error)
 }
 
-// Appendable defines an entity to which data can be appended.
-type Appendable interface {
```
Bplotka: Isn't there an incompatibility issue with removal of public interface?

Gouthamve: Yes, but we don't use this interface in tsdb itself. I am not sure we want to provide any guarantees about stability yet.

Bplotka: Sure, but TSDB is a library. Anyone can depends on that including Prometheus itself. But I don't know, maybe the policy for TSDB is just to always vendor and pin to some version, so this is fine.

krasi-georgiev:
fixes #260

krasi-georgiev:
@gouthamve if you want to rebase and fix the conflicts I would try to review it quickly.

PR #307, opened 2018-03-21
Fix some megacheck and unconvert issues

mjtrangoni:
Hi @Gouthamve,

I fixed some trivial `megacheck` and `unconvert` issues.

### DIFF
--- a/block.go
+++ b/block.go
@@ -474,7 +474,7 @@ func (pb *Block) CleanTombstones(dest string, c Compactor) (bool, error) {
 	numStones := 0
 
 	pb.tombstones.Iter(func(id uint64, ivs Intervals) error {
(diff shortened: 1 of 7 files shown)


block.go
```diff
@@ -474,7 +474,7 @@ func (pb *Block) CleanTombstones(dest string, c Compactor) (bool, error) {
 	numStones := 0
 
 	pb.tombstones.Iter(func(id uint64, ivs Intervals) error {
-		for _ = range ivs {
+		for range ivs {
 			numStones++
 		}
```
fabxc: `numStones += len(ivs)`?

gouthamve: Yep, will open a PR for that.

PR #310, opened 2018-03-28
db: Fixed validateBlockSequence, exported it and added tests.

Bplotka:
`validateBlockSequence` had bug (nothing updated the `prev` var) so only overlap in firs two blocks was detected.

This changes fixes that and prints all overlaps that are in system.
IMO it is useful to have this kind of algorithm exported here that other projects can reuse for proper overlap detection (e.g: thanos)

### DIFF
--- a/db.go
+++ b/db.go
@@ -575,6 +575,10 @@ func validateBlockSequence(bs []*Block) error {
 }
 
 // OverlappingBlocks returns all overlapping blocks from given meta files.
+// We sort blocks by minTime. Then we iterate over each block minTime and treat it as our "current" timestamp.
+// We check all the pending blocks (blocks that we have seen their minTimes, but their maxTime was still ahead current
+// timestamp) if they did not finish. If not, it means they overlap with our current b. In the same time b is assumed as
+// pending.
 func OverlappingBlocks(bm []BlockMeta) (overlaps [][]BlockMeta) {
 	if len(bm) <= 1 {
 		return nil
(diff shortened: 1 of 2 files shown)


db.go
```diff
@@ -556,20 +555,93 @@ func (db *DB) reload(deleteable ...string) (err error) {
...
+		return overlaps
+	}
+
+	// Deduplicate cases like {a, b, c} {b, c} into just {a, b, c}
+	newOverlaps := [][]BlockMeta{overlaps[0]}
+	for i, overlap := range overlaps[1:] {
+		prev := overlaps[i]
+
+		// Check if prev contains all overlap elements.
+		found := false
```
Bplotka: This bad, will improve that soon

db.go
```diff
@@ -556,20 +555,93 @@ func (db *DB) reload(deleteable ...string) (err error) {
...
+	for _, b := range bs {
+		metas = append(metas, b.meta)
+	}
+
+	overlaps := OverlappingBlocks(metas)
+	if len(overlaps) == 0 {
+		return nil
+	}
+
+	return errors.Errorf("block time ranges overlap (%v)", overlaps)
```
Bplotka: I think better print here than just %v would do here

db.go
Bplotka: Fixed

Bplotka:
IMO rdy to be reviewed: @fabxc  @gouthamve

db_test.go
```diff
@@ -892,3 +892,52 @@ func expandSeriesSet(ss SeriesSet) ([]labels.Labels, error) {
...
+	testutil.Equals(t, [][]BlockMeta{{metas[1], o1}}, OverlappingBlocks(append(metas, o1)))
+
+	// o2 overlaps with 20-30 and 30-40.
+	o2 := BlockMeta{MinTime: 21, MaxTime: 31}
+	testutil.Equals(t, [][]BlockMeta{{metas[2], o2}, {o2, metas[3]}}, OverlappingBlocks(append(metas, o2)))
+
+	// o3a and o3b overlaps with 30-40 and each other.
+	o3a := BlockMeta{MinTime: 33, MaxTime: 39}
+	o3b := BlockMeta{MinTime: 34, MaxTime: 36}
+	testutil.Equals(t, [][]BlockMeta{{metas[3], o3a, o3b}}, OverlappingBlocks(append(metas, o3a, o3b)))
```
gouthamve: Hmm, a lot of complexity comes from achieving this, and I don't think there is any extra benefit to say:
`30-40` overlaps with `o3a` and `o3a` overlaps with `o3b`.

If we just check consecutive blocks, we simplify the algorithm a lot and still be able to catch overlapping blocks. I am not sure this complexity is worth it.

Bplotka: It is worth it, because we want 100% accurate algorithm for it for potential "repair" flows. I would vote for leaving it accurate as much as possible. The complexity is not that bad now.

fabxc: There's generally also always some annoyance involved if you return "first match" in a sequence of errors. People have to go in circles then of fixing first match, retrying, getting next match, fix, ...

gouthamve: Yep, instead of just returning the first match, we could return all the matches, I'm just thinking if returning:
`[{1, 2, 3}]` is any better than returning `[{1, 2}, {2, 3}]`. Also we would never be producing a third block if two are already overlapping.

All of the above code simplifies to:
```
prev := bm[0]
for _, b := range bm[1:] {
	if b.MinTime < prev.MaxTime {
		overlaps = append(overlaps, []BlockMeta{prev, b})
	}

	prev = b
}

return overlaps
```

Having said that I don't have a strong opinion here.

Bplotka: I think your algorithm will not catch all the overlaps (for example `[1, 2, 3, 4]` where `2` is overlapping `1, 2 and 3`). That's the problem. IMO either we do "is there any overlap" or we do accurate overlapping groups.. Anything in the middle will confuse while debugging potential issues from logs.

Bplotka: Basically mu rule of thumb is: what is more useful for debug readability and further implementations that will try to repair something?

IMO the grouped one, but I can see it's more complex

Bplotka: Actually to full power of this algo is visible on extreme cases. We had bug in thanos that we have 200 blocks for same time range. This will end up as single group will all 200 of them for my proposed algo. 

If we end up doing just pairs, we will see 200^2 pairs that will not help really to understand the trend/problem

Bplotka:
The offline idea we came is that we need to modify the returned representation to `map[TimeRange][]BlockMeta` were key contains the smallest range timeRange block that overlaps for this group.

Not sure if I can finish that today -and I will be unavailable until Wednesday (no internet access). Feel free to commandeer if needed.

db.go
```diff
@@ -556,20 +556,117 @@ func (db *DB) reload(deleteable ...string) (err error) {
...
+
+	return nil
+}
+
+type TimeRange struct {
+	Min, Max int64
+}
+
+// OverlappingBlocks returns all overlapping blocks from given meta files aggregated by overlaping range.
+// We sort blocks by minTime. Then we iterate over each block minTime and treat it as our "current" timestamp.
```
fabxc: That line onwards belong inside of the function. People reading the `godoc` do not care how it works internally.

db.go
```diff
@@ -556,20 +556,117 @@ func (db *DB) reload(deleteable ...string) (err error) {
...
+			}
+		}
+		overlapGroups[minRange] = overlap
+	}
+
+	return overlapGroups
+}
+
+// SprintOverlappedBlocks returns human readable string form of overlapped blocks.
+func SprintOverlappedBlocks(overlaps map[TimeRange][]BlockMeta) string {
```
fabxc: Maybe make it a dedicated `Overlaps` type and move this to its `String()` method?

Bplotka: yup

Bplotka:
Done. @fabxc PTAL

fabxc:
👍

PR #311, opened 2018-03-28
compact: [Debug- do not merge] dswarbrick compact issue - cannot repro

Bplotka:
As seen here in our local tests, I reproduced @dswarbrick state (thanks to his logs) on tsdb master (2.2.1), but there is no problem. @dswarbrick compactor however build wrong blocks which causes overlap.

I have no idea at that point why @dswarbrick could see this issue. This PR proves that it shouldn't be a case and that maybe there were some side effects impacting @dswarbrick Prometheus server. I have seen some disk fullness issues, maybe it's related? Any ideas?

Related to: https://github.com/prometheus/prometheus/issues/3943

### DIFF
--- a/compact_test.go
+++ b/compact_test.go
@@ -19,6 +19,9 @@ import (
 	"path/filepath"
 	"testing"
 
+	"fmt"
+	"time"
+
 	"github.com/go-kit/kit/log"
 	"github.com/pkg/errors"
 	"github.com/prometheus/tsdb/testutil"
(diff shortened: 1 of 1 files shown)


compact_test.go
```diff
@@ -154,153 +157,231 @@ func TestNoPanicFor0Tombstones(t *testing.T) {
...
-				metaRange("7", 360, 420, nil),
-				metaRange("8", 420, 540, nil),
+				metaRange("10", 1522051200000, 1522087200000, nil), // 10h
+				metaRange("13", 1522087200000, 1522108800000, nil), // 6h
+				metaRange("17", 1522108800000, 1522130400000, nil), // 6h
+				metaRange("20", 1522152000000, 1522159200000, nil),
+				metaRange("21", 1522130400000, 1522152000000, nil), // 6h
 			},
-			expected: []string{"7", "8"},
+			expected: []string{"13", "17", "21"},
```
Bplotka: This is important part proving that 2.2.1 works locally, but not for @dswarbrick

Bplotka:
All good - debian did not have fix under 2.2.1 tag

dswarbrick:
Thanks for your efforts and persistence @Bplotka

PR #316, opened 2018-04-13
db: Match comment case to function name

mattbostock:
`validateBlockSequence` is not exported, make the letter case in the comment match the function name.


brian-brazil:
Thanks!

PR #318, opened 2018-04-17
don't set current segment when can't write to it.

### LINKED ISSUE #317: WAL log samples: log series: write metrics/wal/000315: file already closed
This is something that happens when the disk is full. The caller should probably have a proper health check for that, but it's another story(https://github.com/prometheus/prometheus/issues/3283). tsdb should certainly not try to write to an already closed file.

This must come from:
https://github.com/prometheus/tsdb/blob/b7173eb0e5558c720ec4c939af7e803d38aa1cce/wal.go#L763 and the fact that in https://github.com/prometheus/tsdb/blob/b7173eb0e5558c720ec4c939af7e803d38aa1cce/wal.go#L604 the function returns early without setting `cur` to `nil` on failure.

I suspect that we should instead set `cur` to `nil` early in `cut()` and handle this case in `write()`.

cc: @Thib17

krasi-georgiev:
fixes #317

not really a bug so don't think we need any additonal tests for this.

### DIFF
--- a/wal.go
+++ b/wal.go
@@ -761,10 +761,12 @@ func (w *SegmentWAL) write(t WALEntryType, flag uint8, buf []byte) error {
 		}
 	}
 	n, err := w.writeTo(w.cur, w.crc32, t, flag, buf)
+	if err != nil {
+		return err
+	}
 
 	w.curN += int64(n)
-
-	return err
+	return nil
 }
 
 func (w *SegmentWAL) writeTo(wr io.Writer, crc32 hash.Hash, t WALEntryType, flag uint8, buf []byte) (int, error) {


gouthamve:
Umm, what is wrong with the current logic? The write could fail after writing a little data and the `n` returned reflects that.

krasi-georgiev:
I see what you mean.
in that case need to find another way to void getting the "file already closed" error

gouthamve:
Heh, it's the same thing as the current code :)

krasi-georgiev:
I think the difference is that w.curN remains nil , but will check it again.

krasi-georgiev:
there is a new WAL format to be merged so will revesit after that.

PR #319, opened 2018-04-25
repair: improve error message

simonpasquier:
This fixes https://github.com/prometheus/prometheus/issues/4112.

Here is the log message for a Prometheus server using this PR.
```
level=error ts=2018-04-25T09:03:59.359148734Z caller=main.go:593 err="Opening storage failed block dir: data/01CBVVTQBW3VDA8Z7R2CKMM3F1: unexpected end of JSON input"
```

### DIFF
--- a/repair.go
+++ b/repair.go
@@ -20,11 +20,11 @@ func repairBadIndexVersion(logger log.Logger, dir string) error {
 	// We must actually set the index file version to 2 and revert the meta.json version back to 1.
 	dirs, err := blockDirs(dir)
 	if err != nil {
-		return errors.Wrapf(err, "list block dirs %s", dir)
+		return errors.Wrapf(err, "list block dirs in %q", dir)
 	}
 
 	wrapErr := func(err error, d string) error {
-		return errors.Wrapf(err, "block dir: %s", d)
+		return errors.Wrapf(err, "block dir: %q", d)
 	}
 	for _, d := range dirs {
 		meta, err := readBogusMetaFile(d)
--- a/repair.go
+++ b/repair.go
@@ -5,12 +5,10 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
-	"path"
 	"path/filepath"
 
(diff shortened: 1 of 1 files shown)


fileutil/fileutil.go
```diff
@@ -5,10 +5,38 @@
...
+func isBlockDir(fi os.FileInfo) bool {
+	if !fi.IsDir() {
+		return false
+	}
+	_, err := ulid.Parse(fi.Name())
+	return err == nil
+}
+
+// BlockDirs returns the list of block directories in the given directory.
+func BlockDirs(dir string) ([]string, error) {
```
fabxc: The fileutil package aims to provide helper functionality around file system primitives. I'm not sure this is the right place to add TSDB domain knowledge.

simonpasquier: I had kind of the same feeling when I put it here but didn't find any other package where it would fit. Should it be in its own util package?

fabxc: What's the incentive to move it out of the top-level package to begin with?

repair.go
```diff
@@ -20,20 +18,18 @@ import (
 func repairBadIndexVersion(logger log.Logger, dir string) error {
 	// All blocks written by Prometheus 2.1 with a meta.json version of 2 are affected.
 	// We must actually set the index file version to 2 and revert the meta.json version back to 1.
-	subdirs, err := fileutil.ReadDir(dir)
+	dirs, err := blockDirs(dir)
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "list block dirs %s", dir)
```
juliusv: I'd prefer `%q` instead of `%s` here and below (especially here, if the dir contains spaces or such).

Should it read "list block dirs %q" -> "list block dirs in %q"?

simonpasquier: Fair point. Done.

juliusv:
:+1: (will have to wait until after moratorium for changes ends)

PR #320, opened 2018-05-02
[PROPOSAL] Scan and delete corrupted blocks.

krasi-georgiev:
I have seen few issues where tsdb corruption forces users to delete the entire data folder so I would like to add a new public function `tsdb.Scan` to run different scans for overlapping blocks , corrupted chunks , missing meta files etc.
At the end of the scan It will display a list and prompt if the user wishes to deleted these to allow a clean startup.

I was thinking to implement all this in the tsdb cli tool , but to avoid a lot of code duplication it would need access to many unexported funcs and structs so instead will add the  `Scan` func as part of the tsdb package and just call it from the tsdb cli.

Let me know for any objections or other ideas.


gouthamve:
This sounds good to me but a few things, can we fix the block by excluding corrupt data rather than delete it?

For example, the querier should return the data for everything except that part and also show an error for the corruption. Then people can go ahead, run this tool and cleanup the corrupted data rather than losing potentially days worth of data for a single corruption.

brian-brazil:
Partial data is virtually impossible to deal with correctly, so we would need to lose everything in that time slice to keep things sane.

krasi-georgiev:
I will add block deletion and can explore partial recovery if there are user requests

brian-brazil:
We could quarantine them like we did in 1.x

krasi-georgiev:
@brian-brazil can you point me tome some code or a branch for the quarantine? 
What can we do with the quarantined data after that?

@gouthamve Not fully tested and not ready for a review yet, but thought I would share the current progress.
At the moment it tests for 
 - unreadable meta files
 - blocks that can't be opened
 - overlapping blocks

can we test for anything else that has been a problem with the data corruption?

gouthamve:
Yep, we could start with this and then add cases as we people issues.

krasi-georgiev:
@gouthamve PTAL. let me know if you think some of this should be split in a separate PR.

krasi-georgiev:
in a separate PR I will update the README.md

krasi-georgiev:
@fabxc PTAL and will update the README as well.

Bplotka:
Can we scan for this as well? https://github.com/improbable-eng/thanos/blob/master/pkg/block/index.go#L154 It seems that people having "outsider" chunks (outside of meta time range) for TSDB produced blocks: https://github.com/improbable-eng/thanos/issues/354#issuecomment-394965024

krasi-georgiev:
that is very useful , thanks will add shortly.

krasi-georgiev:
@Bplotka added the IndexRepair. 

`go run cmd/tsdb/main.go scan /tmp/temp..FolderName`

you can use the test to generate some indexes with invalid chunks.
Just comment few lines in the test so it leaves the invalid indexes on the disk.
https://github.com/prometheus/tsdb/pull/320/files#diff-aff26b4b642d0f0234ee1cd4e8c492d2R176
this removes the the db after the test so commenting will leave it on your disk in the /tmp folder.

also comment all lines after
https://github.com/prometheus/tsdb/pull/320/files#diff-aff26b4b642d0f0234ee1cd4e8c492d2R209
this does the actual index repair and  we want to leave these un-repaired so you can test the tsdb scan tool

scan.go
```diff
@@ -0,0 +1,476 @@
...
+	for _, l := range postings.SortedKeys() {
+		if err := indexw.WritePostings(l.Name, l.Value, postings.Get(l.Name, l.Value)); err != nil {
+			return errors.Wrap(err, "write postings")
+		}
+	}
+
+	return nil
+}
+
+// removeOutsiders ensures order of the input chunks and keeps only the first of all overlaping chunkcs.
```
Bplotka: `s/chunkcs/chunks`

scan.go
```diff
@@ -0,0 +1,476 @@
...
+// we are not trying to be smart which chunk to keep so some data loss is to be expected.
+func removeOutsiders(chks []chunks.Meta, mint int64, maxt int64) []chunks.Meta {
```
Bplotka: method name is lying (: you don't remove outsiders only.

cmd/tsdb/main.go
```diff
@@ -58,16 +65,142 @@ func main() {
...
+			}
+		}
+		overlappingBlocks, err := scan.Overlapping()
+		if err != nil {
+			exitWithError(err)
+		}
+		if len(overlappingBlocks) > 0 {
+			didSomething = true
+			fmt.Println("Overlaping blocks.")
+			fmt.Println("Deleting these will remove all data in the listed time range.")
```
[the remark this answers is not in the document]
Bplotka: >> Not sure, wonder if this portion of code will not harm people (they will miss the fact that it is killing their metrics)

> you don't think the comment I added is enough?

I was rather saying that maybe user is not aware that he can recover data for some potential issues. Let me see the new code though.

krasi-georgiev:
> So far I am worried about 2 things:

> It might be tricky to do black box repair. That's why thanos bucket repair only known or safe to repair things: https://github.com/improbable-eng/thanos/tree/master/pkg/verifier
whereas here we just kill things which are broken. I guess it is fine for some use cases (people that do not care about missed metrics much and want to be just unblocked).
I would love to have single index repair logic for both Thanos and TSDB since we could use in Thanos just TSDB repair once we have it downloaded locally. But because the current logic is not repairing much, mostly killing unsure stuff (because we want to be generic here - we don't know the root cause of the issue that generated broken index), I won't be able to base on this code.

I agree on all that and would like to design it in a way that you can also use it so that we would have more people maintaining the code. Could you add comment to the code with the things that need to change to consider it useful for Thanos. 
if possible just describe what each function here is expected to do.

```
type Scanner interface {
	Tombstones() (map[string][]string, error)
	Meta() (map[string][]string, error)
	Indexes() (map[string][]string, []IndexStats, error)
	Overlapping() (Overlaps, error)
}
```


>I guess this tool can go in 2 direction.

> white box: so as Thanos verify package - we add repair step on every known issue
black box: just remove "unsure" corrupted data.
I can see totally the black box tool usefulness and simplicity and maybe we want this. But I have seen people asking questions "can I recover this data", so I would really be explicit here about the approach this tool take.

Isn't  this is what I have more or less with the `Scanner interface`? 
Repair what is possible and return unrepairable pieces and than the cli tool prompts if these should be deleted. My idea would be to make the interface white box, but the cli tool will also offer to unblock users by deleting the unrepairable pieces. 
In your case I guess you can just use the interface and implement your own cli tool. 
So in short is it important to make the interface usable by different cli tools.

> Also one thing that will help here is to separate SCAN step from Repair/delete as we do in Thanos. This way you can safely run this and gather all issues and see how much is broken and how. Then it easier to decide if you want to dig and repair, or maybe only 2h is corrupted so we are fine to delete those.

Can you include some examples as I am not sure what you mean here.

krasi-georgiev:
ping @bwplotka

krasi-georgiev:
ping @bwplotka

cmd/tsdb/main.go
```diff
@@ -58,16 +65,185 @@ func main() {
...
+				exitWithError(errors.Wrap(err, "deleting temp files"))
+			}
+		}
+	}
+}
+
+func dellAll(paths []string) error {
+	for _, p := range paths {
+		if err := os.RemoveAll(p); err != nil {
+			return fmt.Errorf("error deleting: %v, %v", p, err)
```
bwplotka: I though `errors.Wrap` is preferable, but not sure what convention you want to follow everywhere

db.go
```diff
@@ -534,8 +534,9 @@ func (db *DB) reload() (err error) {
 	sort.Slice(blocks, func(i, j int) bool {
 		return blocks[i].Meta().MinTime < blocks[j].Meta().MinTime
 	})
-	if err := validateBlockSequence(blocks); err != nil {
```
bwplotka: I am not sure about changing anything in `db.go` before resolving conflicts (: I would aim for changing as little as possible. (e.g `-type Overlaps map[TimeRange][]BlockMeta` -> `type Overlaps map[TimeRange][]*Block` change)

db.go
```diff
@@ -556,45 +556,26 @@ func (db *DB) reload(deleteable ...string) (err error) {
...
-}
-
 // TimeRange specifies minTime and maxTime range.
 type TimeRange struct {
 	Min, Max int64
 }
 
 // Overlaps contains overlapping blocks aggregated by overlapping range.
-type Overlaps map[TimeRange][]BlockMeta
+type Overlaps map[TimeRange][]*Block
```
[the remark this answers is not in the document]
bwplotka: Why not using `meta.ULID?`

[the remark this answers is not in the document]
krasi-georgiev: I don't see any harm in this refactoring as `validateBlockSequence` is doing the same thing.

krasi-georgiev:
ping @bwplotka

mmencner:
I just hit the issue of overlapping blocks with prometheus 2.6.0 running with local storage, inside a docker container: 
```
err="opening storage failed: invalid block sequence: block time ranges overlap: [mint: 1551146400000, maxt: 1551153600000, range: 2h0m0s, blocks: 597]:
```
When it happened, a number of new directories were created, each time the container was trying to start (unsuccessfully). Only after deleting all of them, we were able to get the prometheus running.

In this case, auto-check/auto-discard would be really helpful.

PR #325, opened 2018-05-08
feat cmd/tsdb: added human readable print for timestamps on block list

FUSAKLA:
@krasi-georgiev 
As you suggested the second PR containing the _human readable_ flag for TSDB cmd client.


krasi-georgiev:
LGTM :zap: :+1:

gouthamve:
Looks good, thanks! Let's keep the timestamp as default for now.

PR #326, opened 2018-05-08
Seek() shouldn't return true when past maxt

free:
`Seek()` shouldn't return `true` when past `maxt`. Also add some tests to show that `Querier.Select()` may return series with no values.


free:
Superseded by #327 due to misguided DCO requirement.

PR #327, opened 2018-05-08
Seek() shouldn't return true when past maxt. Also add some tests to s…

free:
`Seek()` shouldn't return `true` when past `maxt`. Also add some tests to show that `Querier.Select()` may return series with no values.

Signed-off-by: Alin Sinpalean <alin.sinpalean@gmail.com>

### DIFF
--- a/querier.go
+++ b/querier.go
@@ -804,6 +804,9 @@ func (it *chunkSeriesIterator) Seek(t int64) (ok bool) {
 
 	for it.cur.Next() {
 		t0, _ := it.cur.At()
+		if t0 > it.maxt {
+			return false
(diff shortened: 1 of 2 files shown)


free:
Was trying to add a `SeekBefore()` to find the first value __before__ time `t` (which is what `Seek()` should do by default, I believe) and found a number of problems with the `Seek()` implementation.

This is one of the problems. The other is that `Seek()` will actually seek backwards within the same chunk which is (1) broken (even though never relied upon AFAICT) and (2) inefficient because it will start scanning from the beginning of the chunk. I'll address that separately.

PR #329, opened 2018-05-09
Fix issues (1), (2) and (3) described in #328

### LINKED ISSUE #328: SeriesIterator.Seek() is vaguely defined
Context: I was looking into cleaning up the iterator code, as I'm trying to add a separate `SeekBefore()` method, that would return the value at time `t` as defined by Prometheus without always parsing and iterating over a whole chunk covering the previous 5 minutes (when the looked up value is toward the beginning of a chunk).

However, I ran into some issues regarding how chunked series and iterators should actually work. There are a couple of gray areas that are only partially covered by comments or test code, making it impossible to figure out what is the expected correct behavior:

1. `blockQuerier` and the resulting `chunkedSeries` and `chunkSeriesIterator` all have `mint` and `maxt` fields, which (along with `(*chunkSeriesIterator).Seek()` implementation details) suggest that iterators should only return samples with timestamps `mint <= t <= maxt`. However, there are bugs (such as the one I'm attempting to fix with #327) and even [confusing/confused unit tests](https://github.com/prometheus/tsdb/blob/54205903f79fdff71d52682b09dd760259bca201/querier_test.go#L888) -- which appear to create an iterator with `mint = 5`, `maxt = 8` and expect `(*chunkSeriesIterator).Seek()` to return a value with timestamp `3`.<p>Are`mint` and `maxt` supposed to be "hard" limits on the respective iterators or only on which chunks are selected?
2. I gather (from the implementation again) that `SeriesIterator.Seek()` is only supposed to seek forward (which doesn't work as expected, as it will also seek backward within the same chunk). Is this assumption correct?
4. Is `Seek()` supposed to always advance the iterator or should a second `Seek()` call with the same timestamp as the immediately preceding one leave the iterator positioned where it was before?
3. Is it actually a good idea for a `Seek()` with a timestamp `t < mint` to ever succeed? If so a poorly defined series e.g. `[now - 10m, now]` will "successfully" produce a value of the series at `now - 1y` equal to the first value after `now - 10m`.

I'm glad to do the work (including properly documenting the code) regardless of what the answers are, I'd just like to get some "authoritative" input. Thanks.

free:
Fix issues (1), (2) and (3) described in #328

closes #328

Signed-off-by: Alin Sinpalean <alin.sinpalean@gmail.com>


querier.go
```diff
@@ -791,31 +794,48 @@ func (it *chunkSeriesIterator) Seek(t int64) (ok bool) {
...
 	}
 
-	it.cur = it.chunks[it.i].Chunk.Iterator()
-	if len(it.intervals) > 0 {
-		it.cur = &deletedIterator{it: it.cur, intervals: it.intervals}
-	}
+	iCur := it.i
+	for ; it.i < len(it.chunks); it.i++ {
+		// Skip chunks with MaxTime < t.
+		if it.chunks[it.i].MaxTime >= t {
```
gouthamve: This can be simplified to:
```
if it.chunks[it.i].MaxTime >= t {
  continue
}
// Don't reset the iterator unless we've moved on to a different chunk.
if it.i != iCur {
	it.cur = it.chunks[it.i].Chunk.Iterator()
	if len(it.intervals) > 0 {
		it.cur = &deletedIterator{it: it.cur, intervals: it.intervals}
	}
}

for it.cur.Next() {
	t0, _ := it.cur.At()
	if t0 > it.maxt || it.cur.Err() != nil {
		return false
	}
	if t0 >= t {
		return true
	}
}
```

This reduces the indentation and makes it easier to read.

querier_test.go
```diff
@@ -987,7 +1026,7 @@ func TestSeriesIterator(t *testing.T) {
 						{10, 22}, {203, 3493},
 					},
 
-					seek:    203,
+					seek:    101,
```
gouthamve: Why would this return `false`?

free:
Ping?

fabxc:
@gouthamve ^ can you take a look at this?

fabxc:
@free thanks for all the analysis in the issue – seems we've missed quite a few things.
In (1) you mention a particular test that even just adapts the expected result to pass. If this PR fixes (1) though, why is the test not fixed but still passes?

free:
[Apologies for dropping this, GitHub (or some mail server) decided I'm probably not interested in email notifications from this PR.]

> In (1) you mention a particular test that even just adapts the expected result to pass. If this PR fixes (1) though, why is the test not fixed but still passes?

Seems like that particular test was actually only confusing rather than actually broken: the test code at [line 1066](https://github.com/free/tsdb/blob/d741a3eeadb3241b8733b30fc8d533c6745939ea/querier_test.go#L1066) actually filters out samples that fall outside `[mint, maxt]` from the expected output (when testing a chunked iterator), but not below (when testing a chained iterator, which is supposed to return the `t = 3` value).

I'd rather not go down the rabbit hole of rewriting the unit tests, though. I do have a day job that doesn't involve as much Prometheus as I might otherwise like.

alin-amana:
@gouthamve @fabxc: Ping?

querier.go
```diff
@@ -791,31 +794,50 @@ func (it *chunkSeriesIterator) Seek(t int64) (ok bool) {
 		t = it.mint
 	}
 
-	for ; it.chunks[it.i].MaxTime < t; it.i++ {
-		if it.i == len(it.chunks)-1 {
-			return false
-		}
+	// Return early if already past t.
+	if t0, _ := it.cur.At(); t0 >= t {
```
free: Just realized that this might call `At()` on an iterator that `Next()` or `Seek()` might not have been called on. It works if one assumes all timestamps to be strictly positive and that the iterator will return a timestamp of 0 when uninitialized.

Should I remove it?

free: Well, crap. This is the thing that makes `Seek()` not advance the iterator when seeking for a timestamp that's in the past (compared to the iterator's position).

So it looks to me like we need to change the contract of either `chunkenc.Iterator.At()` to return something like `math.MinInt64`; or that of `tsdb.SeriesIterator.Seek()` to always advance the iterator. The latter seems more radical.

free: I've changed `chunkenc.Iterator.At()` to always return `math.MinInt64` before it has advanced (and documented it as such). Everything is good in the world again.

querier_test.go
```diff
@@ -1116,6 +1155,24 @@ func TestChunkSeriesIterator_DoubleSeek(t *testing.T) {
 	testutil.Equals(t, float64(2), v)
 }
 
+// Regression for: https://github.com/prometheus/tsdb/pull/327
```
krasi-georgiev: maybe point this to https://github.com/prometheus/tsdb/issues/328

To be honest I think the test itself explains the desired behaviour so don't think we even need to link it to the issue.

querier.go
```diff
@@ -783,6 +783,9 @@ func newChunkSeriesIterator(cs []chunks.Meta, dranges Intervals, mint, maxt int6
 
 func (it *chunkSeriesIterator) Seek(t int64) (ok bool) {
 	if t > it.maxt {
+		it.i = len(it.chunks)
+	}
+	if it.cur.Err() != nil || it.i >= len(it.chunks) {
 		return false
 	}
```
krasi-georgiev: so here we are first advancing and than failing. Normally my logic would be that if an operation failed it shouldn't change the internal state.

querier_test.go
free: Removed.

querier.go
free: We're not failing, we're simply advancing to the requested timestamp and returning `false`, meaning there are no more values.

If we would not advance, then a `Seek` before `maxt` following a `Seek` past `maxt` would incorrectly position the iterator on a valid value.

alin-amana:
Any chance of getting this merged?

Although please note that due to the requirement that `Seek()` not advance the iterator unless necessary, I have had to change the definition of `chunkenc.Iterator.At()` (and the behavior of its 2 implementations) to return `math.MinInt64` before the iterator is advanced.

querier.go
```diff
@@ -791,31 +795,50 @@ func (it *chunkSeriesIterator) Seek(t int64) (ok bool) {
...
+			if t0 >= t {
+				return true
+			}
 		}
 	}
 	return false
 }
 
 func (it *chunkSeriesIterator) At() (t int64, v float64) {
+	// Should it panic if it.cur.Err() != nil || it.i >= len(it.chunks)?
```
gouthamve: This comment can be removed.

querier.go
gouthamve: Yep, before the contract used to be that `At()` shouldn't be called before `Next()` or `Seek()` and it shouldn't be called after they return `false`.

I like the newer contract of returning `math.MinInt64` for non-initialised `Iterator`s.

querier.go
free: Right, it is now OK to call this at any time. Done.

querier_test.go
```diff
@@ -448,6 +458,18 @@ func TestBlockQuerier(t *testing.T) {
 					),
 				}),
 			},
+			{
```
codesome: This test seems to be working even without this PR. Any reason for adding?

free:
On Wed, Sep 12, 2018 at 11:47 AM Ganesh Vernekar <notifications@github.com>
wrote:

> *@codesome* commented on this pull request.
> ------------------------------
>
> In querier_test.go
> <https://github.com/prometheus/tsdb/pull/329#discussion_r216965492>:
>
> > @@ -448,6 +458,18 @@ func TestBlockQuerier(t *testing.T) {
>  					),
>  				}),
>  			},
> +			{
>
> This test seems to be working even without this PR. Any reason for adding?
>
[Replying by email because I cant, for the life of me, find this comment on
GitHub.]

I don't remember specifically, but if I had to guess it's (a) because of a
bug I had introduced during my change and wanted to verify whether I had
then fixed it; and (b) because it's a use case that's not already covered
by the other test cases.

querier_test.go
```diff
@@ -466,8 +488,8 @@ func TestBlockQuerierDelete(t *testing.T) {
 			},
 		},
 		tombstones: &memTombstones{intvlGroups: map[uint64]Intervals{
-			1: Intervals{{1, 3}},
-			2: Intervals{{1, 3}, {6, 10}},
+			1: Intervals{{1, 2}},
+			2: Intervals{{2, 3}, {6, 10}},
```
krasi-georgiev: Why is this change required?

krasi-georgiev:
ping @free

free:
@krasi-georgiev, as things stand right now, it looks like I have about one more week of heads-down effort on my work project. But I will get back to this as soon as I'm done with 14 hour days. Apologies.

krasi-georgiev:
@free 	no problem , thanks for letting me know.

krasi-georgiev:
@free is this ready for another review? Not sure if the comments I left are still relevant after the latest changes.

krasi-georgiev:
> Nope, not yet,

ping me when ready for a review and I will focus for a proper review.

chunkenc/chunk.go
```diff
@@ -64,6 +65,7 @@ type Appender interface {
 
 // Iterator is a simple iterator that can only get the next value.
 type Iterator interface {
+	// At returns (math.MinInt64, 0.0) before the iterator has advanced.
```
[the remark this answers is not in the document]
bwplotka: adding this https://github.com/prometheus/tsdb/pull/665

querier.go
```diff
@@ -658,7 +658,7 @@ func (s *chunkSeries) Iterator() SeriesIterator {
 type SeriesIterator interface {
 	// Seek advances the iterator forward to the given timestamp.
 	// If there's no value exactly at t, it advances to the first value
-	// after t.
+	// after t. Has no effect if already past t.
```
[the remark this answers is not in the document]
bwplotka: Added to https://github.com/prometheus/tsdb/pull/665 as well

querier.go
bwplotka: Agree with @free  and adding exactly same thing to https://github.com/prometheus/tsdb/pull/665

querier.go
```diff
@@ -780,21 +784,34 @@ func (it *chunkSeriesIterator) Seek(t int64) (ok bool) {
...
+		// Skip chunks with MaxTime < t.
+		if it.chunks[it.i].MaxTime < t {
+			continue
+		}
 
-	for it.cur.Next() {
-		t0, _ := it.cur.At()
-		if t0 >= t {
-			return true
+		// Don't reset the iterator unless we've moved on to a different chunk.
```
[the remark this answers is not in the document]
bwplotka: adjuste https://github.com/prometheus/tsdb/pull/665 as well, although code looks different (less indents).

bwplotka:
Ok, I added all the improvements from here to https://github.com/prometheus/tsdb/pull/665 so closing this. Also we are merging https://github.com/prometheus/prometheus/pull/5805 so I need to redo my PR anyway as well on Prometheus later on.

cc @krasi-georgiev @free @codesome

PR #330, opened 2018-05-14
add rwmutex to prevent concurrent map read when delete series

codwu:
fix https://github.com/prometheus/prometheus/issues/4154
@krasi-georgiev

### DIFF
--- a/block.go
+++ b/block.go
@@ -440,7 +440,7 @@ Outer:
 			if intervalOverlap(mint, maxt, chk.MinTime, chk.MaxTime) {
 				// Delete only until the current values and not beyond.
 				tmin, tmax := clampInterval(mint, maxt, chks[0].MinTime, chks[len(chks)-1].MaxTime)
-				stones.put(p.At(), Intervals{{tmin, tmax}})
+				stones.addInterval(p.At(), Interval{tmin, tmax})
 				continue Outer
 			}
 		}
--- a/block.go
+++ b/block.go
@@ -424,7 +424,7 @@ func (pb *Block) Delete(mint, maxt int64, ms ...labels.Matcher) error {
 	ir := pb.indexr
 
 	// Choose only valid postings which have chunks in the time-range.
-	stones := memTombstones{}
+	stones := NewMemTombstones()
 
 	var lset labels.Labels
 	var chks []chunks.Meta
@@ -440,7 +440,7 @@ Outer:
 			if intervalOverlap(mint, maxt, chk.MinTime, chk.MaxTime) {
 				// Delete only until the current values and not beyond.
 				tmin, tmax := clampInterval(mint, maxt, chks[0].MinTime, chks[len(chks)-1].MaxTime)
-				stones[p.At()] = Intervals{{tmin, tmax}}
+				stones.put(p.At(), Intervals{{tmin, tmax}})
 				continue Outer
 			}
 		}
@@ -452,7 +452,7 @@ Outer:
 
 	err = pb.tombstones.Iter(func(id uint64, ivs Intervals) error {
 		for _, iv := range ivs {
-			stones.add(id, iv)
+			stones.addInterval(id, iv)
 			pb.meta.Stats.NumTombstones++
 		}
 		return nil
--- a/tombstones.go
+++ b/tombstones.go
@@ -183,13 +183,12 @@ func (t *memTombstones) Iter(f func(uint64, Intervals) error) error {
 }
 
 // addInterval to an existing memTombstones
-func (t *memTombstones) addInterval(ref uint64, itvs ...Interval) *memTombstones {
+func (t *memTombstones) addInterval(ref uint64, itvs ...Interval) {
 	t.mtx.Lock()
 	defer t.mtx.Unlock()
 	for _, itv := range itvs {
 		t.mts[ref] = t.mts[ref].add(itv)
 	}
-	return t
 }
 
 func (memTombstones) Close() error {
--- a/tombstones.go
+++ b/tombstones.go
@@ -16,14 +16,12 @@ package tsdb
 import (
 	"encoding/binary"
 	"fmt"
+	"github.com/pkg/errors"
 	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
-
 	"sync"
-
-	"github.com/pkg/errors"
 )
 
 const tombstoneFilename = "tombstones"
@@ -174,8 +172,8 @@ func (t *memTombstones) Get(ref uint64) (Intervals, error) {
 }
 
 func (t *memTombstones) Iter(f func(uint64, Intervals) error) error {
-	t.mtx.Lock()
-	defer t.mtx.Unlock()
+	t.mtx.RLock()
+	defer t.mtx.RUnlock()
 	for ref, ivs := range t.mts {
 		if err := f(ref, ivs); err != nil {
 			return err
@@ -185,16 +183,12 @@ func (t *memTombstones) Iter(f func(uint64, Intervals) error) error {
 }
 
 // addInterval to an existing memTombstones
-func (t *memTombstones) addInterval(ref uint64, itv Interval) {
-	t.mtx.Lock()
-	t.mts[ref] = t.mts[ref].add(itv)
-	t.mtx.Unlock()
-}
-
-func (t *memTombstones) put(ref uint64, itvs Intervals) *memTombstones {
+func (t *memTombstones) addInterval(ref uint64, itvs ...Interval) *memTombstones {
 	t.mtx.Lock()
 	defer t.mtx.Unlock()
-	t.mts[ref] = itvs
+	for _, itv := range itvs {
+		t.mts[ref] = t.mts[ref].add(itv)
+	}
 	return t
 }
 
--- a/tombstones.go
+++ b/tombstones.go
@@ -21,6 +21,8 @@ import (
 	"os"
 	"path/filepath"
 
+	"sync"
+
 	"github.com/pkg/errors"
 )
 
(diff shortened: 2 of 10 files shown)


brian-brazil:
This only looks to be locking a read, where is the write?

codwu:
Sorry, I missed some logs.  I think it is better to add 'lock code' to every function operating `memTombstones` , i.e. `Get()`,`Iter()` and `add()` . I will recommit later

codwu:
@krasi-georgiev  I made a new memTombstones struct which contains a map, same as before, and a rwmutex. Then I modified its Get(), Iter() and add() functions. I also add a put() function. I think the map is now safe for concurrent read and write. I did some tests and these codes run ok with Prometheus. Unfortunately, I am not sure theses fix the bug, because I cannot reproduce the bug at my environment. I need someone to help me test it.

krasi-georgiev:
@games130 do you want to test this PR to see if it fixes the bug. 
Let me know if you need us to build the binary (OS and Architecture)

krasi-georgiev:
@games130 do you want to test the bug with this PR and see if it fixes it? 
I can build the binary if needed. Just need os and arch details.

tombstones.go
```diff
@@ -156,30 +157,55 @@ func readTombstones(dir string) (memTombstones, error) {
...
+	t.mtx.Lock()
+	t.mts[ref] = t.mts[ref].add(itv)
+	t.mtx.Unlock()
+}
+
+func (t *memTombstones) put(ref uint64, itvs Intervals) {
+	t.mtx.Lock()
+	t.mts[ref] = itvs
+	t.mtx.Unlock()
 }
```
krasi-georgiev: hmm.. `add` and `put` are a bit confusing , need to think of some better naming.

tombstones.go
```diff
@@ -156,30 +157,55 @@ func readTombstones(dir string) (memTombstones, error) {
 	return stonesMap, nil
 }
 
-type memTombstones map[uint64]Intervals
+type memTombstones struct {
+	mts map[uint64]Intervals
+	mtx sync.RWMutex
+}
+
+func newEmptyMemTombstones() *memTombstones {
```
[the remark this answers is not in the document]
codwu: I think it is an easier way create an memTombstones object.

tombstones.go
codwu: Maybe `add` ->`addInterval` and `put` ->`addIntervals`?

tombstones.go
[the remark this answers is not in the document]
krasi-georgiev: if you prefer it this way than it can  be simplified a bit. Also the name can be just `newMemTombstones`
```
func newMemTombstones() *memTombstones {
     return &memTombstones{mts: make(map[uint64]Intervals)}
}
```

tombstones.go
krasi-georgiev: hm ,seems we can use a single  Variadic Function.
`func (t *memTombstones) add(ref uint64, itv ...Interval)`

https://gobyexample.com/variadic-functions

games130:
@krasi-georgiev I can do the test but might need some guidance. Currently the prometheus i have running is on docker.. so wondering if it can be a docker image for me to download?

My OS infomantion:
Operating System: CentOS Linux 7 (Core)
CPE OS Name: cpe:/o:centos:centos:7
Kernel: Linux 3.10.0-693.11.1.el7.x86_64
Architecture: x86-64

krasi-georgiev:
sure I can do that , just lets wait to clear the nits and will build one for you.
Thanks

tombstones.go
codwu: `add()` append an `Interval` to the an exsiting `Intervals`, while `put()` replace the old `Intervals`. It's a not good idea to combine them.

krasi-georgiev: I can have a look in few days, but there must be a better way for this.

krasi-georgiev:
@games130 
here is the docker image build from this PR
https://hub.docker.com/r/krasimir/tsdb330/

games130:
okay, i will test it out.

games130:
@krasi-georgiev I posted the log for the container at the original issue post https://github.com/prometheus/prometheus/issues/4154

So far running it crash 2 times. let me know if you need me to provide any further info.
- the crash log is different from the original problem (so i guess the problem with continuous deleting the series is working now)

tombstones.go
```diff
@@ -150,36 +152,50 @@ func readTombstones(dir string) (memTombstones, error) {
...
-
-var emptyTombstoneReader = memTombstones{}
+type memTombstones struct {
+	mts map[uint64]Intervals
+	mtx sync.RWMutex
+}
 
-// EmptyTombstoneReader returns a TombstoneReader that is always empty.
-func EmptyTombstoneReader() TombstoneReader {
-	return emptyTombstoneReader
```
fabxc: We should not be removing exported functionality without replacement.

codwu:
How was the memory usage before?

tombstones.go
krasi-georgiev: @codwu in that case we can export newMemTombstones -> NewMemTombstones

codwu: @krasi-georgiev  ok, I made a new commit.

krasi-georgiev: @fabxc do you remember why was this exposed and if not why should we keep it exposed?
I guess if anyone is actually using it we can always re-expose it given it is really needed.

krasi-georgiev:
@games130 this PR is touching code only around tombstones/deleting_series so don't see how it can affect the overall memory usage. 
Probably the cause is somewhere else. 
At high loads it is normal to see high memory usage so as long as it doesn't get killed by the OS - OOM killed, shouldn't be a concern.

krasi-georgiev:
@games130  thanks for the update and good team work :+1: .
No other test is needed as the changes are only in the function name.

@codwu would you mind signing the commits.

the easyest way would be to squash into a single commit and than sign with
git commit --amend --signoff
than force push to this PR. 
ping me on IRC if you get stuck.

krasi-georgiev:
here is a page we did to auto sign on every commit which is required by many projects these days so might be best if you just do it for all contributions.

https://github.com/prometheus/prometheus/wiki/DCO-signing

tombstones.go
```diff
@@ -150,36 +152,50 @@ func readTombstones(dir string) (memTombstones, error) {
...
+func (t *memTombstones) Get(ref uint64) (Intervals, error) {
+	t.mtx.RLock()
+	defer t.mtx.RUnlock()
+	return t.mts[ref], nil
 }
 
-func (t memTombstones) Iter(f func(uint64, Intervals) error) error {
-	for ref, ivs := range t {
+func (t *memTombstones) Iter(f func(uint64, Intervals) error) error {
+	t.mtx.Lock()
```
gouthamve: Should be an `RLock`

gouthamve:
@codwu Sorry for jumping in late here. This looks good except for the nit, I commented above.

But I agree with @krasi-georgiev above, it doesn't make sense to have two functions given, we only use `put` in tests. The usage of `put` in `block.go` can be be replaced with an `addInterval`. I think `add(ref, itvs ...Interval)` is better idea.

gouthamve:
Further, if we are going to have a `put` function, it should be renamed to `set`?

tombstones.go
codwu: Yes, that is better

codwu:
@gouthamve  `add(ref, itvs ...Interval)` cannot replace the existing value, so we have to keep the  `put` function.  
`set` is a good name, and I will use it.

gouthamve:
Yes, but any existing uses of `put` can be replaced by an `add` producing the same effect. I don't see the need to `put` anywhere (it is being used in one place and there add produces the same effect).


Further the uses in tests can also be replaced by `add` in most if not all cases.

tombstones.go
```diff
@@ -150,36 +150,46 @@ func readTombstones(dir string) (memTombstones, error) {
...
 			return err
 		}
 	}
 	return nil
 }
 
-func (t memTombstones) add(ref uint64, itv Interval) {
-	t[ref] = t[ref].add(itv)
+// addInterval to an existing memTombstones
+func (t *memTombstones) addInterval(ref uint64, itvs ...Interval) *memTombstones {
```
fabxc: Why are we returning ourselves again? Is this only so the changed test can build input test cases more easily?
In general we are not returning the object of a method again unless its a copy and the original object wasn't mutated.

I think for the test case is fine to keep setting the map directly since we are just initializing it.

codwu: @fabxc  Yes, if there is no other cases, I will make it not returning itself.
Do you mean we construct the object directly?
```
		tombstones: &memTombstones{mts: map[uint64]Intervals{
			1: Intervals{{1, 3}},
			2: Intervals{{1, 3}, {6, 10}},
			3: Intervals{{6, 10}},
		}},
```

block.go
```diff
@@ -440,7 +440,7 @@ Outer:
 			if intervalOverlap(mint, maxt, chk.MinTime, chk.MaxTime) {
 				// Delete only until the current values and not beyond.
 				tmin, tmax := clampInterval(mint, maxt, chks[0].MinTime, chks[len(chks)-1].MaxTime)
-				stones[p.At()] = Intervals{{tmin, tmax}}
+				stones.put(p.At(), Intervals{{tmin, tmax}})
```
brian-brazil: I don't think this is right, this should be a addInterval too

tombstones.go
fabxc: Yes, that would be more consistent with what we have elsewhere.

tombstones.go
```diff
@@ -150,36 +150,45 @@ func readTombstones(dir string) (memTombstones, error) {
...
 	}
 
 	return stonesMap, nil
 }
 
-type memTombstones map[uint64]Intervals
-
-var emptyTombstoneReader = memTombstones{}
+type memTombstones struct {
+	mts map[uint64]Intervals
```
krasi-georgiev: `mts` is quite close to `mtx` and a bit confusing
whould you mind changing this to `intvlGroups`

fabxc:
Thanks 👍

Issue #235, opened 2017-12-27
Query and other benchmarks


Gouthamve:
We don't have proper benchmarks yet and that means we cannot see the potential impact of the changes we are introducing.

AlekSi:
"Help wanted"? How can I help?

krasi-georgiev:
we haven't discussed anything yet so it is a greenfield project, but here some random thoughts.
we already have a [benchmarking CI](https://github.com/prometheus/prombench ) for prometheus and can probably reuse a lot from the logic and pieces.
We can do a very simplified version for the  tsdb as well.

1. Implement some tsdb bench tool.
2. When a test is triggered it will build and run the version from the PR and another one from master or a release. 
3. Than just use grafana to display the exposed metrics and compare the results.

@sipian and I have worked on the Prombench project so can help along the way as well.

@gouthamve should have some better ideas where to start.

krasi-georgiev:
not sure if you have looked at the tsdb cli tool, but it already has some benchmarking already so maybe this is a good starting point. Can keep it simple to start with.
https://github.com/prometheus/tsdb/blob/9be883670eaf846e8f5bb65ea577dcd7d28d469e/cmd/tsdb/main.go#L43

krasi-georgiev:
@AlekSi  just found  PR: https://github.com/prometheus/tsdb/pull/240 that adds some more benchmarking.
I guess our first goal is to somehow automate these and display the results.

The way I see it:
1. comment /benchmark
2. Prow starts a job
3. The job runs `go bench` against master and another one against the PR and just displays the logs in the [prow job dashboard](http://prombench.prometheus.io).

krasi-georgiev:
just came across  https://github.com/timescale/tsbs which might help.

Sangeeta04Sharma:
Hello contributors, I would like to work on it as my GSOC project. I have used and worked Prometheus, Kubernetes, and Golang. Please guide me and help to start work on it.

krasi-georgiev:
@Sangeeta04Sharma can you ping me on IRC and will try to give some directions.

PR #331, opened 2018-05-15
Fix benchmarks

### LINKED ISSUE #305: Benchmarks are failing
All of the benchmarks in this repo are failing, it seems files needed in the testdata directory were removed at some point,

AlekSi:
Fixes #305 (which wasn't actually fixed before).


.travis.yml
```diff
@@ -9,4 +9,4 @@ go:
 go_import_path: github.com/prometheus/tsdb
 
 script:
-- go test -timeout 5m ./...
+- go test -v -bench . -benchtime 1ms -timeout 5m ./...
```
brian-brazil: We don't run benchmarks as part of CI

AlekSi: > Just to make sure they still work, not for numbers.

krasi-georgiev:
https://github.com/prometheus/tsdb/issues/235

eventually we should have some proper benchmarking in the CI, but in the meantime I don't think this gives us any real benefit here.

AlekSi:
The real benefit is that with that PR code in the repository actually _runs_, and also will run in the future – Travis CI will prevent future broken PRs from being merged. Without it, the code is broken. It is not about proper benchmarking in the CI at all.

krasi-georgiev:
don't conciser this so crucial to have a separate test, but for the sake of the good spirit don't mind if it gets merged if we get one more :+1:

brian-brazil:
Benchmarks can take a really long time to run, we shouldn't have them slowing down CI.

AlekSi:
That's why I added `-benchtime 1ms`. As you can see ([[1]](https://travis-ci.org/prometheus/tsdb/jobs/379086209), [[2]](https://travis-ci.org/prometheus/tsdb/jobs/379164140)) this increased CI time from 1:22 to 2:00. That sounds acceptable to me.

brian-brazil:
As that result indicates, it's not. All of our travis configs should be consistent, and we have benchmarks elsewhere that take a lot longer than 38 seconds.

krasi-georgiev:
oh if it increases the the tests by more than 1-2sec I am also against it.

krasi-georgiev:
I am not sure about this one , on one side @AlekSi is right that we want to make sure the bench tests work, but on another we want to keep the test times really minimal and running this locally increases the running time by 2min and not sure if this also might cause hitting some Travis limits at some point.

@fabxc , @gouthamve what do you think on this one?

AlekSi:
Extracted the less controversial bit into #345.

krasi-georgiev:
@AlekSi would you mind if we close this and can revisit at some point when we have a better plans for benchmarks which is tracked  in issue https://github.com/prometheus/tsdb/issues/235

PR #332, opened 2018-05-16
wal: add write ahead log package

fabxc:
This adds a new WAL that's agnostic to the actual record contents.
It's much simpler and should be more resilient than the existing one.

It is safe for concurrent reads across different processes.

@brian-brazil @jkohen

Integrating this into the rest of TSDB is another can of worms. I may need to make some slight tweaks here and there, but in general it seems complete and working.

The batch mode is mostly there because I'd like to implement the checkpointing (formerly truncation) by just writing another WAL so we don't need to implement another disk format. But in that batch write scenario, firing off millions of tiny writes wouldn't be ideal.

Also we need some nicer recovery handling than in the last one – but should be simpler to do as well.

### DIFF
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -1,0 +1,279 @@
+// Copyright 2018 The Prometheus Authors
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tsdb
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/pkg/errors"
+	"github.com/prometheus/tsdb/fileutil"
+	"github.com/prometheus/tsdb/wal"
+)
+
+// CheckpointStats returns stats about a created checkpoint.
+type CheckpointStats struct {
+	DroppedSeries     int
+	DroppedSamples    int
+	DroppedTombstones int
+	TotalSeries       int
+	TotalSamples      int
+	TotalTombstones   int
+}
+
+// LastCheckpoint returns the directory name of the most recent checkpoint.
+// If dir does not contain any checkpoints, ErrNotFound is returned.
+func LastCheckpoint(dir string) (string, int, error) {
+	files, err := ioutil.ReadDir(dir)
+	if err != nil {
+		return "", 0, err
+	}
+	// Traverse list backwards since there may be multiple checkpoints left.
+	for i := len(files) - 1; i >= 0; i-- {
+		fi := files[i]
+
+		if !strings.HasPrefix(fi.Name(), checkpointPrefix) {
+			continue
+		}
+		if !fi.IsDir() {
+			return "", 0, errors.Errorf("checkpoint %s is not a directory", fi.Name())
+		}
+		k, err := strconv.Atoi(fi.Name()[len(checkpointPrefix):])
+		if err != nil {
+			continue
+		}
+		return fi.Name(), k, nil
+	}
+	return "", 0, ErrNotFound
+}
+
+// DeleteCheckpoints deletes all checkpoints in dir that have an index
+// below n.
+func DeleteCheckpoints(dir string, n int) error {
+	var errs MultiError
+
+	files, err := ioutil.ReadDir(dir)
+	if err != nil {
+		return err
+	}
+	for _, fi := range files {
+		if !strings.HasPrefix(fi.Name(), checkpointPrefix) {
+			continue
+		}
+		k, err := strconv.Atoi(fi.Name()[len(checkpointPrefix):])
+		if err != nil || k >= n {
+			continue
+		}
+		if err := os.RemoveAll(filepath.Join(dir, fi.Name())); err != nil {
+			errs.Add(err)
+		}
+	}
+	return errs.Err()
+}
+
+const checkpointPrefix = "checkpoint."
+
+// Checkpoint creates a compacted checkpoint of segments in range [m, n] in the given WAL.
+// It includes the most recent checkpoint if it exists.
+// All series not satisfying keep and samples below mint are dropped.
+//
+// The checkpoint is stored in a directory named checkpoint.N in the same
+// segmented format as the original WAL itself.
+// This makes it easy to read it through the WAL package and concatenate
+// it with the original WAL.
+//
+// Non-critical errors are logged and not returned.
+func Checkpoint(logger log.Logger, w *wal.WAL, m, n int, keep func(id uint64) bool, mint int64) (*CheckpointStats, error) {
+	if logger == nil {
+		logger = log.NewNopLogger()
+	}
+	stats := &CheckpointStats{}
+
+	var sr io.Reader
+	{
+		lastFn, k, err := LastCheckpoint(w.Dir())
+		if err != nil && err != ErrNotFound {
+			return nil, errors.Wrap(err, "find last checkpoint")
+		}
+		if err == nil {
+			if m > k+1 {
+				return nil, errors.New("unexpected gap to last checkpoint")
+			}
+			// Ignore WAL files below the checkpoint. They shouldn't exist to begin with.
+			m = k + 1
+
+			last, err := wal.NewSegmentsReader(filepath.Join(w.Dir(), lastFn))
+			if err != nil {
+				return nil, errors.Wrap(err, "open last checkpoint")
+			}
+			defer last.Close()
+			sr = last
+		}
+
+		segs, err := wal.NewSegmentsRangeReader(w.Dir(), m, n)
+		if err != nil {
+			return nil, errors.Wrap(err, "create segment reader")
+		}
+		defer segs.Close()
+
+		if sr != nil {
+			sr = io.MultiReader(sr, segs)
+		} else {
+			sr = segs
+		}
+	}
+
+	cpdir := filepath.Join(w.Dir(), fmt.Sprintf("checkpoint.%06d", n))
+	cpdirtmp := cpdir + ".tmp"
+
+	if err := os.MkdirAll(cpdirtmp, 0777); err != nil {
+		return nil, errors.Wrap(err, "create checkpoint dir")
+	}
+	cp, err := wal.New(nil, nil, cpdirtmp)
+	if err != nil {
+		return nil, errors.Wrap(err, "open checkpoint")
+	}
+
+	r := wal.NewReader(sr)
+
+	var (
+		series  []RefSeries
+		samples []RefSample
+		tstones []Stone
+		dec     RecordDecoder
+		enc     RecordEncoder
+		buf     []byte
+		recs    [][]byte
+	)
+	for r.Next() {
+		series, samples, tstones = series[:0], samples[:0], tstones[:0]
+
+		// We don't reset the buffer since we batch up multiple records
+		// before writing them to the checkpoint.
+		// Remember where  the record for this iteration starts.
+		start := len(buf)
+		rec := r.Record()
+
+		switch dec.Type(rec) {
+		case RecordSeries:
+			series, err = dec.Series(rec, series)
+			if err != nil {
+				return nil, errors.Wrap(err, "decode series")
+			}
+			// Drop irrelevant series in place.
+			repl := series[:0]
+			for _, s := range series {
+				if keep(s.Ref) {
+					repl = append(repl, s)
+				}
+			}
+			if len(repl) > 0 {
+				buf = enc.Series(repl, buf)
+			}
+			stats.TotalSeries += len(series)
+			stats.DroppedSeries += len(series) - len(repl)
+
+		case RecordSamples:
+			samples, err = dec.Samples(rec, samples)
+			if err != nil {
+				return nil, errors.Wrap(err, "decode samples")
+			}
+			// Drop irrelevant samples in place.
+			repl := samples[:0]
+			for _, s := range samples {
+				if s.T >= mint {
+					repl = append(repl, s)
+				}
+			}
+			if len(repl) > 0 {
+				buf = enc.Samples(repl, buf)
+			}
+			stats.TotalSamples += len(samples)
+			stats.DroppedSamples += len(samples) - len(repl)
+
+		case RecordTombstones:
+			tstones, err = dec.Tombstones(rec, tstones)
+			if err != nil {
+				return nil, errors.Wrap(err, "decode deletes")
+			}
+			// Drop irrelevant tombstones in place.
+			repl := tstones[:0]
+			for _, s := range tstones {
+				for _, iv := range s.intervals {
+					if iv.Maxt >= mint {
+						repl = append(repl, s)
+						break
+					}
+				}
+			}
+			if len(repl) > 0 {
+				buf = enc.Tombstones(repl, buf)
+			}
+			stats.TotalTombstones += len(tstones)
+			stats.DroppedTombstones += len(tstones) - len(repl)
+
+		default:
+			return nil, errors.New("invalid record type")
+		}
+		if len(buf[start:]) == 0 {
+			continue // All contents discarded.
+		}
+		recs = append(recs, buf[start:])
+
+		// Flush records in 1 MB increments.
+		if len(buf) > 1*1024*1024 {
+			if err := cp.Log(recs...); err != nil {
+				return nil, errors.Wrap(err, "flush records")
+			}
+			buf, recs = buf[:0], recs[:0]
+		}
+	}
+	// If we hit any corruption during checkpointing, repairing is not an option.
+	// The head won't know which series records are lost.
+	if r.Err() != nil {
+		return nil, errors.Wrap(r.Err(), "read segments")
+	}
+
+	// Flush remaining records.
+	if err := cp.Log(recs...); err != nil {
+		return nil, errors.Wrap(err, "flush records")
+	}
+	if err := cp.Close(); err != nil {
+		return nil, errors.Wrap(err, "close checkpoint")
+	}
+	if err := fileutil.Rename(cpdirtmp, cpdir); err != nil {
+		return nil, errors.Wrap(err, "rename checkpoint file")
+	}
+	if err := w.Truncate(n + 1); err != nil {
+		// If truncating fails, we'll just try again at the next checkpoint.
+		// Leftover segments will just be ignored in the future if there's a checkpoint
+		// that supersedes them.
+		level.Error(logger).Log("msg", "truncating segments failed", "err", err)
+	}
+	if err := DeleteCheckpoints(w.Dir(), n); err != nil {
+		// Leftover old checkpoints do not cause problems down the line beyond
+		// occupying disk space.
+		// They will just be ignored since a higher checkpoint exists.
+		level.Error(logger).Log("msg", "delete old checkpoints", "err", err)
+	}
+	return stats, nil
+}
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -35,9 +35,9 @@ type CheckpointStats struct {
 	DroppedSeries     int
 	DroppedSamples    int
 	DroppedTombstones int
-	TotalSeries       int
-	TotalSamples      int
-	TotalTombstones   int
+	TotalSeries       int // Processed series including dropped ones.
+	TotalSamples      int // Processed samples inlcuding dropped ones.
+	TotalTombstones   int // Processed tombstones including droppes ones.
 }
 
 // LastCheckpoint returns the directory name of the most recent checkpoint.
@@ -129,16 +129,16 @@ func Checkpoint(logger log.Logger, w *wal.WAL, m, n int, keep func(id uint64) bo
 			sr = last
 		}
 
-		segs, err := wal.NewSegmentsRangeReader(w.Dir(), m, n)
+		segsr, err := wal.NewSegmentsRangeReader(w.Dir(), m, n)
 		if err != nil {
 			return nil, errors.Wrap(err, "create segment reader")
 		}
-		defer segs.Close()
+		defer segsr.Close()
 
 		if sr != nil {
-			sr = io.MultiReader(sr, segs)
+			sr = io.MultiReader(sr, segsr)
 		} else {
-			sr = segs
+			sr = segsr
 		}
 	}
 
@@ -169,7 +169,7 @@ func Checkpoint(logger log.Logger, w *wal.WAL, m, n int, keep func(id uint64) bo
 
 		// We don't reset the buffer since we batch up multiple records
 		// before writing them to the checkpoint.
-		// Remember where  the record for this iteration starts.
+		// Remember where the record for this iteration starts.
 		start := len(buf)
 		rec := r.Record()
 
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -37,7 +37,7 @@ type CheckpointStats struct {
 	DroppedTombstones int
 	TotalSeries       int // Processed series including dropped ones.
 	TotalSamples      int // Processed samples inlcuding dropped ones.
-	TotalTombstones   int // Processed tombstones including droppes ones.
+	TotalTombstones   int // Processed tombstones including dropped ones.
 }
 
 // LastCheckpoint returns the directory name of the most recent checkpoint.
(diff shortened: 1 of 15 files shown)


krasi-georgiev:
I smell some offline kubecon discussions :thinking: 
I am just guessing, but probably this is in relation to the WAL replication mentioned in @gouthamve's [talk](https://youtu.be/0UvKEHFNu4Q?t=1316). 

Are there any more discussions how and where would this be used?

krasi-georgiev:
I figured it to be in relation with this proposal
https://docs.google.com/document/d/1TEqqE_Stq04drhjSU1I7Ctmuy0dpsvlPL1AKxqEQoSg

wal/wal.go
```diff
@@ -0,0 +1,628 @@
...
+			}
+		case recLast:
+			if i == 0 {
+				return errors.New("unexpected last record")
+			}
+			return nil
+		default:
+			return errors.Errorf("unexpected record type %d", typ)
+		}
+		// Only increment i for non-zero records since we use it below
```
krasi-georgiev: below -> above

wal/wal.go
```diff
@@ -0,0 +1,628 @@
...
+	// Don't block further writes by handling the last segment.
+	// TODO(fabxc): write a termination page as a marker to detect torn segments?
+	w.actorc <- func() {
+		if err := w.fsync(prev); err != nil {
+			level.Error(w.logger).Log("msg", "sync previous segment", "err", err)
+		}
+		if err := prev.Close(); err != nil {
+			level.Error(w.logger).Log("msg", "close previous segment", "err", err)
+		}
+	}
```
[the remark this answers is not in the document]
fabxc: Unfortunately not, mutex would mean the writing goroutines would block. Fsync can take several seconds or much longer in bad conditions.

fabxc:
> the code is easy to follow and read 👍

Added 3 commits to change that :P

I know it's quite a bit of code for a single PR, but it's well-separated bottom-up along commits. I think some parts are just better to review with the full picture at hand.

Old WAL code was not removed. It needs to stick around anyway for migration procedures.

wal/wal.go
```diff
@@ -0,0 +1,815 @@
...
+	return nil
+}
+
+// flushPage writes the contents of the page to disk.
+// If clear is set or no more records will reasonbly fit into it, its
+// remaining size gets written as zero data and its reset for the next physical page.
```
jkohen: I'm not sure I understand the last part. Is it saying that the page will be padded with byte-0 if clear is true or the page doesn't have room for another record?

wal/wal.go
```diff
@@ -0,0 +1,815 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+func (w *WAL) fsync(f *Segment) error {
+	start := time.Now()
+	err := fileutil.Fsync(f.File)
+	w.fsyncDuration.Observe(time.Since(start).Seconds())
```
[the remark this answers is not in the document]
fabxc: In Prometheus we've a few instrumentation conventions that are not baked into our model. One of them is that we always use base units, i.e. bytes, seconds, ...
The idea is that you don't have to make unit adjustments if you want to do binary ops between metrics and such and can just do a final division/multiplication on the overall result. Dashboard builders also often allow you to set a unit and they'll pick the most sensible scale – you could then just always default to the base unit and don't have to handpick it for each graph.

Since everything is a double in our world, this choice conveniently doesn't really impact data size or similar.

`Time.Seconds()` returns a float64 by the way, so we aren't losing the sub-seconds in case you meant that.

wal/wal.go
fabxc: If `clear` is false, we may fill up the remainer of the page with zero bytes if we think there's not enough space left to fit in another (partial) record. If `clear` is true, we force the zero padding no matter how much space is left. We need the latter on shutdown or when completing a segment (when no partial records may be written), since we never want to leave a segment with a size that's not a multiple of 32KB.

I will make the comment clearer.

checkpoint.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	for _, fi := range files {
+		if !strings.HasPrefix(fi.Name(), checkpointPrefix) {
+			continue
+		}
+		k, err := strconv.Atoi(fi.Name()[len(checkpointPrefix):])
+		if err != nil || k >= n {
+			continue
+		}
+		if err := os.RemoveAll(filepath.Join(dir, fi.Name())); err != nil {
```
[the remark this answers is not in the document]
fabxc: You are right. We shouldn't abort on failure in both places. Especially since those leftover checkpoints don't cause problems in general since we just ignore them.

fabxc:
Some initial feedback after running this in a small Prometheus server with high frequency scrapes:

Memory and CPU usage seem about 3-5% lower on average, which may just be noise. Max usage shows no notable difference.
The page flush to page completion ratio when scraping ~450 series is about 10x. So each 32KB WAL page gets written 10 times. Assuming a 4KB physical page size, we are looking at ~1.25 write amplification – so nothing to worry about there even if it had high variance.

Startups got about 3x slower. This is not super concerning but also not great for big setups and worth investigating.

fabxc:
@brian-brazil @gouthamve can I have a review from one of you?

--- 

We introduced a regression a while back regarding block order. It never made it into Prometheus though. Fixed it in here and added a test.

jkohen:
This is looking good.

fabxc:
During performance analysis of this change, I found a regression. A buffer that seems stack-allocated doesn't pass escape analysis and gets thrown on the heap. That resulted in big spikes. I pushed a commit to fix this.

This show memory usage before and after the change. It shows the released v2.2.1 and a version with the new WAL and in-memory metadata store. 

screen shot 2018-05-24 at 3 50 53 pm

checkpoint.go
```diff
@@ -0,0 +1,279 @@
...
+// Checkpoint creates a compacted checkpoint of segments in range [m, n] in the given WAL.
+// It includes the most recent checkpoint if it exists.
+// All series not satisfying keep and samples below mint are dropped.
+//
+// The checkpoint is stored in a directory named checkpoint.N in the same
+// segmented format as the original WAL itself.
+// This makes it easy to read it through the WAL package and concatenate
+// it with the original WAL.
+//
+// Non-critical errors are locked and not returned.
```
brian-brazil: logged?

checkpoint.go
```diff
@@ -0,0 +1,279 @@
...
+				buf = enc.Tombstones(repl, buf)
+			}
+			stats.TotalTombstones += len(tstones)
+			stats.DroppedTombstones += len(tstones) - len(repl)
+
+		default:
+			return nil, errors.New("invalid record type")
+		}
+		if len(buf[start:]) == 0 {
+			continue // all contents discarded
```
brian-brazil: Comment style

checkpoint.go
```diff
@@ -0,0 +1,279 @@
...
+		}
+		recs = append(recs, buf[start:])
+
+		// flush records in 1 MB increments.
```
brian-brazil: Capital letter

db.go
```diff
@@ -613,10 +616,6 @@ func OverlappingBlocks(bm []BlockMeta) Overlaps {
 	if len(bm) <= 1 {
 		return nil
 	}
-	sort.Slice(bm, func(i, j int) bool {
```
brian-brazil: Why are you removing this?

fileutil/fileutil.go
```diff
@@ -23,3 +24,25 @@ func ReadDir(dirpath string) ([]string, error) {
 	sort.Strings(names)
 	return names, nil
 }
+
+// Rename safely renames a file.
```
brian-brazil: This has a race condition between the remove and the rename

head.go
```diff
@@ -263,66 +275,130 @@ func (h *Head) ReadWAL() error {
...
 	}
 	wg.Wait()
 
+	if unknownRefs > 0 {
+		level.Warn(h.logger).Log("msg", "unknown series references", "count", unknownRefs)
+	}
+	return nil
+}
+
+// Init backfills data from the write ahead log and prepares the head for writes.
```
brian-brazil: This is more a load than a backfill, which is confusing

head.go
```diff
@@ -348,18 +424,37 @@ func (h *Head) Truncate(mint int64) error {
...
 		return h.series.getByID(id) != nil
 	}
-	if err := h.wal.Truncate(mint, keep); err == nil {
-		level.Info(h.logger).Log("msg", "WAL truncation completed", "duration", time.Since(start))
-	} else {
-		level.Error(h.logger).Log("msg", "WAL truncation failed", "err", err, "duration", time.Since(start))
+	if _, err = Checkpoint(h.logger, h.wal, m, n, keep, mint); err != nil {
+		return errors.Wrap(err, "create checkpoint")
 	}
 	h.metrics.walTruncateDuration.Observe(time.Since(start).Seconds())
```
brian-brazil: We should probably have a separate metric for the checkpoint duration

wal/wal.go
```diff
@@ -0,0 +1,825 @@
...
+	}
+	next, err := CreateSegment(w.dir, w.segment.Index()+1)
+	if err != nil {
+		return errors.Wrap(err, "create new segment file")
+	}
+	prev := w.segment
+	w.segment = next
+	w.donePages = 0
+
+	// Don't block further writes by fsyinc the last segment.
```
brian-brazil: fsyncing

wal/wal.go
```diff
@@ -0,0 +1,825 @@
...
+	// Flush the last page and zero out all its remaining size.
+	// We must not flush an empty page as it would falsely signal
+	// the segment is done if we start writing to it again after opening.
+	if w.page.alloc > 0 {
+		if err := w.flushPage(true); err != nil {
+			return err
+		}
+	}
+
+	donec := make(chan struct{})
```
brian-brazil: I think a single-level channel is sufficient here

wal/wal.go
```diff
@@ -0,0 +1,825 @@
...
+}
+
+// Dir returns the directory of the WAL.
+func (w *WAL) Dir() string {
+	return w.dir
+}
+
+func (w *WAL) run() {
+	for {
+		// Processing all pending functions has precedence over shutdown.
```
brian-brazil: What happens if we shutdown with pending functions? The current code seems to allow for that

db.go
fabxc: This was mutating the input slice in a function that should do a read-only check. Thus current master actually has a regression. The sort is moved properly into the reload function.

head.go
fabxc: It was renamed to checkpoint but as @jkohen pointed out, it might break alerts and dashboards. It is essentially still part of the truncation process, so keeping the name seemed fine. The `WAL.Truncate` call is just simple file removal. I don't think there's anything worth measuring in there.

wal/wal.go
fabxc: It's not, we have to wait until run terminated successfully. Currently it returns right when it reads from the channel. But this is quite implementation dependent and would be an easy regression target.

head.go
```diff
@@ -221,17 +223,27 @@ func (h *Head) processWALSamples(
 				h.metrics.chunksCreated.Inc()
 				h.metrics.chunks.Inc()
 			}
+			if s.T > maxt {
+				maxt = s.T
+			}
```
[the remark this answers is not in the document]
fabxc: I actually uncovered several more issues around this general area. I've adjusted things to make it work, but it touches a few more places. So I'd rather add this in a PR on top.

krasi-georgiev:
btw did you try to run this with `--race` enabled ?

fabxc:
@krasi-georgiev yea, that looks all okay.

db.go
fabxc: This is fixed and tested in master via #335 so no longer relevant for this PR.

fabxc:
@brian-brazil how does this one look? I've two follow up PRs ready.

fileutil/fileutil.go
brian-brazil: The doc says it's "safe" however there's a race condition. In what way is the meant to be safe?

wal/wal.go
brian-brazil: It looks like unnecessary complexity currently, especially as WAL is a struct rather than an interface.

brian-brazil:
I'm still waiting on file format docs/protocol before giving it a full review.

fabxc:
The record encoding itself did not change. But so far the WAL format isn't documented at all. I've another PR in the queue that changes the record format and backfills documentation.

I can backport it to this one though without those changes.

fabxc:
@brian-brazil docs added

fabxc:
Yes, I just wanted to start getting this in since there are further changes on top of it and I wanted to enter a thorough benchmarking and validation phase sooner rather than later.

head.go
```diff
@@ -348,18 +424,37 @@ func (h *Head) Truncate(mint int64) error {
...
+	m, n, err := h.wal.Segments()
+	if err != nil {
+		return errors.Wrap(err, "get segment range")
+	}
+	n-- // Never consider last segment for checkpoint.
+	if n < 0 {
+		return nil // no segments yet.
+	}
+	// The lower third of segments should contain mostly obsolete samples.
+	// If we have too few segments, it's not worth checkpointing yet.
```
brian-brazil: It'd be clearer if this said "If we don't have 3 segments"

fabxc:
> As an async reader of this file, how do I know that a page is fully flushed?

Pages are written partially, so we never write 0 bytes that we later fill with real data. A page is thus flushed, when a full page can be read.

checkpoint.go
```diff
@@ -0,0 +1,279 @@
...
+
+			last, err := wal.NewSegmentsReader(filepath.Join(w.Dir(), lastFn))
+			if err != nil {
+				return nil, errors.Wrap(err, "open last checkpoint")
+			}
+			defer last.Close()
+			sr = last
+		}
+
+		segs, err := wal.NewSegmentsRangeReader(w.Dir(), m, n)
```
gouthamve: `segs` seems like an array of segments, `segsr` to indicate reader?

checkpoint.go
```diff
@@ -0,0 +1,279 @@
...
+//
+// Non-critical errors are logged and not returned.
+func Checkpoint(logger log.Logger, w *wal.WAL, m, n int, keep func(id uint64) bool, mint int64) (*CheckpointStats, error) {
+	if logger == nil {
+		logger = log.NewNopLogger()
+	}
+	stats := &CheckpointStats{}
+
+	var sr io.Reader
+	{
```
gouthamve: Why do we have a new block?

checkpoint.go
```diff
@@ -0,0 +1,279 @@
...
+
+		case RecordTombstones:
+			tstones, err = dec.Tombstones(rec, tstones)
+			if err != nil {
+				return nil, errors.Wrap(err, "decode deletes")
+			}
+			// Drop irrelevant tombstones in place.
+			repl := tstones[:0]
+			for _, s := range tstones {
+				for _, iv := range s.intervals {
```
gouthamve: Check for `keep(s.ref)` too?

checkpoint.go
```diff
@@ -0,0 +1,279 @@
...
+
+		case RecordSamples:
+			samples, err = dec.Samples(rec, samples)
+			if err != nil {
+				return nil, errors.Wrap(err, "decode samples")
+			}
+			// Drop irrelevant samples in place.
+			repl := samples[:0]
+			for _, s := range samples {
+				if s.T >= mint {
```
gouthamve: Check for `keep(s.Ref)` too?

checkpoint.go
fabxc: To not pollute the entire function with temporary variables we create here that we just need for a few lines.

checkpoint.go
fabxc: That would help for data size. I wonder about the penalty though of calling a function for each sample – especially since we don't necessarily know how expensive that function is.

checkpoint.go
fabxc: Same as above. I'd err on the side of having a few MB of temporary overhead over calling something with unknown cost a billion times.

docs/format/wal.md
```diff
@@ -0,0 +1,72 @@
...
+e.g. `000000`, `000001`, `000002`, etc., and are limited to 128MB by default.
+A segment is written to in pages of 32KB. Only the last page of the most recent segment
+may be partial. A WAL record is an opaque byte slice that gets split up into sub-records
+should it exceed the remaining space of the current page. Records are never split across
+segment boundaries.
+The encoding of pages is largely borrowed from [LevelDB's/RocksDB's wirte ahead log.][1]
+
+Notable deviations are that the record fragment is encoded as:
+
+┌───────────┬──────────┬────────────┬──────────────┐
```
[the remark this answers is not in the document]
fabxc: Oh, I believe we are just missing the code block formatting.

fileutil/fileutil.go
fabxc: I meant to elaborate on what exactly the race condition is.

head.go
```diff
@@ -520,14 +627,41 @@ func (a *headAppender) AddFast(ref uint64, t int64, v float64) error {
...
+	}
+
+	buf := a.head.getBytesBuffer()
+	defer func() { a.head.putBytesBuffer(buf) }()
+
+	var rec []byte
+	var enc RecordEncoder
+
+	if len(a.series) > 0 {
+		rec = enc.Series(a.series, buf)
```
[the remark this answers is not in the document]
fabxc: Mh, fair point – worth thinking about further. However I'd imagine rollbacks to be extremely rare. Combined with the small fraction of commits that actually create series, I wonder whether this can be an indicator for common issues.

wal/wal.go
```diff
@@ -0,0 +1,828 @@
...
+		return false
+	}
+	r.err = err
+	return r.err == nil
+}
+
+func (r *Reader) next() (err error) {
+	// We have to use r.buf since allocating byte arrays here fails escape
+	// analysis and ends up on the heap, even though it seemingly should not.
+	hdr := r.buf[:7]
```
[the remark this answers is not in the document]
fabxc: 👍

wal/wal.go
```diff
@@ -0,0 +1,828 @@
...
+	f, err := os.OpenFile(SegmentName(dir, k), os.O_WRONLY|os.O_APPEND, 0666)
+	if err != nil {
+		return nil, err
+	}
+	stat, err := f.Stat()
+	if err != nil {
+		f.Close()
+		return nil, err
+	}
+	// If the last page is torn, fill it with zeros.
```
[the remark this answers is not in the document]
fabxc: That a write was torn, i.e. didn't complete entirely, due to a crash for example.

wal/wal.go
```diff
@@ -0,0 +1,828 @@
...
+func (e *CorruptionErr) Error() string {
+	if e.Segment < 0 {
+		return fmt.Sprintf("corruption after %d bytes: %s", e.Offset, e.Err)
+	}
+	return fmt.Sprintf("corruption in segment %d at %d: %s", e.Segment, e.Offset, e.Err)
+}
+
+// OpenWriteSegment opens segment k in dir. The returned segment is ready for new appends.
+func OpenWriteSegment(dir string, k int) (*Segment, error) {
+	// Only .active segments are allowed to be opened for write.
```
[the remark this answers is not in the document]
fabxc: Obsolete comment.

fabxc:
> But for me the biggest question is that I am not seeing any versioning there. The segments have no headers, is that intentional?

Yea, I've been thinking about that a fair bit.
Ultimately I found per-segment versioning very weird. It would imply we may allow a mix of different versions, which we definitely don't want I'd say.
OTOH, the records we write to the WAL are still versioned – that's the most likely area of change. The record fragments the WAL implementation writes also have a flag byte that allows for versioning. That leaves the chance that we may want to migrate to yet another WAL, that cannot even keep the compatibility of having the first byte of a record (fragment) being a flag. That seemed extremely unlikely. And if, we can always add a version file into the WAL dir later that then at least applies to all segments equally. Doing this now seemed premature.

As for magic file headers – I think they are quite common but the practical usefulness seems rather low.

docs/format/wal.md
```diff
@@ -0,0 +1,72 @@
+# WAL Disk Format
+
+The write ahead log operates in segments that that are numbered and sequential,
+e.g. `000000`, `000001`, `000002`, etc., and are limited to 128MB by default.
+A segment is written to in pages of 32KB. Only the last page of the most recent segment
+may be partial. A WAL record is an opaque byte slice that gets split up into sub-records
+should it exceed the remaining space of the current page. Records are never split across
```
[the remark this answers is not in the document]
fabxc: Then we'll end up with a segment larger than the default size of 128MB. Will add that.

head.go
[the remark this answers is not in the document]
fabxc: Do we agree that this observation has nothing to do with this PR in particular and should be investigated/addressed seperately?

checkpoint.go
```diff
@@ -0,0 +1,279 @@
...
+		defer segs.Close()
+
+		if sr != nil {
+			sr = io.MultiReader(sr, segs)
+		} else {
+			sr = segs
+		}
+	}
+
+	cpdir := filepath.Join(w.Dir(), fmt.Sprintf("checkpoint.%06d", n))
```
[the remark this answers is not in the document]
fabxc: I bumped it. This is just the padding though, so even if it overflowed, it still won't wrap around.

wal/wal.go
```diff
@@ -0,0 +1,828 @@
...
+		p.flushed = 0
+		w.donePages++
+	}
+	return nil
+}
+
+type recType uint8
+
+const (
+	recPageTerm recType = 0 // rest of page is empty
```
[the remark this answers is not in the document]
fabxc: I referenced the RocksDB docs which do a good job at explaining the whole approach. I see little value in duplicating it.
In the unlikely case they change their approach and docs, we can always change that.

wal/wal.go
```diff
@@ -0,0 +1,828 @@
...
+		select {
+		case f := <-w.actorc:
+			f()
+		case donec := <-w.stopc:
+			defer close(donec)
+			break Loop
+		}
+	}
+	// Drain and process any remaining functions.
+	for {
```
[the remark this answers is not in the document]
fabxc: It's not being closed anywhere so it wouldn't terminate. Changed this so the range works.

docs/format/wal.md
```diff
@@ -0,0 +1,80 @@
+# WAL Disk Format
+
+The write ahead log operates in segments that that are numbered and sequential,
```
simonpasquier: "that ~~that~~ are numbered"

wal/wal.go
```diff
@@ -0,0 +1,822 @@
...
+	f, err := os.Open(fn)
+	if err != nil {
+		return nil, err
+	}
+	return &Segment{File: f, i: k, dir: filepath.Dir(fn)}, nil
+}
+
+// WAL is a write ahead log that stores records in segment files.
+// It must be read from start to end once before logging new data.
+// If an erroe occurs during read, the repair procedure must be called
```
simonpasquier: s/erroe/error/

fileutil/fileutil.go
brian-brazil: From the time the removeall starts to the rename ends, partial or no file may be present.

checkpoint.go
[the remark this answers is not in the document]
brian-brazil: You're depending on sorting these values as strings though, so could missort.

wal/wal.go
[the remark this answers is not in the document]
brian-brazil: It still needs to be documented here. "largely borrowed from" doesn't tell me the actual spec, which I expect to get entirely from reading this document standalone.

head.go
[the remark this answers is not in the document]
brian-brazil: This PR is changing key code around this, and will likely affect the design of the fix. I'd prefer to see it in here - should be a matter of writing out series on a rollback too.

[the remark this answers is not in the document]
fabxc: This is a big PR already. It doesn't change anything about the issue at hand, so I don't see why one would throw this on top here as well.
We can do another PR right after.

fabxc:

checkpoint.go
```diff
@@ -0,0 +1,279 @@
...
+)
+
+// CheckpointStats returns stats about a created checkpoint.
+type CheckpointStats struct {
+	DroppedSeries     int
+	DroppedSamples    int
+	DroppedTombstones int
+	TotalSeries       int // Processed series including dropped ones.
+	TotalSamples      int // Processed samples inlcuding dropped ones.
+	TotalTombstones   int // Processed tombstones including droppes ones.
```
brian-brazil: dropped

docs/format/wal.md
```diff
@@ -0,0 +1,88 @@
...
+│ type = 3 <1b>                                       │
+├─────────────────────────────────────────────────────┤
+│ ┌─────────┬───────────────────┬───────────────────┐ │
+│ │ id <8b> │ min_time <varint> │ max_time <varint> │ │
+│ └─────────┴───────────────────┴───────────────────┘ │
+│                        . . .                        │
+└─────────────────────────────────────────────────────┘
+```
+
+[1][https://github.com/facebook/rocksdb/wiki/Write-Ahead-Log-File-Format]
```
brian-brazil: GitHub isn't rendering this properly.

wal/wal.go
```diff
@@ -0,0 +1,822 @@
...
+		w.donePages++
+	}
+	return nil
+}
+
+type recType uint8
+
+const (
+	recPageTerm recType = 0 // rest of page is empty
+	recFull     recType = 1 // full record
```
brian-brazil: Full stops and capital letters. There's others in this file too.

brian-brazil:
:+1:

PR #335, opened 2018-05-28
Ensure correct block order on reload

fabxc:
Due to a regression blocks were no longer ordered by time before
being stored in memory. This made data intermittently become unavailable
for queries.

Backport from #332
 
@brian-brazil @krasi-georgiev This must be included in the vendor update in Prometheus before releasing.


brian-brazil:
:+1:

PR #337, opened 2018-05-29
Move to flock lockfile

fabxc:
@brian-brazil 

That's our old flock package so I assume this is battle-tested.


brian-brazil:
Hmm, I don't think the windows code is going any locking.

PR #334, opened 2018-05-24
Update head time when reading wal

krasi-georgiev:
When reading he WAL head mint and maxt didn't get updated.

@gouthamve found this bug I am just submitting it on his behalf.

PR #339, opened 2018-05-29
Properly initialize head time

fabxc:
This fixes various issues when initializing the head time range
under different starting conditions.

Depends on #332 

@krasi-georgiev as promised
@jkohen FYI as it depends on the new WAL stuff, but probably not too interesting for you

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -19,6 +19,7 @@ import (
 	"math"
 	"math/rand"
 	"os"
+	"path"
 	"path/filepath"
 	"sort"
 	"testing"
@@ -30,6 +31,7 @@ import (
 	"github.com/prometheus/tsdb/index"
 	"github.com/prometheus/tsdb/labels"
 	"github.com/prometheus/tsdb/testutil"
+	"github.com/prometheus/tsdb/wal"
 )
 
 func openTestDB(t testing.TB, opts *Options) (db *DB, close func()) {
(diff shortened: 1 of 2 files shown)


head.go
```diff
@@ -462,10 +482,7 @@ func (h *Head) Truncate(mint int64) error {
 // for a compltely fresh head with an empty WAL.
```
krasi-georgiev: compltely -> completely

head.go
```diff
@@ -576,15 +594,16 @@ func (h *Head) putBytesBuffer(b []byte) {
 }
 
 type headAppender struct {
-	head       *Head
-	mint, maxt int64
+	head         *Head
+	minValidTime int64 // No samples below this timestamp are allowed.
+	mint, maxt   int64
```
krasi-georgiev: Maybe some comments why we need separate `mint` and `minValidTime` ?

db_test.go
```diff
@@ -1011,3 +1013,109 @@ func TestOverlappingBlocksDetectsAllOverlaps(t *testing.T) {
 		{Min: 8, Max: 9}: {nc1[8], nc1[9]},                                 // 7-10, 8-9
 	}, OverlappingBlocks(nc1))
 }
+
+func TestInitializeHeadTimestamp(t *testing.T) {
```
krasi-georgiev: shouldn't this test be in  `head_test.go`?

krasi-georgiev: actually there is already a head test where you can just add few lines similar to what I did in the other PR
https://github.com/prometheus/tsdb/pull/334/files#diff-0dc7b53c06c14aba584e5b4d6986fa07R96

fabxc: The test cover cases with existing blocks, which is beyond the scope of `Head`. `DB` has to startup correctly and make some calls of `Head`'s methods in the right order and with the right arguments. That's nothing we can test with an isolated `Head`.

head.go
```diff
@@ -226,10 +226,27 @@ func (h *Head) processWALSamples(
...
+func (h *Head) updateMinMaxTime(mint, maxt int64) {
+	for {
+		lt := h.MinTime()
+		if mint >= lt {
+			break
+		}
+		if atomic.CompareAndSwapInt64(&h.minTime, lt, mint) {
+			break
+		}
+	}
```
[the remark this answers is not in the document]
fabxc: Because `CompareAndSwap` can fail and we have to retry it then.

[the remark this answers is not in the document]
krasi-georgiev: thanks! I always thought these are blocking until they succeed.

db_test.go
```diff
@@ -1011,3 +1013,109 @@ func TestOverlappingBlocksDetectsAllOverlaps(t *testing.T) {
...
+	t.Run("clean", func(t *testing.T) {
+		dir, err := ioutil.TempDir("", "test_head_init")
+		testutil.Ok(t, err)
+		defer os.RemoveAll(dir)
+
+		db, err := Open(dir, nil, nil, nil)
+		testutil.Ok(t, err)
```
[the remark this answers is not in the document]
krasi-georgiev: looking it again it is the same number of lines so all the same.

fabxc:

brian-brazil:
:+1:

PR #340, opened 2018-05-29
Migrate write ahead log

fabxc:
On startup, rewrite the old write ahead log into the new format once.

@jkohen @brian-brazil

### DIFF
--- a/wal.go
+++ b/wal.go
@@ -1214,6 +1214,9 @@ func (r *walReader) decodeDeletes(flag byte, b []byte, res *[]Stone) error {
 
 // MigrateWAL rewrites the deprecated write ahead log into the new format.
 func MigrateWAL(logger log.Logger, dir string) (err error) {
+	if logger == nil {
+		logger = log.NewNopLogger()
+	}
 	// Detect whether we still have the old WAL.
 	fns, err := sequenceFiles(dir)
 	if err != nil && !os.IsNotExist(err) {
@@ -1222,7 +1225,8 @@ func MigrateWAL(logger log.Logger, dir string) (err error) {
 	if len(fns) == 0 {
 		return nil // No WAL at all yet.
 	}
-	// Check header of first segment.
+	// Check header of first segment to see whether we are still dealing with an
+	// old WAL.
 	f, err := os.Open(fns[0])
 	if err != nil {
 		return errors.Wrap(err, "check first existing segment")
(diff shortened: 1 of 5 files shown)


wal.go
```diff
@@ -1206,3 +1211,89 @@ func (r *walReader) decodeDeletes(flag byte, b []byte, res *[]Stone) error {
...
+			err = repl.Log(enc.Tombstones(s, b[:0]))
+		},
+	)
+	if decErr != nil {
+		return errors.Wrap(err, "decode old entries")
+	}
+	if err != nil {
+		return errors.Wrap(err, "write new entries")
+	}
+	if err := w.Close(); err != nil {
```
jkohen: Don't you want to close both WALs regardless of errors?

krasi-georgiev:
I was thinking that with more versions and more changes this repair/convert blocks might "contaminate" :hankey:  the code. Why not add this as part of the tsdb cli tool?

I have already  added [tsdb scan](https://github.com/prometheus/tsdb/pull/320) so  maybe something like `tsdb upgrade`

fabxc:
Because our stability guarantee is that you can just restart Prometheus
with the next minor and things will work. Telling people to use a CLI tool
is breaking that guarantee.

fabxc:

fabxc:

wal.go
```diff
@@ -173,6 +176,8 @@ func newCRC32() hash.Hash32 {
 }
 
 // SegmentWAL is a write ahead log for series data.
+//
+// DEPRECATED: use wal pkg combined with the record coders instead.
```
brian-brazil: s/coders/codex/

wal.go
```diff
@@ -1206,3 +1211,100 @@ func (r *walReader) decodeDeletes(flag byte, b []byte, res *[]Stone) error {
...
+
+	tmpdir := dir + ".tmp"
+	if err := os.RemoveAll(tmpdir); err != nil {
+		return errors.Wrap(err, "cleanup replacement dir")
+	}
+	repl, err := wal.New(logger, nil, tmpdir)
+	if err != nil {
+		return errors.Wrap(err, "open new WAL")
+	}
+	// We close it once already before as part of finalization.
```
brian-brazil: This doesn't read properly

wal.go
```diff
@@ -1206,3 +1211,100 @@ func (r *walReader) decodeDeletes(flag byte, b []byte, res *[]Stone) error {
...
+	if decErr != nil {
+		return errors.Wrap(err, "decode old entries")
+	}
+	if err != nil {
+		return errors.Wrap(err, "write new entries")
+	}
+	if err := repl.Close(); err != nil {
+		return errors.Wrap(err, "close new WAL")
+	}
+	if err := fileutil.Rename(tmpdir, dir); err != nil {
```
brian-brazil: This no longer works given that dir already exists and Rename no longer deletes the destination first.

fabxc:
Merging this into the branch for the new WAL so we can have a final pass over all changes together.

PR #341, opened 2018-05-31
Cleanup new blocks on 'CleanTombstones' faliure

codesome:
Addresses https://github.com/prometheus/prometheus/issues/4200


krasi-georgiev:
We should add a test as well.

block.go
```diff
@@ -470,7 +470,7 @@ Outer:
 
 // CleanTombstones will rewrite the block if there any tombstones to remove them
 // and returns if there was a re-write.
```
simonpasquier: The comment should be updated as we've got an additional return parameter.

db.go
```diff
@@ -835,34 +835,50 @@ func (db *DB) Delete(mint, maxt int64, ms ...labels.Matcher) error {
...
 	defer db.metrics.tombCleanTimer.Observe(time.Since(start).Seconds())
 
+	newUIDs := []ulid.ULID{}
+	defer func() {
+		// If any error is caused, we need to delete all the new directory created.
+		if err != nil {
+			for _, uid := range newUIDs {
+				dir := filepath.Join(db.Dir(), uid.String())
+				if err := os.RemoveAll(dir); err != nil {
+					level.Error(db.logger).Log("msg", "failed to delete block after failed `CleanTombstones`", "err", err)
```
simonpasquier: (suggestion) log the `dir` value.

block.go
```diff
@@ -480,14 +480,15 @@ func (pb *Block) CleanTombstones(dest string, c Compactor) (bool, error) {
 	})
 
 	if numStones == 0 {
-		return false, nil
+		return false, [16]byte{}, nil
```
simonpasquier: any reason why you return `[16]byte{}` instead of `ulid.ULID{}`?

codesome: Oops, didn't think about that. Changed.

block.go
```diff
@@ -469,8 +469,9 @@ Outer:
 }
 
 // CleanTombstones will rewrite the block if there any tombstones to remove them
-// and returns if there was a re-write.
-func (pb *Block) CleanTombstones(dest string, c Compactor) (bool, error) {
+// and returns a bool if there was a re-write. If there was a re-write,
+// it also returns the ulid of the new directory created.
```
fabxc: The doc comment doesn't make much sense to me. It says the method will rewrite the block, then it says it returns true _if_ there was a rewrite, implying in doesn't actually do this.

Also seems like this can simply be solved by returning the ULID of the result block, which may be the same one.

I know this isn't part of this PR, but I've issues with this API in general. Blocks are immutable, so any mutating function does not make sense to begin with. More so if it actually creates a new block, which makes the block object completely invalid after it returns. And why would I have to pass in a compactor for removing a subset of data from a single block?

@gouthamve

db.go
```diff
@@ -835,34 +835,50 @@ func (db *DB) Delete(mint, maxt int64, ms ...labels.Matcher) error {
...
 		}
 	}
 
 	if len(deleted) == 0 {
 		return nil
 	}
 
-	return errors.Wrap(db.reload(deleted...), "reload blocks")
+	err = errors.Wrap(db.reload(deleted...), "reload blocks")
+	return err
```
fabxc: Just `return errors.Wrap(db.reload(deleted...), "reload blocks")` is sufficient.

block.go
fabxc: AFAICS, it also doesn't re-read the meta file either, which makes the block object actually hold wrong data.

db.go
```diff
@@ -835,34 +835,50 @@ func (db *DB) Delete(mint, maxt int64, ms ...labels.Matcher) error {
...
-		if err != nil {
-			return errors.Wrapf(err, "clean tombstones: %s", b.Dir())
+		ok, uid, er := b.CleanTombstones(db.Dir(), db.compactor)
+		if er != nil {
+			err = errors.Wrapf(er, "clean tombstones: %s", b.Dir())
+			return err
 		}
 
 		if ok {
 			deleted = append(deleted, b.Dir())
```
fabxc: Can we rename this to `deleteable` aligned with the parameter of `reload`? It suggests those blocks are already gone, which briefly made me think we are wiping all rewritten blocks in the `defer` if a single on of them fails.

db.go
```diff
@@ -835,34 +835,50 @@ func (db *DB) Delete(mint, maxt int64, ms ...labels.Matcher) error {
...
 	db.mtx.RUnlock()
 
 	deleted := []string{}
 	for _, b := range blocks {
-		ok, err := b.CleanTombstones(db.Dir(), db.compactor)
-		if err != nil {
-			return errors.Wrapf(err, "clean tombstones: %s", b.Dir())
+		ok, uid, er := b.CleanTombstones(db.Dir(), db.compactor)
+		if er != nil {
+			err = errors.Wrapf(er, "clean tombstones: %s", b.Dir())
```
fabxc: Same as below. Any particular reason why you want to split this up into two lines?

db.go
krasi-georgiev: I think this is because the defer block checks `err!=nil`

fabxc: Ohh, I see – that makes perfec. I think Go's semantics in this area are quite non-obvious and confusing at times. https://play.golang.org/p/RqSRG3vL4CU shows a few examples.

What I end up doing most of the time is naming the return argument. Then I can return whatever I want and defer will always bind to the actually returned value.
Unless you change scope, overwrite the named parameter with a new definition and call defer on that one (`f5`), which is quite unlikely in realistic control flow.
That seemed to be the most reliable variant for me so far.

codesome: As I see the example in the link (`f3`), a direct return will work. I will change then.

db.go
codesome: `f5` in the code snippet that you shared. (Yes, you already mentioned :))

db.go
fabxc: Yes, but only if the return param becomes named :)

krasi-georgiev: `f3()` looks most logical and IIUC this is the one you suggest as well.

`f5()` is really awkward , took few head scratches to digest :stuck_out_tongue_winking_eye:

codesome:
@fabxc did you have anymore refactoring in mind?

db.go
```diff
@@ -835,34 +835,49 @@ func (db *DB) Delete(mint, maxt int64, ms ...labels.Matcher) error {
...
+			return err
 		}
 
-		if ok {
-			deleted = append(deleted, b.Dir())
+		if uid != nil { // New block was created.
+			deletable = append(deletable, b.Dir())
+			newUIDs = append(newUIDs, *uid)
 		}
```
krasi-georgiev: another way of writing this if you wanted to to save few lines would be:
```
if uid, er := b.CleanTombstones(db.Dir(), db.compactor); er != nil {
	err = errors.Wrapf(er, "clean tombstones: %s", b.Dir())
	return err
} else if uid != nil { // New block was created.
	deletable = append(deletable, b.Dir())
	newUIDs = append(newUIDs, *uid)
}
```

codesome:
@fabxc @gouthamve, this is ready for another review.

krasi-georgiev:
@codesome you have one extra file in there.

testdata/repair_index_version/lock

PR #343, opened 2018-06-06
Added storage size based retention method and new metrics

### LINKED ISSUE #124: disk size based retention
With talk about 2.0 defaulting to +Inf retention time, there is also talk about adding a flag which does retention based on the data size on disk. Do we want to support that?

If yes, a simple hacky approach would be to check the size of a block after persisting and then updating meta.json with that info. On startup and reload, update the metric. But one obvious problem with that is the WAL size.

mknapphrt:
Added methods needed to retain data based on a byte limitation rather than time. Limitation is only applied if the flag is set (defaults to 0). Both blocks that are older than the retention period and the blocks that make the size of the storage too large are removed.

2 new metrics for keeping track of the size of the local storage folder and the amount of times data has been deleted because the size restriction was exceeded.

This PR is in conjunction with a PR with prometheus/prometheus. The idea is to allow a user to set a maximum number of bytes that the local storage can take up, in conjunction with the retention period, and whenever that limit is exceeded the oldest block will be deleted to regain space.

closes #124
closes https://github.com/prometheus/prometheus/issues/3684

Signed-off-by: Mark Knapp <mknapp@hudson-trading.com>


mknapphrt:
Prometheus PR: https://github.com/prometheus/prometheus/pull/4230

db.go
```diff
@@ -45,6 +45,7 @@ import (
 var DefaultOptions = &Options{
 	WALFlushInterval:  5 * time.Second,
 	RetentionDuration: 15 * 24 * 60 * 60 * 1000, // 15 days in milliseconds
+	MaxBytes:          0, // Don't use size based retention by default, uses 0 instead of -1 in case it is unintialized
```
free: This seems confusing. (I for one don't understand what the comment is saying.)

Why not have both this and the flag default to zero?

mknapphrt: I suppose it wasn't very clear, when running it through the tests it failed because I was assuming the default would be -1, indicating that I don't want to limit the data storage by size. But one of the test supplied an options of 

```
// reopen DB from snapshot
	db, err = Open(snap, nil, nil, &Options{
		RetentionDuration: 10,
		BlockRanges:       []int64{50},
	})
```
, which didn't include the MaxBytes field, and so MaxBytes was assigned 0 instead of -1. So it was attempting to limit the storage size to 0. I changed it to 0 now as the default, to indicate no limit, in the case that MaxBytes isn't supplied in the options when opening the database. I can't really think of a reasonable use case for 0 as the MaxBytes so I don't think this will be an issue. The flag didn't get changed to 0 too simply because i forgot to do it, I'll make that adjustment now.

free: Thanks, it's clearer now.

You should probably also drop the -1 from the comment too, to avoid any confusion.

db.go
```diff
@@ -460,6 +500,59 @@ func retentionCutoffDirs(dir string, mint int64) ([]string, error) {
...
+
+	return delDirs, nil
+
+}
+
+func dirSize(path string) (int64, error) {
+	var size int64
+	err := filepath.Walk(path, func(_ string, info os.FileInfo, err error) error {
+		if !info.IsDir() {
+			size += int64(info.Size())
```
juliusv: @brian-brazil pointed out on IRC that this will double-count blocks that are part of snapshots, which use hard links and thus appear twice in the FS but don't use twice the storage space. Somehow it has to be ensured that every real block is only counted once.

db.go
```diff
@@ -161,6 +167,14 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
 		Name: "prometheus_tsdb_tombstone_cleanup_seconds",
 		Help: "The time taken to recompact blocks to remove tombstones.",
 	})
+	m.storageBytes = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "prometheus_tsdb_storage_bytes",
+		Help: "The number of bytes that are currently used for local storage.",
+	})
+	m.dataLimitDeletions = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "prometheus_tsdb_data_limit_deletions",
```
juliusv: Hmm, how about `prometheus_tsdb_data_limit_deletions` -> `prometheus_tsdb_size_limit_deletions`?

db.go
```diff
@@ -312,14 +333,33 @@ func (db *DB) retentionCutoff() (b bool, err error) {
 		return false, nil
 	}
 
+	// Size based retention, if MaxBytes is still -1, it has not been set by the user
+	// and only time based retention will be used.
+	var dirsBySize []string
+	if (db.opts.MaxBytes > 0){
```
juliusv: You don't need parentheses around `if` expressions in Go. Same below.

Also, it doesn't seem like all the code has been formatted correctly, could you run `gofmt` (or configure your editor to run it automatically on save)?

db.go
```diff
@@ -460,6 +500,59 @@ func retentionCutoffDirs(dir string, mint int64) ([]string, error) {
...
+	defer df.Close()
+
+	dataDirSize, err := dirSize(dir)
+	if err != nil {
+		return nil, err
+	}
+
+	delDirs := []string{}
+	if (dataDirSize <= limit) {
+		return delDirs, nil
```
juliusv: If the list is empty, you can just return `nil` here for it and define the variable `delDirs` further below where it's actually needed.

db.go
```diff
@@ -460,6 +500,59 @@ func retentionCutoffDirs(dir string, mint int64) ([]string, error) {
...
+	}
+	// Limit has been exceeded
+	sizeToDelete := int64(0)
+
+	dirs, err := blockDirs(dir)
+	for _, dir := range dirs {
+
```
juliusv: style nit: no empty lines at the beginning of blocks

juliusv:
I just left comments for things that stood out to me, don't treat it as an exhaustive review though, since I'm not a TSDB expert :)

(could you update the PR description to say `0` instead of `-1`?)

db.go
```diff
@@ -312,14 +333,33 @@ func (db *DB) retentionCutoff() (b bool, err error) {
...
 
 	mint := last.Meta().MaxTime - int64(db.opts.RetentionDuration)
-	dirs, err := retentionCutoffDirs(db.dir, mint)
+	dirsByTime, err := retentionCutoffDirs(db.dir, mint)
 	if err != nil {
 		return false, err
 	}
 
+	// Remove blocks that fall under either retention criteria
+	dirs := append(dirsBySize, dirsByTime...)
```
juliusv: I think this needs to dedupe blocks in case a block was both too old and needs to be deleted for size too? Otherwise in `reload()` it'll try to delete it twice and log a warning, IIUC.

db.go
```diff
@@ -460,6 +500,59 @@ func retentionCutoffDirs(dir string, mint int64) ([]string, error) {
...
+	}
+
+	delDirs := []string{}
+	if (dataDirSize <= limit) {
+		return delDirs, nil
+	}
+	// Limit has been exceeded
+	sizeToDelete := int64(0)
+
+	dirs, err := blockDirs(dir)
```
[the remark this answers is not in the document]
mknapphrt: I was under the impression that it did return blocks ordered by time range because in the code for retentionCutoffDirs it calls blockDirs, then passes through each block and adds it to the list of blocks to be deleted until it finds one that is after the cutoff time. There's a comment saying
```
// The first block we encounter marks that we crossed the boundary
// of deletable blocks.
if meta.MaxTime >= mint {
	break
}
```
If it isn't in order, wouldn't this also not function properly?

db.go
mknapphrt: I could dedupe it, but the reload method passes through each block in the directory and then checks if it's in the 'deleteable' list that's passed into reload, and if so then deletes it. So I don't think it should ever try to delete a block twice.

db.go
mknapphrt: Would it be inappropriate to assume that snapshots will be contained to db.dir()/snapshot/... where they're originally made and that the user wouldn't move things around? It would keep complexity way down if that was fair to assume that.

db.go
```diff
@@ -312,14 +333,33 @@ func (db *DB) retentionCutoff() (b bool, err error) {
 		return false, nil
 	}
 
+	// Size based retention, if MaxBytes is still -1, it has not been set by the user
```
simonpasquier: nit but the comment isn't accurate anymore. How about "..., if MaxBytes is less than or equal to 0, only the time based retention will be used"?

db.go
simonpasquier: (nit) a `_total` suffix is expected too.

db.go
simonpasquier: The other option would be to iterate over `blockDirs(db.dir)` to calculate how much space the blocks use on disk and store their respective sizes in a map. This would ensure that we don't account for snapshots (or anything else). Then if the total size exceeds the limit, remove directories based on the values stored in the map.

db.go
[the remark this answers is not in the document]
simonpasquier: Block directories are [ULID](https://github.com/oklog/ulid) which are lexicographically sortable identifiers.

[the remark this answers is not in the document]
juliusv: Ah great, didn't know that :)

krasi-georgiev:
I am not an expert either, but can have a look in few days.

db.go
brian-brazil: du is not a great way to track how much space you're using. Instead track the actual sizes of the blocks that are loaded in.

mknapphrt: @brian-brazil filepath.Walk uses a stat call to get the size. I don't really understand what you're recommending for getting the size, do you think you could explain a little more?

brian-brazil: Each block should know its own sized based on the files that constitute it. Track that and work from it, no walking should be required.

mknapphrt: Are you suggesting using the meta data to calculate the size of a block? I don't really know a lot about how the blocks are created/kept track of.

brian-brazil: I don't believe we have that metadata currently, but you could stat the blocks as they are loaded.

mknapphrt: Do block sizes ever change after they're created? When they compact they get a new ULID, right? So it wouldn't be the same block anymore.

brian-brazil: No, blocks are immutable.

krasi-georgiev: @brian-brazil do you mean something like this?
https://www.socketloop.com/tutorials/get-filesize-go

this means we need to read the size of  all WAL files ,  index , tombstone and all chunks files.

krasi-georgiev: > I don't believe we have that metadata currently, but you could stat the blocks as they are loaded.

I don't quite understand what you mean here. when we start Prometheus only the WAL is loaded 
No other blocks are loaded until we actually query some data.

Could you point to some code to make this a bit more clear.

krasi-georgiev: I hope @mknapphrt understood what you mean as I still don't see it.

When we open the tsdb blocks it populates the meta file ,  the index , chunk and tombstone readers, but at first glance can't see an easy way to get the sizes from there.

juliusv: Based on [this comment](https://github.com/prometheus/tsdb/pull/343#discussion_r195726392) it seems like @brian-brazil is suggesting that while we currently only load the metadata etc. at startup, we should add step to that stratup procedure that stats the block and remembers the value.

mknapphrt: I was of the same undestanding as juliusv, I updated the PR to implement a "cache" of the sizes of each block when they're loaded now

krasi-georgiev: I thought he means to read the actual total bytes size when opening a block  rather than using `info.Size()`.

brian-brazil: More exactly, it shouldn't require a Walk to get the sizes. We should know the file names in advance that we're working with.

mknapphrt: When the db reloads, it does at some point figure out the names of the sequence files for each chunk and the wal, but it doesn't store them anywhere. It makes use of ``` func sequenceFiles(dir string) ([]string, error) ``` that reads the directory, but then just "forgets" them. Are you saying to use that function in place of the walk? Or that when the chunk/wal read the sequence files they store the sequences somewhere? And then stat those stored sequences? I don't see any other way of knowing the file names.

brian-brazil: Something like that sounds fine for me, but didn't we say we'd ignore the WAL for the first version to keep things simple?

simonpasquier: Maybe it would work if you'd store the size of each block whenever it gets opened (eg `tsdb.OpenBlock()` call). Then `*db.reload()` would compute the overall sum across all loaded blocks.

Some proof-of-concept here: https://github.com/simonpasquier/tsdb/commit/3c08afbf679abd2ea1647093d2cae17ea1829154

mknapphrt: I made some changes (similar to what Simon suggested/posted). When a block is loaded it now reads the necessary files and stats them and stores it in each block. If you think this should approach needs some reworking (which is likely), let me know.

mknapphrt: The test fails because it checks some blocks after doing a reload but it doesn't provide the size of the block when it checks it, but reloading causes the size to be calculated it. So the blocks that exists and the block meta that it's comparing to don't match.

block.go
```diff
@@ -162,6 +162,7 @@ type BlockStats struct {
 	NumSeries     uint64 `json:"numSeries,omitempty"`
 	NumChunks     uint64 `json:"numChunks,omitempty"`
 	NumTombstones uint64 `json:"numTombstones,omitempty"`
+	NumBytes      int64  `json:"-"`
```
brian-brazil: Maybe we should make this one work like the rest?

db.go
```diff
@@ -57,6 +58,9 @@ type Options struct {
 	// Duration of persisted data to keep.
 	RetentionDuration uint64
 
+	// Maximum number of bytes to be retained.
```
brian-brazil: bytes in blocks

db.go
```diff
@@ -161,6 +169,14 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
 		Name: "prometheus_tsdb_tombstone_cleanup_seconds",
 		Help: "The time taken to recompact blocks to remove tombstones.",
 	})
+	m.storageBytes = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "prometheus_tsdb_storage_bytes",
```
brian-brazil: for blocks

db.go
```diff
@@ -312,19 +331,42 @@ func (db *DB) retentionCutoff() (b bool, err error) {
...
+		msg := fmt.Sprintf("data limit was exceeded and %d records will be deleted", len(dirsBySize))
+		level.Warn(db.logger).Log("msg", msg)
+	}
+
+	// Time based retention
 	last := blocks[len(db.blocks)-1]
 
 	mint := last.Meta().MaxTime - int64(db.opts.RetentionDuration)
-	dirs, err := retentionCutoffDirs(db.dir, mint)
+	dirsByTime, err := retentionCutoffDirs(db.dir, mint)
```
brian-brazil: The code around here just got changed under you, so you'll need to update it.

chunks/chunks.go
```diff
@@ -285,10 +285,13 @@ type Reader struct {
 	cs []io.Closer
 
 	pool chunkenc.Pool
+
+	// The size of the chunks
+	size int64
```
brian-brazil: I don't think you're actually using this anywhere

db.go
mknapphrt: So something more like ```"prometheus_tsdb_storage_blocks_bytes"```?

brian-brazil: yes

chunks/chunks.go
mknapphrt: It's being set when you create a new dir reader. ``func NewDirReader(dir string, pool chunkenc.Pool)`` is where the size is calculated and then passed along when it calls ``newReader(bs, cs, pool, sz)``

mknapphrt:
I rebased the code and changed my code to better match how deletions are being done now. I fixed one or two things on github's editor to fix conflicts and didn't add a signature to those commits. Not sure how to fix that now cause it's still complaining about them.

block.go
```diff
@@ -162,6 +162,14 @@ type BlockStats struct {
 	NumSeries     uint64 `json:"numSeries,omitempty"`
 	NumChunks     uint64 `json:"numChunks,omitempty"`
 	NumTombstones uint64 `json:"numTombstones,omitempty"`
+	NumBytes      int64  `json:"numBytes,omitempty"`
+}
+
+// BlockDesc describes a block by ULID and time range.
```
simonpasquier: You need to remove L168-174 to fix the test.

simonpasquier:
> Not sure how to fix that now cause it's still complaining about them.

@mknapphrt No need to worry about it now, you can squash the commit once the PR is ready to go.

block.go
mknapphrt: Not quite sure how I missed that, thanks though

mknapphrt: This test will still fail though because of this test in the `TestDB_reloadOrder` function: 
```
	testutil.Equals(t, *metas[1], blocks[0].Meta())
``` 
on line 92, because after it reloads the block it calculates the size. Should I go in and modify that test then so it checks against the correct size?

mknapphrt:
So it turns out that @juliusv was right before, even though ULID's are lexicographically sortable identifiers as was pointed out, the time they are generated doesn't coincide with the time range of the block itself, because when they are compacted they get a new ULID which is more recent. I changed it so that it first sorted the blocks by their times ranges and then decided which blocks to delete.

mknapphrt:
Any opinions? @brian-brazil Is this more what you were thinking of instead of walking over the directory?

db.go
```diff
@@ -161,6 +168,14 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
 		Name: "prometheus_tsdb_tombstone_cleanup_seconds",
 		Help: "The time taken to recompact blocks to remove tombstones.",
 	})
+	m.storageBytes = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "prometheus_tsdb_storage_blocks_bytes",
+		Help: "The number of bytes that are currently used for local storage.",
```
brian-brazil: ...by blocks

db.go
```diff
@@ -477,6 +493,36 @@ func (db *DB) reload() (err error) {
...
+		blocksSize += b.Size()
+	}
+	// Sort the blocks to make sure the oldest is removed first
+	sort.Slice(allBlocks, func(i, j int) bool {
+		return allBlocks[i].Meta().MinTime < allBlocks[j].Meta().MinTime
+	})
+	// Select blocks for deletion
+	for _, block := range allBlocks {
+		if db.opts.MaxBytes > 0 && blocksSize > db.opts.MaxBytes {
+			fmt.Println(block.String())
```
brian-brazil: Debug code?

mknapphrt:
Thanks for all the feedback guys, I really appreciate it. I've gotten a couple responses saying they're not the tsdb expert, so I was wondering who is the tsdb expert that I could bother to look over this? @juliusv  @krasi-georgiev Thanks again

juliusv:
@mknapphrt The biggest experts here would be @fabxc and @gouthamve according to https://github.com/prometheus/tsdb/graphs/contributors

mknapphrt:
Doesn't seem there's been much activity around here recently, but I just wanted to try nudging @fabxc and @gouthamve  to see if you guys could take a look over this. Thanks

juliusv:
Maybe @gouthamve has more time after https://twitter.com/putadent/status/1026013754251653120 now :)

krasi-georgiev:
@mknapphrt I can have a look again after Promcon as I already have a use case for this so would be great so see it merged.

mknapphrt:
Congrats @gouthamve! And thanks @krasi-georgiev, I appreciate it

krasi-georgiev:
@mknapphrt we would need some test to make sure the size based retention kicks in when expected.

mknapphrt:
I'm not sure what exactly you guys are looking for in testing. I've been running an instance for ~a month with a fairly low cap (100 MB) and scraping a few node exporters. It's been bouncing up and down between ~92 and 100MB for the last month depending on the timing of the deletions and the compactions. It's done 2851 deletions so far and stores about 3.5 hours of data at any given moment.

If you've got a good test idea please let me know and I'd be happy to spin it up

chunks/chunks.go
```diff
@@ -309,11 +312,11 @@ func newReader(bs []ByteSlice, cs []io.Closer, pool chunkenc.Pool) (*Reader, err
 }
 
 // NewReader returns a new chunk reader against the given byte slices.
-func NewReader(bs []ByteSlice, pool chunkenc.Pool) (*Reader, error) {
+func NewReader(bs []ByteSlice, pool chunkenc.Pool, size int64) (*Reader, error) {
 	if pool == nil {
 		pool = chunkenc.NewPool()
 	}
-	return newReader(bs, nil, pool)
+	return newReader(bs, nil, pool, size)
```
krasi-georgiev: why do you need to take the size an argument to `func NewReader(bs []ByteSlice, pool chunkenc.Pool, size int64)` ?

why not just pass 0  to `return newReader(bs, nil, pool, 0)`

db.go
```diff
@@ -477,17 +493,13 @@ func (db *DB) reload() (err error) {
...
 	}
-	// Load new blocks into memory.
+	// allBlocks := make([]*Block, 0)
+	// Get storage blocks total size
 	for _, dir := range dirs {
 		meta, err := readMetaFile(dir)
 		if err != nil {
 			return errors.Wrapf(err, "read meta information %s", dir)
 		}
-		// Don't load blocks that are scheduled for deletion.
```
krasi-georgiev: I think the logic after this is wrong. The `deleteable` blocks are left over after the compaction has merged these so these should be excluded from the check.
Anyway I want to try to simplify this as it is a bit difficult to follow.

db.go
```diff
@@ -161,6 +168,14 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
 		Name: "prometheus_tsdb_tombstone_cleanup_seconds",
 		Help: "The time taken to recompact blocks to remove tombstones.",
 	})
+	m.storageBytes = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "prometheus_tsdb_storage_blocks_bytes_total",
+		Help: "The number of bytes that are currently used for local storage by blocks.",
+	})
+	m.dataLimitDeletions = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "prometheus_tsdb_size_limit_deletions_total",
+		Help: "The number of times that blocks were deleted because the maximum number of bytes was exceeded.",
```
krasi-georgiev: Just curious, where do you think this metric could be useful?

chunks/chunks.go
```diff
@@ -329,22 +332,30 @@ func NewDirReader(dir string, pool chunkenc.Pool) (*Reader, error) {
...
 	}
-	return newReader(bs, cs, pool)
+	return newReader(bs, cs, pool, sz)
 }
 
 func (s *Reader) Close() error {
 	return closeAll(s.cs...)
 }
 
+func (s *Reader) Size() int64 { return s.size }
```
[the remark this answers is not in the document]
mknapphrt: Is that better though, recalculating the size every time that it gets called? The block size should never change so it seemed more efficient to just store it.

db.go
mknapphrt: I think less so the number of deletions that occurred, but more so that deletions are occurring. It would let the user know that the cap is being reached, and that data is removed because of the cap, not because of the retention period being hit.

db.go
mknapphrt: I don't really remember why or how that got deleted, definitely should still be in there.

chunks/chunks.go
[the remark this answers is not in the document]
krasi-georgiev: even if we have hundreds of blocks this would mean only few cpu cycles so wouldn't cause any overhead. Making the code a lot more clear to follow is a good trade off. 
After all golang is all about simplicity :smiley_cat:

krasi-georgiev:
> If  you've got a good test idea please let me know and I'd be happy to spin it up

I would imagine something like 
 1. create few blocks (can use `createEmptyBlock()`)
 1. compare `db size > retention settings` - to make sure we have something that needs deleting.
 1. trigger the retention (reload the db)
 1. compare `db size < retention settings` - to make sure retention works as expecte

`TestSizeBasedRetention`

chunks/chunks.go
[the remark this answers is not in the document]
krasi-georgiev: I wouldn't mind to hear another  opinion, but I really prefer simplicity over saving few CPU cycles.

some caching can be added if it shows some performance hit, but a first glance I doubt it will need any.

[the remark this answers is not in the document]
mknapphrt: Sorry for the late response, I'm fine with doing it inside of the Size() method, I'm not super partial to either approach. I feel it's a small difference in understandability/simplicitiy, but if it's better practice then I'm all for it.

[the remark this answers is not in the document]
krasi-georgiev: lets have a look how it will looks and will try to simplify more if possible.

mknapphrt:
So I added the test, I did it a little different than you suggested, the general idea was:

1. Create a test DB and add a couple empty blocks
2. Reload the db so that blocks are recognized by the db
3. Check that the `size of the blocks < limit` as well as ensure the correct number of blocks and oldest blocks were deleted
4. Adjust the limit to something smaller and test again

I can't actually check if the `db size > limit` because the retention policy kicks in while reloading the db.

krasi-georgiev:
Thanks I will have another look soon.

db_test.go
```diff
@@ -935,6 +935,50 @@ func TestDB_Retention(t *testing.T) {
...
 }
 
+func TestSizeBasedretention(t *testing.T) {
+	db, close := openTestDB(t, &Options{
+		BlockRanges: []int64{100},
+		MaxBytes:    1000,
+	})
+	defer close()
+
+	// Add some empty blocks, each empty block is 249 bytes.
```
krasi-georgiev: the only thing it worries me that the test will fail when the default block size changes.

What if you set the `db.Opts.MaxBytes`  after you create the empty block by using it `OpenBlock()` and than `block.Size()`.

mknapphrt:
`loadAll() returns: goodBlocks [ulid.ULID]*Block`

Is there a reason you suggested returning a map of ulids to blocks for the good blocks?

krasi-georgiev:
So we can pass it to filterDeletable() - and each block that is added to deletable should also be removed from the goodBlocks map

mknapphrt:
Right, it just feels redundant to have a map of the ULID to the block, when the block has the ULID already. Why not just a `[]*Block`? I suppose lookup is easier with the map.

mknapphrt:
I changed the test case to be more robust, let me know if you think it would still run into any issues. Also I split up the reload method, I didn't really see a reason to make a separate method for validateSequence, it seemed like a pretty short section already.

mknapphrt:
Just a gentle nudge @krasi-georgiev, wondering what you think of it right now

db.go
```diff
@@ -465,51 +508,75 @@ func (db *DB) reload() (err error) {
...
+	return filteredBlocks, deleteable, nil
+}
 
+func (db *DB) validateBlockSequence() {
+
+}
+
+func (db *DB) removeDeleteable(blocks []*Block, deleteable map[ulid.ULID]struct{}) error {
 	// Swap in new blocks first for subsequently created readers to be seen.
 	// Then close previous blocks, which may block for pending readers to complete.
```
krasi-georgiev: I think the next few lines(up to `db.mtx.Unlock()`) fit better as they were before in the `reload()` func.

db.go
```diff
@@ -519,28 +586,29 @@ func (db *DB) reload() (err error) {
...
-		if _, ok := opened[b.Meta().ULID]; ok {
+		if containsBlockWithULID(b.Meta().ULID, blocks) {
 			continue
 		}
 		if err := b.Close(); err != nil {
 			level.Warn(db.logger).Log("msg", "closing block failed", "err", err)
 		}
 	}
-	// Delete all obsolete blocks. None of them are opened any longer.
+	//Delete all obselete blocks. None of them are opened any longer.
```
krasi-georgiev: `// Delete all obsolete blocks. None of them are opened any longer.`

db.go
```diff
@@ -519,28 +586,29 @@ func (db *DB) reload() (err error) {
...
 	for ulid := range deleteable {
 		if err := os.RemoveAll(filepath.Join(db.dir, ulid.String())); err != nil {
-			return errors.Wrapf(err, "delete obsolete block %s", ulid)
+			return errors.Wrapf(err, "delete obsolete blocks %s", ulid)
```
krasi-georgiev: should be singular, the error is returned when removing a single block.

db.go
```diff
@@ -519,28 +586,29 @@ func (db *DB) reload() (err error) {
...
 		}
 	}
+	return nil
+}
 
-	// Garbage collect data in the head if the most recent persisted block
-	// covers data of its current time range.
-	if len(blocks) == 0 {
-		return nil
```
krasi-georgiev: why did you remove this comment and the check below?

db_test.go
```diff
@@ -79,9 +79,9 @@ func TestDB_reloadOrder(t *testing.T) {
 	defer db.Close()
 
 	metas := []*BlockMeta{
-		{ULID: ulid.MustNew(100, nil), MinTime: 90, MaxTime: 100},
-		{ULID: ulid.MustNew(200, nil), MinTime: 70, MaxTime: 80},
-		{ULID: ulid.MustNew(300, nil), MinTime: 100, MaxTime: 110},
+		{ULID: ulid.MustNew(100, nil), MinTime: 90, MaxTime: 100, Stats: BlockStats{NumBytes: 248}},
+		{ULID: ulid.MustNew(200, nil), MinTime: 70, MaxTime: 80, Stats: BlockStats{NumBytes: 247}},
+		{ULID: ulid.MustNew(300, nil), MinTime: 100, MaxTime: 110, Stats: BlockStats{NumBytes: 249}},
```
krasi-georgiev: I think this is a lot lekely to break like:

```
...
metas := []*BlockMeta{
		{ULID: ulid.MustNew(100, nil), MinTime: 90, MaxTime: 100},
		{ULID: ulid.MustNew(200, nil), MinTime: 70, MaxTime: 80},
		{ULID: ulid.MustNew(300, nil), MinTime: 100, MaxTime: 110},
	}
	for _, m := range metas {
		bdir := filepath.Join(db.Dir(), m.ULID.String())
		createEmptyBlock(t, bdir, m)
	}

	testutil.Ok(t, db.reload())
	blocks := db.Blocks()
	blocks[0].meta.Stats.NumBytes = 0
	blocks[1].meta.Stats.NumBytes = 0
	blocks[2].meta.Stats.NumBytes = 0
....
```

db_test.go
```diff
@@ -935,6 +935,55 @@ func TestDB_Retention(t *testing.T) {
...
+	}
+	for _, m := range blocks {
+		createEmptyBlock(t, filepath.Join(db.Dir(), m.ULID.String()), m)
+	}
+
+	// Figure out how large an empty block is.
+	emptyBlockDir := filepath.Join(db.Dir(), blocks[0].ULID.String())
+	emptyBlock, err := OpenBlock(emptyBlockDir, db.chunkPool)
+	testutil.Ok(t, err)
+	db.opts.MaxBytes = 4 * emptyBlock.Size()
```
krasi-georgiev: can't we use something like `len(blocks) * emptyBlock.Size()` ? (the same for few other places below.) This would be more predictable if we decide to test with more blocks.

also a small comment of what is this line for would be helpful.

db_test.go
```diff
@@ -935,6 +935,55 @@ func TestDB_Retention(t *testing.T) {
...
+
+	// Reload and ensure that actual size is less than limit.
+	testutil.Ok(t, db.reload())
+
+	var size int64
+	for _, b := range db.blocks {
+		size += b.Size()
+	}
+	testutil.Assert(t, size <= db.opts.MaxBytes, "actual size (%v) is expected to be less than or equal to limit (%v)", size, db.opts.MaxBytes)
+	testutil.Equals(t, 4, len(db.blocks))
```
krasi-georgiev: another comment here as well, why are we checking this?

db_test.go
```diff
@@ -935,6 +935,55 @@ func TestDB_Retention(t *testing.T) {
...
+	testutil.Equals(t, int64(300), db.blocks[0].meta.MaxTime) // Ensure oldest block was deleted.
+
+	// Reduce the size limit and test again.
+	db.opts.MaxBytes = 1 * emptyBlock.Size()
```
krasi-georgiev: why not just  : `db.opts.MaxBytes = emptyBlock.Size()` or `db.opts.MaxBytes -= emptyBlock.Size()`  :stuck_out_tongue_winking_eye:

db_test.go
mknapphrt: I can't tell, are you actually asking me why I'm checking this? Or just that the comment should answer that question?

db_test.go
mknapphrt: Yeah, I'm not exactly sure why I did that haha

db.go
```diff
@@ -465,51 +508,75 @@ func (db *DB) reload() (err error) {
...
 		return blocks[i].Meta().MinTime < blocks[j].Meta().MinTime
 	})
-	if err := validateBlockSequence(blocks); err != nil {
-		return errors.Wrap(err, "invalid block sequence")
+	// Select blocks for deletion.
+	for _, block := range blocks {
+		if db.opts.MaxBytes > 0 && blocksSize > db.opts.MaxBytes {
+			blocksSize -= block.Size()
+			deleteable[block.Meta().ULID] = struct{}{}
+			for _, b := range block.Meta().Compaction.Parents {
```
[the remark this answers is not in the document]
mknapphrt: The same check is being done earlier on for block beyond the retention period, and I didn't change that logic from the original reload method. The parents were added to deleteable there too, so it only made sense to do the same for blocks break size retention policy. Do you want both of those checks moved then? And are those block really "corrupt"? It doesn't seem like the right word to use for them, deleteable feels more appropriate.

db.go
```diff
@@ -465,51 +508,75 @@ func (db *DB) reload() (err error) {
...
+		}
 	}
+	// Filter out blocks that aren't scheduled for deletion
+	for ulid, block := range goodBlocks {
+		if _, ok := deleteable[ulid]; ok {
+			continue
+		}
+		filteredBlocks = append(filteredBlocks, block)
+	}
+	return filteredBlocks, deleteable, nil
```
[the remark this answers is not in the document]
mknapphrt: Those have different types though? goodBlocks is a `map[ulid.ULID]*Block`, while filteredBlocks was a `[]*Block`

db.go
```diff
@@ -437,22 +450,52 @@ func (db *DB) reload() (err error) {
...
+
+	if len(filteredBlocks) == 0 {
+		return nil
+	}
+	maxt := filteredBlocks[len(filteredBlocks)-1].Meta().MaxTime
+
+	return errors.Wrap(db.head.Truncate(maxt), "head truncate failed")
+}
+
+func (db *DB) loadAll() (goodBlocks map[ulid.ULID]*Block, corruptedBlocks map[ulid.ULID]error, err error) {
```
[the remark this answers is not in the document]
mknapphrt: loadableBlocks twice?

db_test.go
krasi-georgiev: yeah just a comment to specify why are we checking this.
 It is clear for me and you, but for someone not familiar with the test will wonder.

krasi-georgiev:
ping me when ready for another review.

db.go
```diff
@@ -125,6 +128,8 @@ type dbMetrics struct {
 	cutoffs              prometheus.Counter
 	cutoffsFailed        prometheus.Counter
 	tombCleanTimer       prometheus.Histogram
+	storageBytes         prometheus.Gauge
+	dataLimitDeletions   prometheus.Counter
```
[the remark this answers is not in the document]
krasi-georgiev: so  this is incremented once when the blocks reach max bytes
https://github.com/prometheus/tsdb/pull/343/files/6ae576b37c52a9544b98bec3ac198ce02b7e8482#diff-7a117cbfd5485ed35efcbb2995c37aa0R555

and also every time regardless of whether max bytes was reached
https://github.com/prometheus/tsdb/pull/343/files/6ae576b37c52a9544b98bec3ac198ce02b7e8482#diff-7a117cbfd5485ed35efcbb2995c37aa0R558


I thought that the metric is to specify how many times the size based retention happened.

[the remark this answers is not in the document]
mknapphrt: It is, that was a mistake that I found while looking at the names of these metrics the other day. It's only supposed to increment when the max bytes cap has been reached. I'll fix that in the next commit.

mknapphrt:
Uhh @krasi-georgiev, I got a check for go 1.9.x and an x for 1.10.x, but there's no job log for either of them.

db.go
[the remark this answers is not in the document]
krasi-georgiev: yes you are right lets move this check in `openBlocks` , but call the return variables `deleteable`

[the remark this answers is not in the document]
mknapphrt: I don't really follow why it should go in openBlocks if these are deleteable blocks. The purpose of `loadableBlocks` is to determine which blocks can be loaded, and which can be deleted. So figuring out what can be deleted outside of the method seems to be out of place.

db.go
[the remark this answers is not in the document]
mknapphrt: I don't think that will work though, because if you start modifying the thing that you're ranging over it's going to mess things up. For example, if index 3 and 4 are both 'deleteable', when it hits index 3, it's going to remove that value form goodBlocks, and everything after gets shifted down one index. So then it the checks index 4, but index 4 is now the original index 5, and the original index 4 is now in index 3, so it will be skipped over.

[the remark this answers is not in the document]
mknapphrt: `goodBlocks` can still be a slice, it doesn't need to be a map, but removing certain values like that will lead to issues.

db.go
```diff
@@ -437,22 +477,50 @@ func (db *DB) reload() (err error) {
 		db.metrics.reloads.Inc()
 	}()
 
+	goodBlocks, corruptedBlocks, err := db.openBlocks()
+	if err != nil {
+		return err
+	}
+	loadableBlocks, deleteableBlocks, err := db.loadableBlocks(goodBlocks, corruptedBlocks)
```
krasi-georgiev: goodBlocks , corruptedBlocks , deleteableBlocks, loadableBlocks : yeah the logic here is still a bit hard to follow.If you don't mind give me few days and I will think how to simplify.

block.go
```diff
@@ -270,6 +273,8 @@ func OpenBlock(dir string, pool chunkenc.Pool) (*Block, error) {
 		return nil, err
 	}
 
+	updateBlockSize(logger, dir, cr, meta)
```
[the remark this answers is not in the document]
mknapphrt: It would definitely work, I'm just not sure what the benefit is per se.

db.go
mknapphrt: Haha, yeah, it's definitely a little hard to follow

block.go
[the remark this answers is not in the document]
mknapphrt: Nope, I'm convinced. Just curious what the motivation was :)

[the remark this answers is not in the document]
mknapphrt: Should `updateBlockSize` still be a function that separates the blockSize call and errors from `OpenBlock`?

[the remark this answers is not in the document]
krasi-georgiev: sorry didn't understand?

why would we need the `updateBlockSize` ?

the meta write hapens in the else statement if this is what you mean.
```
else {
    meta.Stats.NumBytes = bs
    err := writeMetaFile(dir, meta)
    if err != nil {
		level.Warn(logger).Log("msg", "couldn't write the meta file for the block size", "block", dir, "err", err)
     }
}
```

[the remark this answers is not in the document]
mknapphrt: Also, one more thing. The meta, ir, and tr would all need to know the dir too, wouldn't they?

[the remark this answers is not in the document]
mknapphrt: I didn't know if you wanted to have this logic
```
bs,err:=blockSize( cr, ir, tr, meta) ; err != nil {
   level.Warn(logger).Log("msg", "couldn't get block size", "block", dir, "err", err)
} else {
    meta.Stats.NumBytes = bs
    err := writeMetaFile(dir, meta)
    if err != nil {
		level.Warn(logger).Log("msg", "couldn't write the meta file for the block size", "block", dir, "err", err)
     }
}
```

isolated from the OpenBlock method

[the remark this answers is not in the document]
krasi-georgiev: yeah. I think so

 for example:

 memTombstones should have a new field dir 
```
readTombstones(dir string) (*memTombstones, error) {
memTombstones.dir ==dir

func (t *memTombstones) Size() (int64,err){
   stat t.dir .....
}
...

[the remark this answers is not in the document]
Comment (krasi-georgiev) [reply block.go]: actually don't even need a new interface just add the Size function to 

```
// BlockReader provides reading access to a data block.
type BlockReader interface {
        // Size returns the block size in bytes.
        Size() int64,error
	// Index returns an IndexReader over the block's data.
	Index() (IndexReader, error)

	// Chunks returns a ChunkReader over the block's data.
	Chunks() (ChunkReader, error)

	// Tombstones returns a TombstoneReader over the block's deleted data.
}
```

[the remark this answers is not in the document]
Comment (krasi-georgiev) [reply block.go]: > I didn't know if you wanted to have this logic

> bs,err:=blockSize( cr, ir, tr, meta) ; err != nil {
   level.Warn(logger).Log("msg", "couldn't get block size", "block", dir, "err", err)
} else {
    meta.Stats.NumBytes = bs
    err := writeMetaFile(dir, meta)
    if err != nil {
		level.Warn(logger).Log("msg", "couldn't write the meta file for the block size", "block", dir, "err", err)
     }
}
> ... isolated from the OpenBlock method


this is small enough to keep it there.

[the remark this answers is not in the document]
Comment (mknapphrt) [reply block.go]: Why wouldn't you need a new interface? ChunkReader, IndexReader, BlockMeta and TombstoneReader don't implement BlockReader, so adding Size() there wouldn't allow us to call size on cr, ir, tr, and meta.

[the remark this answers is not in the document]
Comment (krasi-georgiev) [reply block.go]: yes you are right , sorry.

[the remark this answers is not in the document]
Comment (mknapphrt) [reply block.go]: So for adding a `Size()` function to BlockMeta, it would need the dir to do a stat, which is stored in Block. It doesn't really seem to make sense to store `dir` in BlockMeta, but I'm not sure how to get that path to it without changing the `Size()` to be `Size(dir)`. Any suggestions?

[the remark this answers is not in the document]
Comment (krasi-georgiev) [reply block.go]: Yeah I just looked as well and it is a tricky one.
To be honest I don't think we should even include the meta files when calculating the size. These are only 200-300 bytes.

File: db.go (hunk 35 above)
Comment (krasi-georgiev) [reply db.go]: ping me when you address my other comments and I will think how to simplify that part again.

**Comment by krasi-georgiev** [comment]
I have fixed most nits and simplified the names and the workflow.
https://github.com/mknapphrt/tsdb/pull/2

Now just need to try the last suggestion for implementing the Size() interface.

**Comment by mknapphrt** [comment]
@krasi-georgiev Apologies for the delay, I've been really busy for the last few days. I'll take a look at it soon and push the Size() changes

**Comment by krasi-georgiev** [comment]
Sure no problem. 
If you don't think you would have time let me know and I can complete the rest of the changes.

File: block.go (hunk 36 above)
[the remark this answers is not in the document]
Comment (mknapphrt) [reply block.go]: For the tombstones, I'm running into another hiccup. I changed the signature of `NewMemTombstones()` to `NewMemTombstones(dir string)` so that I could stat the tombstones. But there's several references to `NewMemTombstones`, one is in `NewHead`, and in `LoopChunkSeries`. Neither of those functions has the data dir currently. 

I could modify `NewHead` to take in dir, because the only reference to `NewHead` comes from `block.Open` which has the dir. `LoopChunkSeries` is a bit weirder, it doesn't seem to have a great option for supplying the dir. One (not very good) option is to take dir from the `IndexReader` that now stores the index, which is in the `blockQuerier`. I don't really like this approach at all.

One other option is to change the Size() function to take in a dir. It doesn't seem logical for a size function to need to take in the dir, but it simplifies a lot of this and results in less changes in other places.

Any thoughts?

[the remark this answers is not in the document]
Comment (krasi-georgiev) [reply block.go]: I just had another look and `MemTombstones` doesn't even look like a good place for this. Since these are in memory struct so it is not a very logical to hold disk file size information.

This is what I came up:
```
// SizeReader .....
type SizeReader interface {
	Size() int
}

// TombstoneFile .....
type TombstoneFile struct {
	size int
}

// Size .....
func (t *TombstoneFile) Size() int {
	return t.size
}

func readTombstones(dir string) (*memTombstones, SizeReader, error) {
	b, err := ioutil.ReadFile(filepath.Join(dir, tombstoneFilename))
	if os.IsNotExist(err) {
		return NewMemTombstones(), nil, nil
	} else if err != nil {
		return nil, nil, err
	}

	sr := &TombstoneFile{
		size: len(b),
	}
```

File: db.go [hunk 37]
```diff
@@ -428,7 +471,6 @@ func (db *DB) getBlock(id ulid.ULID) (*Block, bool) {
 
 // reload on-disk blocks and trigger head truncation if new blocks appeared. It takes
 // a list of block directories which should be deleted during reload.
-// Blocks that are obsolete due to replacement or retention will be deleted.
```
krasi-georgiev: this comment needs to be updated.

db.go
```diff
@@ -437,110 +479,148 @@ func (db *DB) reload() (err error) {
...
+	return loadable, corrupted, nil
+}
 
-	// Drop old blocks from memory.
-	for _, b := range oldBlocks {
-		if _, ok := opened[b.Meta().ULID]; ok {
-			continue
+// deletableBlocks returns all blocks past retention policy
+// and blocks superseeded by parents after a compaction.
+func (db *DB) deletableBlocks(loadable []*Block) map[ulid.ULID]*Block {
```
krasi-georgiev: we can rename this to just `blocks []*Block`

db.go
```diff
@@ -437,110 +479,148 @@ func (db *DB) reload() (err error) {
...
-	// Delete all obsolete blocks. None of them are opened any longer.
-	for ulid := range deleteable {
+
+	return deletable
+}
+
+// deleteBlocks closes and deletes blocks from the disk.
+// When the map contains a non nil block object it means it is loaded in memory
+// so needs to be closed first as it might need to wait for pending readers to complete.
+func (db *DB) deleteBlocks(deleteable map[ulid.ULID]*Block) error {
```
krasi-georgiev: we can rename this to just `blocks []*Block`

db.go
```diff
@@ -437,110 +479,148 @@ func (db *DB) reload() (err error) {
...
 		}
 	}
+	return nil
+}
 
-	// Garbage collect data in the head if the most recent persisted block
-	// covers data of its current time range.
-	if len(blocks) == 0 {
-		return nil
+func containsBlockWithULID(ulid ulid.ULID, blocks []*Block) bool {
```
krasi-georgiev: don't need this anymore

block.go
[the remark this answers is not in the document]
krasi-georgiev: probably can use the same logic for the meta file in `readMetaFile`

krasi-georgiev:
not sure if you saw my comment so posting it here again.

I just had another look and `MemTombstones` doesn't even look like a good place for this. Since these are in memory struct so it is not a very logical to hold disk file size information.

This is what I came up:
```
// SizeReader .....
type SizeReader interface {
	Size() int
}

// TombstoneFile .....
type TombstoneFile struct {
	size int
}

// Size .....
func (t *TombstoneFile) Size() int {
	return t.size
}

func readTombstones(dir string) (*memTombstones, SizeReader, error) {
	b, err := ioutil.ReadFile(filepath.Join(dir, tombstoneFilename))
	if os.IsNotExist(err) {
		return NewMemTombstones(), nil, nil
	} else if err != nil {
		return nil, nil, err
	}

	sr := &TombstoneFile{
		size: len(b),
	}
```

probably can use the same logic for the meta file in readMetaFile

db.go
```diff
@@ -305,6 +320,34 @@ func (db *DB) beyondRetention(meta *BlockMeta) bool {
...
+	})
+
+	deleteable = make(map[ulid.ULID]*Block)
+
+	var blocksSize = int64(0)
+
+	for i, block := range blocks {
+		if db.opts.MaxBytes > 0 && blocksSize+block.Size() > db.opts.MaxBytes {
+			// Add this and all previous blocks for deletion.
+
```
krasi-georgiev: remove extra new line

db_test.go
```diff
@@ -935,6 +938,57 @@ func TestDB_Retention(t *testing.T) {
 	testutil.Equals(t, int64(100), db.blocks[0].meta.MaxTime) // To verify its the right block.
 }
 
+func TestSizeBasedRetention(t *testing.T) {
+	db, close := openTestDB(t, &Options{
+		BlockRanges: []int64{100},
+	})
+	defer close()
```
krasi-georgiev: defer db.Close()

krasi-georgiev:
just figured the problem with the failing tests.

emptyBlock size is calculated incorrectly because of the meta file size.

1. Metafile size is added to the total Block size (BlockStats: NumBytes - is empty)
1. Block size total is written to the metafile. (BlockStats: NumBytes - has some value)
1. Now the metafile size has increased (BlockStats: NumBytes - occupies  few extra bytes) so the next time we open the same block it will report a bigger metafile size.

Long story short if we want 100% accurate size we should NOT read the existing metafile on disk, but calculate the size that would be written after updating it with the new BlockStats: NumBytes 

anyway since this would add extra lines of code and considering that the metafile would always be very small I think we should omit it all together from the Block size total.

mknapphrt:
Welp, rebasing to fix the merge conflict added a whole bunch of commits here that I wasn't expecting, but the current pr hopefully reflects the changes you were thinking of @krasi-georgiev

chunks/chunks.go
```diff
@@ -345,6 +345,14 @@ func (s *Reader) Close() error {
 	return closeAll(s.cs...)
 }
 
+func (s *Reader) Size() (int64, error) {
```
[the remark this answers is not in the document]
mknapphrt: The only reason I used int64 for everything is because os.Stat(...) returns FileInfo, and the Size() function on that returns int64, not int.

block.go
```diff
@@ -155,6 +163,9 @@ type BlockMeta struct {
 
 	// Version of the index format.
 	Version int `json:"version"`
+
```
krasi-georgiev:
```suggestion
```

block.go
```diff
@@ -155,6 +163,9 @@ type BlockMeta struct {
...
+	// The directory of the meta file, used for getting the block size.
```
krasi-georgiev:
```suggestion
```

index/index.go
```diff
@@ -935,6 +938,15 @@ func (r *Reader) SortedPostings(p Postings) Postings {
 	return p
 }
 
+// Size returns the size of an index file.
+func (r *Reader) Size() (int64, error) {
+	fInfo, err := os.Stat(r.dir)
+	if err != nil {
+		return -1, err
+	}
+	return fInfo.Size(), nil
```
krasi-georgiev: double check the implementation again , but I am pretty sure you don't need all this.

can just simply
`return int64(size), nil`

mknapphrt: I think you're right. I still think using int64 may be necessary for someone on a 32 bit system that's collecting a lot of metrics. If there's enough chunks they'll exceed the max size of an int, won't they?

krasi-georgiev: yes you are right , but I think we can to this only at the end in.
`func blockSize(rr ...SizeReader) (int64, error)`


note to myself: max file size on a 32 bit system is 4gb (4294967296 bytes) , int32 max is 2147483647 , uint32 max is 4294967295

mknapphrt: I'm not so sure because the majority of the data is in the chunks. If the overflow were to occur, wouldn't it most likely happen in `Size()` of the chunkreader?

krasi-georgiev: oh yeah that is true, leave as is than.

krasi-georgiev: ping me when ready for another review.

krasi-georgiev: probably time to sign the commits as well.

mknapphrt:
@krasi-georgiev I'll squash the last commit if you think this looks alright.

krasi-georgiev:
you still have many references to a `dir` variable. This is not needed anymore anywhere.

mknapphrt:
lmk if I missed any

mknapphrt:
@krasi-georgiev Just wanted to double check that you saw the last pushes.

krasi-georgiev:
yep I missed the notification somehow, thanks for the reminder.

There is one more problem that I just spotted. I wouldn't expect that setting the block size would happen while opening it,I think it makes the most sense to be in

```
if err := c.populateBlock(blocks, meta, indexw, chunkw); err != nil {
	return errors.Wrap(err, "write compaction")
}

if err = writeMetaFile(tmp, meta); err != nil {
	return errors.Wrap(err, "write merged meta")
}
```

so `populateBlock` will update the meta file and `writeMetaFile` will just write it.

I had a quick look and I think if you add `Size()` to the IndexReader,ChunkReader,TombstoneReader interfaces should be easy to implement.

mknapphrt:
I don't know a whole lot about the compactor, but i'm not sure if that method will be run when we want it to. The comment at the top says:

```
```

Does this mean that the compactor is run only when there is more than 1 block? Or will that still be run with 1 block?

And then I think this would also break the test case? If the block size isn't calculated on load but rather on compaction, I don't believe the test does any compaction.

mknapphrt:
Just to clarify, because I'm not sure I'm on the same page. Is the proposal to sum the size of each block going into populate block and write that as the new block size? If so I'm a little confused because isn't there some level compacting that goes on? Wouldn't the resulting block be somewhat smaller than the blocks going into it?

krasi-georgiev:
I will open a PR against your branch so you can see the idea and let me know if you see any problems.

krasi-georgiev:
@mknapphrt 	opened a PR against your branch, not sure if it pops up in your notifications.

mknapphrt:
I just found it this morning haha, it hadn't popped up this time. I left a comment but I'll just throw it in here too. For some reason we have:
```
var singleBlockSize int64
for _, m := range blocks {	
       singleBlockSize = createPopulatedBlock(t, db.Dir(), 100, m.MinTime, m.MaxTime).Size()
}
```
but I can't figure out why it's in a loop.

krasi-georgiev:
unfortunately this will need rebasing and resolving the conflicted files.

mknapphrt:
So I had to change a few things with the test cases, when using `createPopulatedBlock`, not all the blocks were the same size. So, when calculating the "MaxLimit" it wasn't quite accurate. I summed the the total db size in that loop now and stored the "oldest" block size for setting the max byte limit. I subtracted that size from the total size to set the limit.

krasi-georgiev:
that is strange when I run the tests locally all blocks were of the same size.

mknapphrt:
Yeah, I have no idea why. I looked through it for a while and couldn't figure out a reason for it. It was only a small difference, ~100 bytes or so, but it was enough to break the test case.

krasi-georgiev:
ok ping me when ready for a final review.

mknapphrt:
I'm not sure what I'm doing wrong but I cannot get these merge conflicts to go away while also squashing all the commits. If you want to look over it @krasi-georgiev  while try and figure it out feel free.

mknapphrt:
@krasi-georgiev ping in case my last message was too vague.

krasi-georgiev:
just opened another PR against your branch and once merged will start some real tests in Prometheus.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## master / unreleased
 
+ - [FEATURE]  Size base retention through `Options.MaxBytes`. As part of this change added new metrics - `prometheus_tsdb_storage_blocks_bytes_total`, `prometheus_tsdb_size_retentions_total`, `prometheus_tsdb_time_retentions_total` and new ppublic interface `SizeReader: Size() int64`
```
codesome:
```suggestion
 - [FEATURE]  Size base retention through `Options.MaxBytes`. As part of this change added new metrics - `prometheus_tsdb_storage_blocks_bytes_total`, `prometheus_tsdb_size_retentions_total`, `prometheus_tsdb_time_retentions_total` and new public interface `SizeReader: Size() int64`
```

block.go
```diff
@@ -140,6 +142,12 @@ type Appendable interface {
 	Appender() Appender
 }
 
+// SizeReader returns the size of the object in bytes.
+type SizeReader interface {
+	//Size returns the size in bytes.
```
codesome:
```suggestion
	// Size returns the size in bytes.
```

block.go
```diff
@@ -288,6 +309,17 @@ func OpenBlock(dir string, pool chunkenc.Pool) (*Block, error) {
 	return pb, nil
 }
 
+func blockSize(rr ...SizeReader) int64 {
+	var total int64
+	for _, r := range rr {
+		if r != nil {
+			t := r.Size()
+			total += t
```
codesome: We can directly do it as `total += r.Size()` instead of having an extra variable declaration.

chunks/chunks.go
```diff
@@ -345,6 +345,15 @@ func (s *Reader) Close() error {
 	return closeAll(s.cs...)
 }
 
+// Size returns the size of the chunks.
+func (s *Reader) Size() int64 {
```
codesome: Instead of calculating this everytime we call `Size()`, it is possible to calculate only once and store it in `Reader`? I believe this will optimise calculation of `prometheus_tsdb_storage_blocks_bytes_total` metric.

krasi-georgiev: yes, good suggestion, thanks, I have refactored it.

krasi-georgiev:
@codesome 	thanks, I made the changes with your suggestions.

Now one last look from @gouthamve 	and we are ready to go.

zhanglijingisme:
Sorry for disturbing here...
Is there any possibility to get approximate date for this feature? 
It could be a great help if we can know when & in which Prometheus version this could be available. 
Sincerelly thanks.

krasi-georgiev:
probably 2.7 which should be in about a month.

krasi-georgiev:
can't figure out why the test is failing under windows so I opened an issues and it might be an actual bug in golang
https://github.com/golang/go/issues/29424

db.go
```diff
@@ -489,110 +481,191 @@ func (db *DB) reload() (err error) {
...
+	corrupted = make(map[ulid.ULID]error)
+
+	dirs, err := blockDirs(db.dir)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "find blocks")
+	}
+
 	for _, dir := range dirs {
-		meta, err := readMetaFile(dir)
+		ulid, err := ulid.Parse(filepath.Base(dir))
```
codesome: I think it would be better to parse from meta file as before. Any reason for this change?

db.go
```diff
@@ -489,110 +481,191 @@ func (db *DB) reload() (err error) {
...
+		return nil
+	}
+
+	maxt := loadable[len(loadable)-1].Meta().MaxTime
+
+	return errors.Wrap(db.head.Truncate(maxt), "head truncate failed")
+}
+
+func (db *DB) openBlocks() (blocks []*Block, corrupted map[ulid.ULID]error, err error) {
+	corrupted = make(map[ulid.ULID]error)
```
codesome: nit: you can have this after you check the error for `dirs, err := blockDirs(db.dir)` (avoiding unnecessary allocation).

block.go
```diff
@@ -272,11 +284,20 @@ func OpenBlock(dir string, pool chunkenc.Pool) (*Block, error) {
 		return nil, err
 	}
 
-	tr, err := readTombstones(dir)
+	tr, tsr, err := readTombstones(dir)
 	if err != nil {
 		return nil, err
 	}
 
+	// TODO refactor to set this at block creation time as
```
codesome: Will this be added in this PR?

db.go
krasi-georgiev: I can't remember, I am fine either way.

block.go
krasi-georgiev: yes, I couldn't find a good way to calculate the block size at its creation so would like to revisit this at some point.

db.go
codesome: I see ulid being accessed from meta file, so to keep it consistent I would say from meta file.

codesome:
> I am fine either way. The current logic is that both can be enabled.

Okay I get it now. If both of them are not set, the data will stay forever. Am I correct?

krasi-georgiev:
finally all tests are passing even for windows.

krasi-georgiev:
> Okay I get it now. If both of them are not set, the data will stay forever. Am I correct?

yeah and all the different combinations.

brian-brazil:
> yeah and all the different combinations.

Hmm, we should probably log a warning at startup if storage is going to be infinite.

krasi-georgiev:
yep not a bad idea.
@mknapphrt could you add a warn log.

mknapphrt:
A warning should be added in the prometheus code, right? Done at the time of parsing the flags or on creation of the tsdb I would think. Or do you think it would fit better in the tsdb?

krasi-georgiev:
aah yeah definitively in Prometheus.

I don't want to delay this one any longer, but it would be great if @gouthamve 	or @fabxc have a final look , but if not will wait few more days and will merge.

mknapphrt:
Ok, are we still going to be doing any benchmarking or at this point should I take out the tsdb changes in the prometheus PR?

krasi-georgiev:
Lets merge and will move to Prometheus next.

db.go
```diff
@@ -485,110 +477,191 @@ func (db *DB) reload() (err error) {
...
-			ulid, err2 := ulid.Parse(filepath.Base(dir))
-			if err2 != nil {
-				level.Error(db.logger).Log("msg", "not a block dir", "dir", dir)
-				continue
-			}
-			corrupted[ulid] = err
-			continue
+	for ulid, err := range corrupted {
+		if _, ok := deletable[ulid]; !ok {
+			return errors.Wrap(err, "unexpected corrupted block")
```
gouthamve: This will always trigger an error no?

`deletable := db.deletableBlocks(loadable)` means the `deletable` list will only have blocks from the `loadable` list which don't have any corrupted blocks?

db.go
```diff
@@ -485,110 +477,191 @@ func (db *DB) reload() (err error) {
...
+			block, err = OpenBlock(db.logger, dir, db.chunkPool)
 			if err != nil {
-				return errors.Wrapf(err, "open block %s", dir)
+				corrupted[ulid] = err
 			}
+
 		}
-		blocks = append(blocks, b)
-		opened[meta.ULID] = struct{}{}
+		blocks = append(blocks, block)
```
gouthamve: We append a nil block on encountering a corruption, not sure that is wanted.

db.go
```diff
@@ -485,110 +477,191 @@ func (db *DB) reload() (err error) {
...
+		for _, b := range block.Meta().Compaction.Parents {
+			deletable[b.ULID] = nil
+		}
+	}
+
+	// Sort the blocks by time - newest to oldest (largest to smallest timestamp).
+	// This ensures that the retentions will remove the oldest  blocks.
 	sort.Slice(blocks, func(i, j int) bool {
-		return blocks[i].Meta().MinTime < blocks[j].Meta().MinTime
+		return blocks[i].Meta().MaxTime > blocks[j].Meta().MaxTime
```
gouthamve: Why was this changed? Curious if it changes anything....

db.go
gouthamve: This further is bad because we never print the causes for what corrupted that block leaving the user dumbfounded.

db.go
```diff
@@ -485,110 +477,191 @@ func (db *DB) reload() (err error) {
...
+}
+
+// deleteBlocks closes and deletes blocks from the disk.
+// When the map contains a non nil block object it means it is loaded in memory
+// so needs to be closed first as it might need to wait for pending readers to complete.
+func (db *DB) deleteBlocks(blocks map[ulid.ULID]*Block) error {
+	for ulid, block := range blocks {
+		if block != nil {
+			if err := block.Close(); err != nil {
+				level.Warn(db.logger).Log("msg", "closing block failed", "err", err)
```
gouthamve: Should we return here?

block.go
```diff
@@ -257,7 +266,10 @@ type Block struct {
 
 // OpenBlock opens the block in the directory. It can be passed a chunk pool, which is used
 // to instantiate chunk structs.
-func OpenBlock(dir string, pool chunkenc.Pool) (*Block, error) {
+func OpenBlock(logger log.Logger, dir string, pool chunkenc.Pool) (*Block, error) {
```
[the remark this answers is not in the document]
krasi-georgiev: :+1:

db.go
krasi-georgiev: ok, sounds reasonable

db.go
krasi-georgiev: yeah could have an extra check here

```
if block!=nil{
    blocks = append(blocks, block)
}
```

although in `reload()` it will exit if it encounters a corrupted block that is not scheduled for deletion.

```
for ulid, err := range corrupted {
		if _, ok := deletable[ulid]; !ok {
			return errors.Wrap(err, "unexpected corrupted block")
		}
	}
```

db.go
krasi-georgiev: I really can't remember but reverting it fails the TimeRetention test

db.go
krasi-georgiev: This was the default behaviour before the refactoring. Not sure why.
Maybe it will at least allow deleting it under linux even when closing fails.

db.go
gouthamve: Yes, but if you see, the blocks passed to `deletableBlocks` will have nil blocks and this will panic: https://github.com/prometheus/tsdb/pull/343/files#diff-7a117cbfd5485ed35efcbb2995c37aa0R585

A `continue` right after https://github.com/prometheus/tsdb/pull/343/files#diff-7a117cbfd5485ed35efcbb2995c37aa0R563 will fix it

gouthamve:
One final thing, I still don't see the handling of corrupted blocks is right:


This block expects that `deletable` has the possibility of having `corrupted` blocks: https://github.com/prometheus/tsdb/blob/2933dbca539b963f55d78021702e52bfeba428bf/db.go#L487-L492

But from here:
https://github.com/prometheus/tsdb/blob/2933dbca539b963f55d78021702e52bfeba428bf/db.go#L481-L485

The `loadable` and `corrupted` are mutually exclusive lists, and we pass `loadable` to the `deletableBlocks` function in `L485`. Which means that `deletable` won't have any blocks from `corrupted`.

I think the best way to handle `corrupted` is to fail early, rather than expect them to be deleted.

krasi-georgiev:
@mknapphrt thanks for the great work and sorry for hijacking the last few commits. We just wanted to have this ready for the 2.7 Prometheus release.

mknapphrt:
@krasi-georgiev No worries! I'm happy to see it got finished up, i've been following along quietly but haven't had the time to do much in the last few days and you definitely seemed to have a handle on it. Is there an expected release date for 2.7? I'd be happy to work on the Prometheus side of this, I don't think it should take quite as long as this change haha

codesome:
@mknapphrt with #374 and #501 getting merged, you may now go ahead and open a PR for this in Prometheus.

gouthamve:
I'm actually in the process of doing a PR based on https://github.com/prometheus/prometheus/pull/4230

Sorry for hijacking it but there is some time pressure to do Prometheus 2.7 today.

mknapphrt:
Thanks for all the help guys! Glad to see this finally got somewhere. I'd be happy to help in the future as I'm assuming the WAL size is a goal down the road.

krasi-georgiev:
sure will ping you if we need to get to that. Thanks for your help as well.

PR #344, opened 2018-06-08
chunks: fix potential "index out of range" error

simonpasquier:
When `binary.Uvarint()` fails, the returned number of bytes is less than or equal to zero. I'm not sure how it can happen except with corrupted data but an "index out of range" panic has been reported in https://github.com/prometheus/prometheus/issues/4128.


krasi-georgiev:
well done, thanks Simon.

LGTM

brian-brazil:
Can you add a unittest for this?

simonpasquier:
@brian-brazil unit tests added.

brian-brazil:
:+1:

krasi-georgiev:
@gouthamve PTAL.

PR #348, opened 2018-06-13
Make sure blocks don't overlap to avoid outsider chunks

### LINKED ISSUE #347: Blocks can contain chunks that go past MaxTime
I've been investigating improbable-eng/thanos#354, and I think I've finally identified what's going on.

The issue came up because Thanos checks that all the chunks in a block are between `block.MinTime` and `block.MaxTime`, but in some circumstances, it detected chunks with `block.MaxTime < chunk.MaxTime`. Those chunks also happened to satisfy `block.MaxTime == chunk.MinTime`, which we'll see is no coincidence. After suspecting that the issue came from Thanos, I'm now pretty sure it's in tsdb.

New chunks are create when appending data to head using [append()](https://github.com/prometheus/tsdb/blob/master/head.go#L1205-L1243), specifically this part:

```go
if t >= s.nextAt {
	c = s.cut(t)
	chunkCreated = true
}
```

where `t` is the timestamp of the data being added, and `s.nextAt` can be equal to `block.MaxTime`. So if Prometheus happens to scrape a metric whose timestamp is equal to `block.MaxTime`, it will create a chunk with `chunk.MinTime == block.MaxTime`.

Now, when head needs to be persisted to disk in a new block, [this function](https://github.com/prometheus/tsdb/blob/master/head.go#L722-L753) determines which chunks are part of the block, specifically:

```go
// Do not expose chunks that are outside of the specified range.
if c == nil || !intervalOverlap(mint, maxt, h.mint, h.maxt) {
	return nil, ErrNotFound
}
```

with

```go
func intervalOverlap(amin, amax, bmin, bmax int64) bool {
	// Checks Overlap: http://stackoverflow.com/questions/3269434/
	return amin <= bmax && bmin <= amax
}
```

So a chunk is included in the block if it overlaps it in any way, including the case where `chunk.MinTime == block.MaxTime`.

It isn't ideal, because it means that most data in that last chunk is useless (it won't be read because it's outside the range `[block.MinTime, block.MaxTime]`), and the chunk is going to be part of the next block by the same criteria (it overlaps it as well), so we end up storing the same data twice.

It seems to me that the whole issue stems from the fact that by the way `intervalOverlap()` is defined, blocks overlap by 1 millisecond. For instance, with 2-hour blocks, if `blocks[0].MinTime = 0`, `blocks[0].MaxTime = 7200000`, then the next block will be `blocks[1].MinTime = 7200000`, `blocks[1].MaxTime = 14400000`. So a metric with `t = 7200000` would be in `blocks[0]` and `blocks[1]` at the same time.

A possible way to fix this would be to say that `block.MaxTime` is an exclusive upper bound (so a chunk would only be included if `chunk.MinTime < block.MaxTime && block.MinTime <= chunk.MaxTime`). The other option would be to make sure that blocks don't overlap by the current definition of `intervalOverlap()`, i.e. make sure that `blocks[1].MinTime = blocks[0].MaxTime + 1`.

I've taken that second approach in [this proof of concept](https://github.com/BenoitKnecht/tsdb/commit/23b26e777afa13407ccf09efc318769f5c9fc7e3).

To make reproducing the issue and testing the fix easier, I patched Prometheus to round timestamps down to the minute (so that they would fall on the block boundary): https://github.com/BenoitKnecht/prometheus/tree/fix-block-boundaries. You can use it with

```console
$ git clone -b fix-block-boundaries https://github.com/BenoitKnecht/prometheus.git
$ cd prometheus
$ make build
$ ./prometheus --log.level=debug --storage.tsdb.min-block-duration=10m
```

with `prometheus.yml` containing:

```yaml
scrape_configs:
- job_name: self
  metrics_path: /metrics
  scrape_interval: 60s
  static_configs:
  - targets:
    - localhost:9090
```

BenoitKnecht:
Fixes #347

### DIFF
--- a/block.go
+++ b/block.go
@@ -447,7 +447,7 @@ Outer:
 		}
 
 		for _, chk := range chks {
-			if intervalOverlap(mint, maxt, chk.MinTime, chk.MaxTime) {
+			if chk.OverlapsClosedInterval(mint, maxt) {
 				// Delete only until the current values and not beyond.
 				tmin, tmax := clampInterval(mint, maxt, chks[0].MinTime, chks[len(chks)-1].MaxTime)
 				stones[p.At()] = Intervals{{tmin, tmax}}
@@ -537,6 +537,13 @@ func (pb *Block) Snapshot(dir string) error {
 	}
 
 	return nil
+}
+
+// Returns true if the block overlaps [mint, maxt].
+func (pb *Block) OverlapsClosedInterval(mint, maxt int64) bool {
+	// The block itself is a half-open interval
+	// [pb.meta.MinTime, pb.meta.MaxTime).
+	return pb.meta.MinTime <= maxt && mint < pb.meta.MaxTime
 }
 
 func clampInterval(a, b, mint, maxt int64) (int64, int64) {
--- a/db.go
+++ b/db.go
@@ -360,7 +360,13 @@ func (db *DB) compact() (changes bool, err error) {
 		head := &rangeHead{
 			head: db.head,
 			mint: mint,
-			maxt: maxt,
+			// We remove 1 millisecond from maxt because block
+			// intervals are half-open: [b.MinTime, b.MaxTime). But
+			// chunk intervals are closed: [c.MinTime, c.MaxTime];
+			// so in order to make sure that overlaps are evaluated
+			// consistently, we explicitly remove the last value
+			// from the block interval here.
+			maxt: maxt - 1,
 		}
 		if _, err = db.compactor.Write(db.dir, head, mint, maxt, nil); err != nil {
 			return changes, errors.Wrap(err, "persist head block")
--- a/db.go
+++ b/db.go
@@ -762,8 +762,7 @@ func (db *DB) Querier(mint, maxt int64) (Querier, error) {
 	defer db.mtx.RUnlock()
 
 	for _, b := range db.blocks {
-		m := b.Meta()
-		if intervalOverlap(mint, maxt, m.MinTime, m.MaxTime) {
+		if b.OverlapsClosedInterval(mint, maxt) {
 			blocks = append(blocks, b)
 		}
 	}
@@ -805,8 +804,7 @@ func (db *DB) Delete(mint, maxt int64, ms ...labels.Matcher) error {
 	defer db.mtx.RUnlock()
 
 	for _, b := range db.blocks {
-		m := b.Meta()
-		if intervalOverlap(mint, maxt, m.MinTime, m.MaxTime) {
+		if b.OverlapsClosedInterval(mint, maxt) {
 			g.Go(func(b *Block) func() error {
 				return func() error { return b.Delete(mint, maxt, ms...) }
 			}(b))
@@ -852,11 +850,6 @@ func (db *DB) CleanTombstones() (err error) {
 		}
 	}
 	return errors.Wrap(db.reload(), "reload blocks")
-}
-
-func intervalOverlap(amin, amax, bmin, bmax int64) bool {
-	// Checks Overlap: http://stackoverflow.com/questions/3269434/
-	return amin <= bmax && bmin <= amax
 }
 
 func isBlockDir(fi os.FileInfo) bool {
--- a/db_test.go
+++ b/db_test.go
@@ -26,6 +26,8 @@ import (
 
 	"github.com/oklog/ulid"
 	"github.com/pkg/errors"
+	"github.com/prometheus/tsdb/chunks"
+	"github.com/prometheus/tsdb/index"
 	"github.com/prometheus/tsdb/labels"
 	"github.com/prometheus/tsdb/testutil"
 )
@@ -1095,4 +1097,59 @@ func TestOverlappingBlocksDetectsAllOverlaps(t *testing.T) {
 		{Min: 5, Max: 6}: {nc1[5], nc1[7]},                                 // 2-6, 5-7
 		{Min: 8, Max: 9}: {nc1[8], nc1[9]},                                 // 7-10, 8-9
 	}, OverlappingBlocks(nc1))
+}
+
+// Regression test for https://github.com/prometheus/tsdb/issues/347
+func TestChunkAtBlockBoundary(t *testing.T) {
+	db, close := openTestDB(t, nil)
+	defer close()
+	defer db.Close()
+
+	app := db.Appender()
+
+	blockRange := DefaultOptions.BlockRanges[0]
+	label := labels.FromStrings("foo", "bar")
+
+	for i := int64(0); i < 3; i++ {
+		_, err := app.Add(label, i*blockRange, 0)
+		testutil.Ok(t, err)
+		_, err = app.Add(label, i*blockRange+1000, 0)
+		testutil.Ok(t, err)
+	}
+
+	err := app.Commit()
+	testutil.Ok(t, err)
+
+	_, err = db.compact()
+	testutil.Ok(t, err)
+
+	for _, block := range db.blocks {
+		r, err := block.Index()
+		testutil.Ok(t, err)
+		defer r.Close()
+
+		meta := block.Meta()
+
+		p, err := r.Postings(index.AllPostingsKey())
+		testutil.Ok(t, err)
+
+		var (
+			lset labels.Labels
+			chks []chunks.Meta
+		)
+
+		chunkCount := 0
+
+		for p.Next() {
+			err = r.Series(p.At(), &lset, &chks)
+			testutil.Ok(t, err)
+			for _, c := range chks {
+				testutil.Assert(t, meta.MinTime <= c.MinTime && c.MaxTime <= meta.MaxTime,
+					"chunk spans beyond block boundaries: [block.MinTime=%d, block.MaxTime=%d]; [chunk.MinTime=%d, chunk.MaxTime=%d]",
+					meta.MinTime, meta.MaxTime, c.MinTime, c.MaxTime)
+				chunkCount++
+			}
+		}
+		testutil.Assert(t, chunkCount == 1, "expected 1 chunk in block %s, got %d", meta.ULID, chunkCount)
+	}
 }
--- a/db_test.go
+++ b/db_test.go
@@ -1153,3 +1153,35 @@ func TestChunkAtBlockBoundary(t *testing.T) {
 		testutil.Assert(t, chunkCount == 1, "expected 1 chunk in block %s, got %d", meta.ULID, chunkCount)
 	}
 }
+
+func TestQuerierWithBoundaryChunks(t *testing.T) {
+	db, close := openTestDB(t, nil)
+	defer close()
+	defer db.Close()
+
+	app := db.Appender()
+
+	blockRange := DefaultOptions.BlockRanges[0]
+	label := labels.FromStrings("foo", "bar")
+
+	for i := int64(0); i < 5; i++ {
+		_, err := app.Add(label, i*blockRange, 0)
+		testutil.Ok(t, err)
+	}
+
+	err := app.Commit()
+	testutil.Ok(t, err)
+
+	_, err = db.compact()
+	testutil.Ok(t, err)
+
+	testutil.Assert(t, len(db.blocks) >= 3, "invalid test, less than three blocks in DB")
+
+	q, err := db.Querier(blockRange, 2*blockRange)
+	testutil.Ok(t, err)
+	defer q.Close()
+
+	// The requested interval covers 2 blocks, so the querier should contain 2 blocks.
+	count := len(q.(*querier).blocks)
+	testutil.Assert(t, count == 2, "expected 2 blocks in querier, got %d", count)
+}
--- a/chunks/chunks.go
+++ b/chunks/chunks.go
@@ -57,6 +57,12 @@ func (cm *Meta) writeHash(h hash.Hash) error {
 	return nil
 }
 
+// Returns true if the chunk overlaps [mint, maxt].
+func (cm *Meta) OverlapsClosedInterval(mint, maxt int64) bool {
+	// The chunk itself is a closed interval [cm.MinTime, cm.MaxTime].
+	return cm.MinTime <= maxt && mint <= cm.MaxTime
+}
+
 var (
 	errInvalidSize     = fmt.Errorf("invalid size")
 	errInvalidFlag     = fmt.Errorf("invalid flag")
--- a/compact.go
+++ b/compact.go
@@ -592,7 +592,7 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta, 
 		if len(dranges) > 0 {
 			// Re-encode the chunk to not have deleted values.
 			for i, chk := range chks {
-				if !intervalOverlap(dranges[0].Mint, dranges[len(dranges)-1].Maxt, chk.MinTime, chk.MaxTime) {
+				if !chk.OverlapsClosedInterval(dranges[0].Mint, dranges[len(dranges)-1].Maxt) {
 					continue
 				}
 
--- a/head.go
+++ b/head.go
@@ -735,19 +735,14 @@ func (h *headChunkReader) Chunk(ref uint64) (chunkenc.Chunk, error) {
 	s.Lock()
 	c := s.chunk(int(cid))
 
-	// This means that the chunk has been garbage collected.
-	if c == nil {
+	// This means that the chunk has been garbage collected or is outside
+	// the specified range.
+	if c == nil || !c.OverlapsClosedInterval(h.mint, h.maxt) {
 		s.Unlock()
 		return nil, ErrNotFound
 	}
-
-	mint, maxt := c.minTime, c.maxTime
 	s.Unlock()
 
-	// Do not expose chunks that are outside of the specified range.
-	if c == nil || !intervalOverlap(mint, maxt, h.mint, h.maxt) {
-		return nil, ErrNotFound
-	}
 	return &safeChunk{
 		Chunk: c.chunk,
 		s:     s,
@@ -852,7 +847,7 @@ func (h *headIndexReader) Series(ref uint64, lbls *labels.Labels, chks *[]chunks
 
 	for i, c := range s.chunks {
 		// Do not expose chunks that are outside of the specified range.
-		if !intervalOverlap(c.minTime, c.maxTime, h.mint, h.maxt) {
+		if !c.OverlapsClosedInterval(h.mint, h.maxt) {
 			continue
 		}
 		*chks = append(*chks, chunks.Meta{
@@ -1289,6 +1284,11 @@ func (s *memSeries) head() *memChunk {
 type memChunk struct {
 	chunk            chunkenc.Chunk
 	minTime, maxTime int64
+}
+
+// Returns true if the chunk overlaps [mint, maxt].
+func (mc *memChunk) OverlapsClosedInterval(mint, maxt int64) bool {
+	return mc.minTime <= maxt && mint <= mc.maxTime
 }
 
 type memSafeIterator struct {


db.go
```diff
@@ -550,7 +550,7 @@ func (db *DB) reload(deleteable ...string) (err error) {
 	if len(blocks) == 0 {
 		return nil
 	}
-	maxt := blocks[len(blocks)-1].Meta().MaxTime
+	maxt := blocks[len(blocks)-1].Meta().MaxTime + 1
```
Bplotka: is that really needed?

Bplotka: Ok, it is, because if block A is `0-10` and block B is `10-20`, sample t=10 must be in both.

However, this is quite risky change, affecting lots of things. Wonder if we can abort this and try to make compactor deduplicating these chunks correctly. Looking into that.

Bplotka: Ok, talked offline with @fabxc and in fact we can kill this change and leave just adjusting the nextAt change.

This is because:
> if a block has `minTime: 10` and `maxTime: 20` the intend was for this to be interval `[10,20)`

db.go
```diff
@@ -807,7 +807,7 @@ func (db *DB) Querier(mint, maxt int64) (Querier, error) {
 
 func rangeForTimestamp(t int64, width int64) (mint, maxt int64) {
 	mint = (t / width) * width
-	return mint, mint + width
+	return mint, mint + width - 1
```
Bplotka: let's add comment that we explicitly don't want chunk minT sample to be in maxTime block if it is the case because block is `[minT, maxT)`

db_test.go
```diff
@@ -1098,3 +1100,51 @@ func TestOverlappingBlocksDetectsAllOverlaps(t *testing.T) {
 		{Min: 8, Max: 9}: {nc1[8], nc1[9]},                                 // 7-10, 8-9
 	}, OverlappingBlocks(nc1))
 }
+
+func TestChunkAtBlockBoundary(t *testing.T) {
```
Bplotka: Please add comment that this is regression test for issue https://github.com/prometheus/tsdb/issues/347

Bplotka:
> That's confusing, and it means that intervalOverlap() should behave differently if comparing chunk intervals or block intervals. 

Unfortunately that is the case now. There is small intentional inconsistency here that we need to address in this PR IMO.

```
               // Wrap head into a range that bounds all reads to it.
		head := &rangeHead{
			head: db.head,
			mint: mint,
			// Block range is always [minT, maxt), so we need to explicitly exclude maxt samples.
			maxt: maxt-1,
		}
```
..is one way of doing this.

Bplotka:
> So I can try to rework this PR to interpret block intervals as [a, b), but chunk intervals would still be interpreted as [a, b], which will most likely cause confusion down the line. Do you still think it's a better approach?

From what I got from @fabxc this is the best approach I think.

db.go
```diff
@@ -381,7 +381,13 @@ func (db *DB) compact() (changes bool, err error) {
 		head := &rangeHead{
 			head: db.head,
 			mint: mint,
-			maxt: maxt,
+			// We remove 1 millisecond from maxt because block
+			// intervals are half-open: [b.MinTime, b.MaxTime). But
+			// chunk intervals are closed ([c.MinTime, c.MaxTime]),
```
Bplotka: `([c.MinTime, c.MaxTime])` -> `[c.MinTime, c.MaxTime]` 

Let's remove brackets to not confuse here (:

db_test.go
```diff
@@ -1098,3 +1100,52 @@ func TestOverlappingBlocksDetectsAllOverlaps(t *testing.T) {
...
+
+		var (
+			lset labels.Labels
+			chks []chunks.Meta
+		)
+
+		for p.Next() {
+			err = r.Series(p.At(), &lset, &chks)
+			testutil.Ok(t, err)
+			for _, c := range chks {
```
Bplotka: can we at least count chunks here? We should have 2 blocks (?) one chunk in each I think. That is the only thing I wanted to add in my extension. (wanted to assert on exact chunk time range, but just counting these would be nice as well)

Bplotka:
@fabxc @gouthamve @krasi-georgiev can we have review on this? It is rdy to go in my opinion (despite my minor comments)

db_test.go
BenoitKnecht: I think we can assert the number of chunks per block, yes.

Not sure about checking the number of blocks, because it depends on this condition in `db.compact()`:

```go
if db.head.MaxTime()-db.head.MinTime() <= db.opts.BlockRanges[0]/2*3 {
        break
}
```

which could change and break the test for no reason. (In fact, we only have one block here because of this condition.)

db.go
BenoitKnecht: Good point, done.

BenoitKnecht:
I noticed that we also need to change the following in order to be consistent:

```diff
diff --git a/db.go b/db.go
index bd3fdf4..6ddbbd3 100644
--- a/db.go
+++ b/db.go
@@ -785,7 +785,7 @@ func (db *DB) Querier(mint, maxt int64) (Querier, error) {
 
        for _, b := range db.blocks {
                m := b.Meta()
-               if intervalOverlap(mint, maxt, m.MinTime, m.MaxTime) {
+               if intervalOverlap(mint, maxt, m.MinTime, m.MaxTime-1) {
                        blocks = append(blocks, b)
                }
        }
@@ -828,7 +828,7 @@ func (db *DB) Delete(mint, maxt int64, ms ...labels.Matcher) error {
 
        for _, b := range db.blocks {
                m := b.Meta()
-               if intervalOverlap(mint, maxt, m.MinTime, m.MaxTime) {
+               if intervalOverlap(mint, maxt, m.MinTime, m.MaxTime-1) {
                        g.Go(func(b *Block) func() error {
                                return func() error { return b.Delete(mint, maxt, ms...) }
                        }(b))
```

Rather than just add this as a commit, I'm wondering if it wouldn't be clearer to replace 1afe3dbd6bb8718f79abc5826fc9128f3065d8bb with

```diff
diff --git a/head.go b/head.go
index 2c7c7ec..ecfca36 100644
--- a/head.go
+++ b/head.go
@@ -742,7 +742,7 @@ func (h *headChunkReader) Chunk(ref uint64) (chunkenc.Chunk, error) {
        s.Unlock()
 
        // Do not expose chunks that are outside of the specified range.
-       if c == nil || !intervalOverlap(mint, maxt, h.mint, h.maxt) {
+       if c == nil || !intervalOverlap(mint, maxt, h.mint, h.maxt-1) {
                return nil, ErrNotFound
        }
        return &safeChunk{
@@ -849,7 +849,7 @@ func (h *headIndexReader) Series(ref uint64, lbls *labels.Labels, chks *[]chunks
 
        for i, c := range s.chunks {
                // Do not expose chunks that are outside of the specified range.
-               if !intervalOverlap(c.minTime, c.maxTime, h.mint, h.maxt) {
+               if !intervalOverlap(c.minTime, c.maxTime, h.mint, h.maxt-1) {
                        continue
                }
                *chks = append(*chks, chunks.Meta{
```

It feels more systematic, because don't treat `rangeHead` and regular blocks differently. Everytime `intervalOverlap()` compares a block (or head) range with any other range, we remove `1` from the block's `MaxTime` to make the interval half-open.

What do you think?

Bplotka:
Is `intervalOverlap` really used only against chunk? This change would be really hard to figure out if it is not  breaking anything else.

db_test.go
Bplotka: I think it is a priority to test if actually, next block will include the block we ignore with 
```
head := &rangeHead{
 			head: db.head,	 			head: db.head,
 			mint: mint,	 			mint: mint,
			// We remove 1 millisecond from maxt because block
			// intervals are half-open: [b.MinTime, b.MaxTime). But
			// chunk intervals are closed: [c.MinTime, c.MaxTime];
			// so in order to make sure that overlaps are evaluated
			// consistently, we explicitly remove the last value
			// from the block interval here.
			maxt: maxt - 1,
 		}	 		}
```
Can we somehow make it test resilient on change here?
```
if db.head.MaxTime()-db.head.MinTime() <= db.opts.BlockRanges[0]/2*3 {
        break
}
```

Bplotka:
talked offline, can agree that changing how we use `internalOverlap` for chunks might be necessary. This PR initial changeset indeed forgets about `headIndexReader`

BenoitKnecht:
OK, unless someone else objects, I'll update this PR with the modified `intervalOverlap()` arguments.

Bplotka:
The only arguments towards `rangeHead` is single hacky point in single place that is responsible for writing block, not two somewhere buried deep down in overall chunk retrieval. But it's up to you .

krasi-georgiev:
I am still digging through the code to understand why block time range doesn't match the chunk time range.

it would be great if @fabxc or @gouthamve can give us some hints.

BenoitKnecht:
Alright, talking with @Bplotka on Slack, I can get behind the idea of changing `rangeHead` and not treating head intervals like block intervals when we call `intervalOverlap()`.

But we do need to pass `m.MaxTime-1` to `intervalOverlap()` in `db.Querier` and `db.Delete`. The idea being, if your blocks are `[a, b)`, `[b, c)`, `[c, d)` and you make a query for the interval `[b, c]`, the current implementation will return a querier over all three blocks; but you only need `[b, c)` and `[c, d)`, because `[a, b)` doesn't contain `b`.

BenoitKnecht:
(But I can submit that fix in a different PR, if it simplifies review.)

Bplotka:
> why block time range doesn't match the chunk time range.
yea it is big unknown.

Yea, agreed. These 3 places (`rangeHead`,  `db.Delete`, `db.Querier`) are where we are wrongly assuming `[t1, t2]` range for block time ranges whereas block time range is always `[t1, t2)`.

I think changing `intervalOverlap` inside `Series` and `Chunk` so basically:
> treating head intervals like block intervals when we call intervalOverlap().
is not safe, because the TSDB code is definitely not prepared for this (might more tricky to find further bugs)

Bplotka:
No, this PR is fine for it I think

BenoitKnecht:
> No, this PR is fine for it I think

OK, I'll push the fix and a regression test.

db_test.go
BenoitKnecht: Fair point, I'll look into it.

gouthamve:
I don't remember if anything in particular made us choose this. I think it should aligned with what prometheus expects when a query comes in for `a to b`.

gouthamve:
Prometheus expects it to be `[t0, t1]` and I think for the sake of consistency, let us keep it inclusive everywhere.

I am not averse to: `maxt: maxt - 1,` when creating the persisted block. i.e, making the maxt of a block look like `xxxx999`

Bplotka:
yea, @krasi-georgiev, I got clear message also that this inconsistency is only for *esthetic reasons* (no `[100000-299999]`). 

This means that in theory, we could change block range to be inclusive `[t1 t2]` just fine, but it won't be a trivial change... For example, we really need backward compatibility here, so we need to make compactor being able to handle `[ )` as well as `[ ]` blocks..

BenoitKnecht:
@gouthamve I also favor this solution, I agree it's more consistent.

@Bplotka I think we can handle backwards compatibility in the compactor by checking if the blocks being compacted overlap (when interpreted as `[a, b]` blocks), and if so dropping any chunk that starts at `b`. We may also want to check the last block for an outsider chunk, but this is where things become ambiguous, and the outsider chunk won't hurt until we compact the block further.

db.go
```diff
@@ -381,7 +381,13 @@ func (db *DB) compact() (changes bool, err error) {
...
-			maxt: maxt,
+			// We remove 1 millisecond from maxt because block
+			// intervals are half-open: [b.MinTime, b.MaxTime). But
+			// chunk intervals are closed: [c.MinTime, c.MaxTime];
+			// so in order to make sure that overlaps are evaluated
+			// consistently, we explicitly remove the last value
+			// from the block interval here.
+			maxt: maxt - 1,
 		}
 		if _, err = db.compactor.Write(db.dir, head, mint, maxt); err != nil {
```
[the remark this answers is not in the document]
BenoitKnecht: If we want blocks to be closed intervals *and* not overlap, the correct fix is to drop everything else in this PR and do this instead:

```diff
diff --git a/db.go b/db.go
index 28cb14f..ec6f64f 100644
--- a/db.go
+++ b/db.go
@@ -550,7 +550,7 @@ func (db *DB) reload(deleteable ...string) (err error) {
        if len(blocks) == 0 {
                return nil
        }
-       maxt := blocks[len(blocks)-1].Meta().MaxTime
+       maxt := blocks[len(blocks)-1].Meta().MaxTime + 1
 
        return errors.Wrap(db.head.Truncate(maxt), "head truncate failed")
 }
@@ -807,7 +807,7 @@ func (db *DB) Querier(mint, maxt int64) (Querier, error) {
 
 func rangeForTimestamp(t int64, width int64) (mint, maxt int64) {
        mint = (t / width) * width
-       return mint, mint + width
+       return mint, mint + width - 1
 }
 
 // Delete implements deletion of metrics. It only has atomicity guarantees on a per-block basis.
diff --git a/head.go b/head.go
index 2c7c7ec..7926cdf 100644
--- a/head.go
+++ b/head.go
@@ -1133,7 +1133,10 @@ func (s *memSeries) cut(mint int64) *memChunk {
 
        // Set upper bound on when the next chunk must be started. An earlier timestamp
        // may be chosen dynamically at a later point.
-       _, s.nextAt = rangeForTimestamp(mint, s.chunkRange)
+       _, maxt := rangeForTimestamp(mint, s.chunkRange)
+       // rangeForTimestamp() returns the maximum timestamp for the previous
+       // chunk; we want to cut right after that, so we add 1 millisecond
+       s.nextAt = maxt + 1
 
        app, err := c.chunk.Appender()
        if err != nil {
```

(At least that's what I had originally implemented and tested.)

I'm not updating this PR right away because I'm not sure there's a consensus on whether we want `[a, b]` or `[a, b)` blocks yet. @fabxc @krasi-georgiev What's your take on this?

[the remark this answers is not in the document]
gouthamve: Yep, while I think `[a, b]` makes sense, not sure what the others think.

I'm really sorry to have you go back and forth on the changes @BenoitKnecht :( Thanks for keeping up, really appreciate it!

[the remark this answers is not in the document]
BenoitKnecht: No worries :) It's not an insignificant change, it's worth making sure we make the right decision.

Bplotka:
> @Bplotka I think we can handle backwards compatibility in the compactor by checking if the blocks being compacted overlap (when interpreted as [a, b] blocks), and if so dropping any chunk that starts at b. We may also want to check the last block for an outsider chunk, but this is where things become ambiguous, and the outsider chunk won't hurt until we compact the block further.

Yes, but I don't think by having just SINGLE block, you can tell if it's `[a,b]` or `[a,b)`. ): That's super important issue, because you can have compaction with missing block in between (so for example `[t-100, t-50]` and `[t-20, t]` and you have no idea if `[t-100, t-50]` is actually `[t-100, t-50)`. That's why I think that our workaround should be merged first (to clearly here chunks `[ ]` and blocks `[)` logic everywhere), and the block being `[ ]` should be done carefully and slowly in different PR. 

But can see that you want to do it immediately the huge change?

BenoitKnecht:
@Bplotka I don't think it's really an issue though. Blocks that have been created so far are `[a, b]` blocks, not `[a, b)`. They were intended to be `[a, b)` blocks, but that's not how they were written on disk, and it's not how they were defined in code either. So for all intents and purposes, blocks created so far are `[a, b]` blocks.

The issue with those blocks only manifests itself when they're contiguous and compacted together, because if there was a chunk starting at `b` and you compact `[a, b]` with `[b, c]`, you end up with that chunk being duplicated.

But taken individually, or if they don't overlap (such as in the non-contiguous case you mentioned), it's perfectly correct to treat them as `[a, b]` blocks; they do contain data from that range, including `b`.

So I don't think that the proposal to keep treating blocks as `[a, b]` intervals (but making sure that they don't overlap by writing them as `[a, b-1]` from now on) is a bigger change than starting to treat all blocks as `[a, b)` intervals.

krasi-georgiev:
I am in big favour of making blocks and chunk ranges consistent. 
Since this doesn't require an urgent fix lets spend enough time to make it right.

fabxc:
Thanks for all the debugging work you've put into this.

Looking at the current fix, we account for that in `rangeForTimestamp` to make it return the closed interval for a given timestamp by doing `-1` on the max time – makes sense. That will affect the actual interval written to `meta.json`.

What will happen if I now have a TSDB that has blocks written the old way and this new way?
It seems we would have to account for that difference.
Generally the change would require a bump of the `meta.json` format to `3` (`2` has been used in an erroneous format bump before and must be skipped). Then we could set `block.MaxTime--` when loading a block with meta version 1 and we should be good.

Please be aware this change makes debugging harder since the new maximum timestamps are hard to parse for humas. Tooling can help here I'm sure.

fabxc:
Discussion on whether half open or closed interval are better for this particular case, I do want to point out that this change is disruptive (esp. with the meta version change) and may cause other issues where we've forgotten the `-1` change to be considered.

From what I understand currently chunks boundaries are in the right place. The only problem being that when compacting from the head, we consider one chunk too much. Truncation works correctly as well since the chunk won't be dropped from the head and will be found in the next compacted block as well.

The only error would thus be that we consider one chunk too much. Would it potentially be safer to do nothing but change this [single line](https://github.com/prometheus/tsdb/blob/40766622eeedfcd6287feaa73c9ba8c72d60f1c9/db.go#L384) to `maxt-1`?

https://github.com/prometheus/tsdb/blob/40766622eeedfcd6287feaa73c9ba8c72d60f1c9/db.go#L380-L385

It's just a thought I had, please triple check me on it and let me know whether that's viable.

One thing worth noting is that block ranges and chunk ranges are semantically different things. The chunk range describes the lowest and highest timestamp of actually existing samples.
The block range specifies an abstract range in which samples may exist. It doesn't imply `mint` and `maxt` have a real sample in any case.

gouthamve:
> The block range specifies an abstract range in which samples may exist.

Which essentially means block range is both included right: `[  ]`?

fabxc:
No, that's up to how you define it and right now our definition is `[ )`. And not changing that currently appears much safer from my point of view. Regardless of how one may judge the initial decision.

Bplotka:
Yea, I need to agree with Fabian here. 

We were already on the same page that some change like this (some comments above):
https://github.com/prometheus/tsdb/pull/348#issuecomment-396922274
(and same thing as @fabxc proposed)

...will fix this issue entirely. However, we tried to understand what is the point of inconsistency and if it is worth to make it consistent. 

After the research, it is clear in my opinion, that there is a huge risk and effort in changing this inconsistency (making blocks range exclusive `[ ]`) and minimal (if any) actual benefit of having consistent ranges. 

Or am I missing these benefits? Can we enumerate those?

krasi-georgiev:
> One thing worth noting is that block ranges and chunk ranges are semantically different things. The chunk range describes the lowest and highest timestamp of actually existing samples.
> The block range specifies an abstract range in which samples may exist. It doesn't imply mint and maxt have a real sample in any case.

if the block range specifies an abstract range in which samples may exist having an overlap even by a 1 microsecond is still an overlap, no?
Making it consistent should make it easier to understand and troubleshoot bugs.

> Please be aware this change makes debugging harder since the new maximum timestamps are hard to parse for humas. Tooling can help here I'm sure.

do you mean the 999 at the end would make it more difficult to read or something else? I always use some online converted even with the current timestamps so the 999s would make it more difficult for me.

Bplotka:
> Making it consistent should make it easier to understand and troubleshoot bugs.

Maybe a little bit. But changing this will introduce thousands of other issues and debug problems (mix of "legacy blocks" and "new blocks").

krasi-georgiev:
> Maybe a little bit. But changing this will introduce thousands of other issues and debug problems (mix of "legacy blocks" and "new blocks").

Isn't this why we have the Meta file versions? Start Prom , read the Meta versions and convert where needed , just few lines of code, no?

fabxc:
Yes, we have versions so that we can make changes where necessary. Doesn't mean we should do it whenever possible.

> Start Prom , read the Meta versions and convert where needed , just few lines of code, no?

The concern is exactly that things aren't that easy in practice.

BenoitKnecht:
All right, so it looks like the majority leans towards considering blocks as `[ )` intervals, as it's less likely to introduce new bugs and should fix the initial issue. Thank you all for your thoughtful comments!

I'm going to push that implementation later today.

BenoitKnecht:
I've pushed the new code. I tried to make the distinction between block intervals and chunk intervals as explicit as possible by replacing the generic `intervalOverlap()` function with `OverlapsClosedInterval()` methods on the block and chunk objects.

Let me know what you think.

block.go
```diff
@@ -539,6 +539,13 @@ func (pb *Block) Snapshot(dir string) error {
 	return nil
 }
 
+// Returns true if the block overlaps [mint, maxt].
```
Bplotka: Bit confused. You have 2 method name `OverlapsClosedInterval` and comment `Returns true if the block overlaps [mint, maxt].`  but with different logic.

The inner comments and logic make sense though. I think we could leave it as methods and name it just `isOverlapping` (or `...halfOpen..` for block and `...closed...` for chunk) and move inner comments to method comment.

Bplotka: These methods are good place to explain the difference in time ranges between block and chunks maybe?

BenoitKnecht: And you're right, I should expand on the difference between block and chunk intervals in those method descriptions.

Bplotka: Wait, but one is Closed, second is Half-Closed interval, right?

Bplotka: Ok, this "closed" is related to arguments... hm this is super confusing (for me), I would just leave `OverlapsInterval` for both chunk and block in this case.

Bplotka: or even `Overlaps` .. basically we want to know if `mint` and `maxt` overlaps with block or chunk. And it gives us right answer (e.g: if `mint` is touching `block.maxt` is does NOT overlap)

brian-brazil:
What do we need to do to get this in?

Bplotka:
@brian-brazil what do you mean? From my perspective, the current code (except discussion around method naming) is what we need to fix https://github.com/prometheus/tsdb/issues/347

brian-brazil:
This code review appears to be ongoing, so I'm wondering what changes (if any) the maintainers are looking for.

block.go
BenoitKnecht: I'm not 100% convinced it's clearer to just call it `Overlaps`. I wanted to highlight the fact the `mint` and `maxt` arguments refer to a closed interval to avoid someone using it to compare a block with another block, in which case `[mint, maxt)` would be a half-open interval, whereas the method expects it to be closed.

That being said, if it's the only thing keeping this PR from being merged, I'm happy to make the change.

Bplotka: 1. We need someone from maintainer to approve this PR anyway, so maybe someone can put his view on this as well, I don't want to block this PR either (: 
2. If I do `block1.Overlaps(block2.min, block2.max)` is there anything confusing by the fact that it will check overlaps using `[a, b)` because blocks are like this? I think having something like `block1.OverlapsClosedInterval(block2.min, block2.max)` would be confusing for future readers of this code, because it is saying `closed` but actually it is half open overlap logic. But I might be wrong here!

BenoitKnecht: But `block1.OverlapsClosedInterval(block2.min, block2.max)` would return the wrong result, because `[block2.min, block2.max)` isn't a closed interval. You would get `block1.min <= block2.max && block2.min < block1.max` instead of the correct value for two `[a, b)` intervals, which is `block1.min < block2.max && block2.min < block1.max`.

That's why I wanted the method name to explicitly state that it requires a closed interval as its argument; it doesn't work on other types of intervals.

Bplotka: Got you, agree (:

PR #351, opened 2018-06-26
Limit the returned db.Querier to the requested time range

free:
Limit the returned `db.Querier` to the requested time range. Preallocate the `baseChunkSeries.lset` and `baseChunkSeries.chks` slices to the previous series' slice sizes to avoid unnecessary grow slice reallocations.

Tested for correctness and performance against prometheus/prometheus head. [Before](https://github.com/prometheus/tsdb/files/2136119/before.txt) and [after](https://github.com/prometheus/tsdb/files/2136124/after.txt) [`bench_test`](https://github.com/prometheus/prometheus/blob/dd6781add2390eb2f2251e108f2ef835eba44b05/promql/bench_test.go) results ([comparison](https://github.com/prometheus/tsdb/files/2136145/benchcmp.txt)) show reductions of up to 40% in execution times and up to 90% in memory allocations in range queries for plain series retrievals.

These are somewhat inflated numbers, as the test uses a single in-memory block spanning 26.4 hours with timeseries at 10 second resolution, but non-negligible on my Prometheus setup (5 second scrape, 10 second eval), as one can clearly spot a sawtooth pattern with a 2 hour period in both memory and, more tellingly, CPU utilization:

rangehead

### DIFF
--- a/querier.go
+++ b/querier.go
@@ -499,17 +499,10 @@ func (s *baseChunkSeries) Err() error { return s.err }
 
 func (s *baseChunkSeries) Next() bool {
 	var (
-		lset     labels.Labels
-		chkMetas []chunks.Meta
+		lset     = make(labels.Labels, len(s.lset))
+		chkMetas = make([]chunks.Meta, len(s.chks))
 		err      error
 	)
-
-	if lsetLen := len(s.lset); lsetLen > 0 {
-		lset = make(labels.Labels, lsetLen)
-	}
-	if chksLen := len(s.chks); chksLen > 0 {
-		chkMetas = make([]chunks.Meta, chksLen)
-	}
 
 	for s.p.Next() {
 		ref := s.p.At()
--- a/querier.go
+++ b/querier.go
@@ -498,9 +498,16 @@ func (s *baseChunkSeries) At() (labels.Labels, []chunks.Meta, Intervals) {
 func (s *baseChunkSeries) Err() error { return s.err }
 
 func (s *baseChunkSeries) Next() bool {
+	lsetLen := len(s.lset)
+	chksLen := len(s.chks)
+	if lsetLen == 0 {
+		lsetLen = 10
+		chksLen = 10
+	}
+
 	var (
-		lset     labels.Labels
-		chkMetas []chunks.Meta
+		lset     = make(labels.Labels, lsetLen)
+		chkMetas = make([]chunks.Meta, chksLen)
 		err      error
 	)
 
--- a/querier.go
+++ b/querier.go
@@ -498,18 +498,18 @@ func (s *baseChunkSeries) At() (labels.Labels, []chunks.Meta, Intervals) {
 func (s *baseChunkSeries) Err() error { return s.err }
 
 func (s *baseChunkSeries) Next() bool {
-	lsetLen := len(s.lset)
-	chksLen := len(s.chks)
-	if lsetLen == 0 {
-		lsetLen = 10
-		chksLen = 10
-	}
-
 	var (
-		lset     = make(labels.Labels, lsetLen)
-		chkMetas = make([]chunks.Meta, chksLen)
+		lset     labels.Labels
+		chkMetas []chunks.Meta
 		err      error
 	)
+
+	if lsetLen := len(s.lset); lsetLen > 0 {
+		lset = make(labels.Labels, lsetLen)
+	}
+	if chksLen := len(s.chks); chksLen > 0 {
+		chkMetas = make([]chunks.Meta, chksLen)
+	}
 
 	for s.p.Next() {
 		ref := s.p.At()
--- a/db.go
+++ b/db.go
@@ -784,7 +784,11 @@ func (db *DB) Querier(mint, maxt int64) (Querier, error) {
 		}
 	}
 	if maxt >= db.head.MinTime() {
-		blocks = append(blocks, db.head)
+		blocks = append(blocks, &rangeHead{
+			head: db.head,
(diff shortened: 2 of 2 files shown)


free:
Oh, and the one +66.13% memory usage increase for `expr=a_hundred_-_b_hundred` is due to the fact that my initial (speculative) slice size is 10 (for both labels and chunks), whereas the golang runtime will go from `nil` -> 2 -> 4 -> 8 etc. (doubling the size afterwards) and this particular query produces only 2 labels and 1 chunk, which would otherwise only require going from a `nil` slice to a slice of size 2, and thus no grow slice reallocations.

brian-brazil:
That's a nice improvement. I'm a bit suspicious that only one benchmark slows down due to this, and am wondering if one change is compensating for the other. What happens if you benchmark the changes independently?

fabxc:
Those are two independent optimizations so it would be good to get before/after benchmarks for both independently.

Do I understand correctly that one benchmark uses a head that spans a range of a size that won't happen in real world setups? If so, benchmarking against impossible inputs seems to skew the picture as well.

free:
I did benchmark the `querier.go` change independently too, at first, but primarily looked at the times, not the memory. Will try to drop the initial slice size of 10 and benchmark without that.

There are actually 3 optimizations in this PR:
 1. Only return chunks that may contain relevant samples, not all chunks in head. This doesn't have much of a downside in any situation, apart from the indirection of going though `rangeHead`.
 2. Pre-allocate the label set and chunks slices when iterating over the series returned by a query, to the same sizes as the slices of previously returned series. This has a chance of slightly increasing memory usage if e.g. querying something like `{job="prometheus"}`, which will return series with varying numbers of labels and/or chunks (something which is not covered by the benchmark), but in the average case queries will have a metric name and so will return series with the same number of labels and (apart from instance churn) the same number of chunks.
 3. Pre-allocate the label set and chunks slices (of size 10) for the first series returned by a query. This is arguably overkill, particularly when querying for instant values (only one chunk) or series with few labels and will over-allocate compared to the "initialize with `nil` and keep appending" approach. It will always be at least as fast, though.

>  I'm a bit suspicious that only one benchmark slows down due to this, and am wondering if one change is compensating for the other.

It doesn't slow down, benchmarks are all faster (or the same) across the board. There is one benchmark that uses considerably more memory, though.

> Do I understand correctly that one benchmark uses a head that spans a range of a size that won't happen in real world setups? If so, benchmarking against impossible inputs seems to skew the picture as well.

Correct. But you can get pretty close to that setup by scraping at 1 second intervals. (The test creates a head that spans 86400 + 100000 seconds, so about 50 hours, at 10 second resolution. That's only 2.5x as large as a real-life head of 2 hours at 1 second resolution.)

free:
I redid the benchmarks (on a different machine, so not directly comparable to the earlier benchmarks, [new baseline](https://github.com/prometheus/tsdb/files/2138818/before.txt)), first [only with change (1)](https://github.com/prometheus/tsdb/files/2138826/after1.txt) above, then with [(1)+(2)](https://github.com/prometheus/tsdb/files/2138828/after2.txt), then with [(1)+(2)+(3)](https://github.com/prometheus/tsdb/files/2138830/after3.txt) and finally with [(1)+(2)+(3')](https://github.com/prometheus/tsdb/files/2138834/after3b.txt) (where 3' is 3, with 6 labels and 2 chunks initially, instead of 10 and 10).

As it turns out, while (3) alone did produce an improvement without (1) and (2), it is basically a wash in terms of time and increases memory usage in a handful of cases when applied together with them. (1)+(2)+(3') is almost identical to (1)+(2), but that may be different in the real world (where series usually have more than 2 or 3 labels, which is what the benchmark uses).

(1) is an improvement across the board, while (1)+(2) seems equivalent in terms of time but it reduces memory usage for large queries (20%+ for either 1000+ steps or `1d` ranges) while marginally increasing it (1-2%) for small queries.

Oh, and as far as I can tell the 66% increase in memory utilization for just the one benchmark was a fluke, as I cannot see any increase in memory utilization above 1% between the new baseline and the (1)+(2)+(3) variant.

brian-brazil:
Looking at the numbers, 1 is a pretty clear win. 2 also seems okay. The rest are unclear.

free:
I've dropped (3). It's now only the clear wins.

krasi-georgiev:
@free started a benchmark for these changes if you want to follow the progress.
https://github.com/prometheus/prometheus/pull/4763

querier.go
```diff
@@ -504,6 +504,13 @@ func (s *baseChunkSeries) Next() bool {
 		err      error
 	)
 
+	if lsetLen := len(s.lset); lsetLen > 0 {
```
krasi-georgiev: why not just

```
var (
		lset     = make(labels.Labels, len(s.lset))
		chkMetas = make([]chunks.Meta, len(s.chks))
		err      error
	)
```

free: I thought a `make(..., 0)` would allocate memory, since it produces a non-nil slice: https://play.golang.org/p/9ISeSKFbJaM

But [apparently](https://stackoverflow.com/questions/44305170/nil-slices-vs-non-nil-slices-vs-empty-slices-in-go-language) it will simply create a slice with a `Data` pointer referencing some fixed address, so there's no memory penalty in making a zero-sized slice.

Replaced with your suggestion.

krasi-georgiev: yeah I also did some benchmarking and it showed the same results.

krasi-georgiev:
LGTM

@gouthamve

PR #354, opened 2018-06-27
Add resilience to crashes during deletion

fabxc:
@brian-brazil would this work for what we want to achieve?
If yes, I'll check whether there are any new cases the existing tests don't cover and add them.

### DIFF
--- a/block.go
+++ b/block.go
@@ -178,6 +178,8 @@ type BlockMetaCompaction struct {
 	Level int `json:"level"`
 	// ULIDs of all source head blocks that went into the block.
 	Sources []ulid.ULID `json:"sources,omitempty"`
+	// Short descriptions of the direct blocks that were used to create
+	// this block.
 	Parents []BlockDesc `json:"parents,omitempty"`
 	Failed  bool        `json:"failed,omitempty"`
 }
--- a/block.go
+++ b/block.go
@@ -164,6 +164,13 @@ type BlockStats struct {
 	NumTombstones uint64 `json:"numTombstones,omitempty"`
 }
 
+// BlockDesc describes a block by ULID and time range.
+type BlockDesc struct {
+	ULID    ulid.ULID `json:"ulid"`
+	MinTime int64     `json:"minTime"`
+	MaxTime int64     `json:"maxTime"`
+}
+
 // BlockMetaCompaction holds information about compactions a block went through.
 type BlockMetaCompaction struct {
 	// Maximum number of compaction cycles any source block has
@@ -171,6 +178,7 @@ type BlockMetaCompaction struct {
 	Level int `json:"level"`
 	// ULIDs of all source head blocks that went into the block.
 	Sources []ulid.ULID `json:"sources,omitempty"`
+	Parents []BlockDesc `json:"parents,omitempty"`
 	Failed  bool        `json:"failed,omitempty"`
 }
 
@@ -475,19 +483,17 @@ func (pb *Block) CleanTombstones(dest string, c Compactor) (*ulid.ULID, error) {
 
 	pb.tombstones.Iter(func(id uint64, ivs Intervals) error {
 		numStones += len(ivs)
-
 		return nil
 	})
-
 	if numStones == 0 {
 		return nil, nil
 	}
 
-	uid, err := c.Write(dest, pb, pb.meta.MinTime, pb.meta.MaxTime)
+	meta := pb.Meta()
+	uid, err := c.Write(dest, pb, pb.meta.MinTime, pb.meta.MaxTime, &meta)
 	if err != nil {
 		return nil, err
 	}
-
 	return &uid, nil
 }
 
--- a/db.go
+++ b/db.go
@@ -267,17 +267,9 @@ func (db *DB) run() {
 		case <-db.compactc:
 			db.metrics.compactionsTriggered.Inc()
 
-			_, err1 := db.retentionCutoff()
-			if err1 != nil {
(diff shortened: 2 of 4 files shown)


block.go
```diff
@@ -164,13 +164,21 @@ type BlockStats struct {
...
+}
+
 // BlockMetaCompaction holds information about compactions a block went through.
 type BlockMetaCompaction struct {
 	// Maximum number of compaction cycles any source block has
 	// gone through.
 	Level int `json:"level"`
 	// ULIDs of all source head blocks that went into the block.
 	Sources []ulid.ULID `json:"sources,omitempty"`
+	Parents []BlockDesc `json:"parents,omitempty"`
```
brian-brazil: This should have a comment

db.go
```diff
@@ -267,17 +267,9 @@ func (db *DB) run() {
 		case <-db.compactc:
 			db.metrics.compactionsTriggered.Inc()
 
-			_, err1 := db.retentionCutoff()
```
brian-brazil: We're now relying on compact() to call reload() and thus delete old data, which only happens after compaction. We probably need to call it more often.

db.go
```diff
@@ -492,31 +423,50 @@ func (db *DB) reload(deleteable ...string) (err error) {
 	if err != nil {
 		return errors.Wrap(err, "find blocks")
 	}
+	// We delete old blocks that have been superseded by new ones by gathering all parents
```
brian-brazil: The comment should mention how this can happen, so that future readers know this is designed to fix the relevant race condition. A matching comment should go in the delete/rename compaction code pointing to here.

db.go
```diff
@@ -492,31 +423,50 @@ func (db *DB) reload(deleteable ...string) (err error) {
...
 		}
-
+		for _, b := range meta.Compaction.Parents {
+			deleteable[b.ULID] = struct{}{}
+		}
+	}
+	for _, dir := range dirs {
+		meta, err := readMetaFile(dir)
+		if err != nil {
+			level.Warn(db.logger).Log("msg", "read meta information", "err", err, "dir", dir)
```
brian-brazil: This changes us to no longer failing to startup if meta isn't readable. I think we should only do that if the block was deleteable.

db.go
fabxc: I thought about this too. But then I realized that our retention decision is based on the most recent block. So effectively nothing can change about the retention state unless `compact()` did something.

Thus it is sufficient.

brian-brazil: I'm not convinced. Time can move forward, and a user may have (inadvisably) disabled compaction.

fabxc: Certainly worth a doc comment though.

fabxc: Time can move forward but TSDB has nothing to do with wall time at all. Everything is purely based on timestamps of inserted data.
Adding wall time would come with a ton of complications and would make TSDB unsuitable for any data that's not timestamped within a very close range to "real world now".

That was actually my motivation once to have Prometheus continuously write a single series with the current wall timestamp into TSDB in the background. To enforce the coupling.

brian-brazil: That makes sense.

krasi-georgiev:
I was thinking about a different approach:
we write all blocks to a sub-folder first `/data/compaction` -> when compaction is complete delete old blocks and move new block inside /data . 
at startup just delete everything in `/data/compaction` this way OOM-ing during compaction is not dangerous.

fabxc:
We've thought through this approach as well. The problem is that it is still prone to leave your data directory in a bad state. With this particular approach, we would even be losing data.
If Prometheus crashes after the old blocks are deleted but before the new one was moved, it comes back up and deletes `/data/compaction` and with that all data in the compacted range is now lost.

The problem is that what we are trying to do cannot be done atomically.
So to handle crashes gracefully, we must persist state about the parent blocks of a newly compacted one.

db.go
```diff
@@ -492,31 +430,68 @@ func (db *DB) reload(deleteable ...string) (err error) {
...
+		blocks     []*Block
+		corrupted  = map[ulid.ULID]error{}
+		opened     = map[ulid.ULID]struct{}{}
+		deleteable = map[ulid.ULID]struct{}{}
 	)
+	for _, dir := range dirs {
+		meta, err := readMetaFile(dir)
+		if err != nil {
+			// The block was potentially in the middle of being deleted during a crash.
+			// Skip it since we may delete it properly further down again.
```
krasi-georgiev: Can't see where are these being deleted?

fabxc:
@Bplotka this is something you might want to adopt for the Thanos compactor.

db.go
fabxc: We are not deleting all blocks that hit this error. But this block may be set in `deleteable` by another block.

krasi-georgiev: for some reason my local cache was missing the `corrupted` map. it is all clear now thanks.

krasi-georgiev:
LGTM

db.go
```diff
@@ -492,31 +430,67 @@ func (db *DB) reload(deleteable ...string) (err error) {
...
+		}
+		if db.beyondRetention(meta) {
+			deleteable[meta.ULID] = struct{}{}
+			continue
+		}
+		for _, b := range meta.Compaction.Parents {
+			deleteable[b.ULID] = struct{}{}
+		}
+	}
+	// Blocks we failed to open should all be those we are want to delete anyway.
```
krasi-georgiev: // Blocks we failed to open should all be those we ~are~ want to delete anyway.

PR #356, opened 2018-06-28
Log series on rollback

### LINKED ISSUE #355: Series created in an appender that is rolled back never written to WAL
Series creation is a side-effect of an appender, even when rolled back. However we only commit them to the WAL when an appender is commited. Any future references to such a series would fail to be read back at startup.

fabxc:
Fixes #355 

@brian-brazil


brian-brazil:
:+1:

PR #357, opened 2018-07-04
querier shouldn't return samples before the retention time.

krasi-georgiev:
fixes: https://github.com/prometheus/prometheus/issues/4348

@gouthamve , @Bplotka

### DIFF
--- a/db.go
+++ b/db.go
@@ -750,6 +750,16 @@ func (db *DB) Snapshot(dir string, withHead bool) error {
 // Querier returns a new querier over the data partition for the given time range.
 // A goroutine must not handle more than one open Querier.
 func (db *DB) Querier(mint, maxt int64) (Querier, error) {
+	// Substract the retention from the current time and convert it to milliseconds so we can compare it with mint, maxt.
+	retentionCutoff := time.Now().Add(-time.Duration(db.opts.RetentionDuration)*time.Millisecond).UnixNano() / int64(time.Millisecond)
+	// Limit the querer to the retention time.
+	if mint < retentionCutoff {
+		mint = retentionCutoff
+	}
+	if maxt < retentionCutoff {
+		maxt = retentionCutoff
+	}
+
 	var blocks []BlockReader
 
 	db.mtx.RLock()


db.go
```diff
@@ -750,6 +750,16 @@ func (db *DB) Snapshot(dir string, withHead bool) error {
 // Querier returns a new querier over the data partition for the given time range.
 // A goroutine must not handle more than one open Querier.
 func (db *DB) Querier(mint, maxt int64) (Querier, error) {
+	// Substract the retention from the current time and convert it to milliseconds so we can compare it with mint, maxt.
+	retentionCutoff := time.Now().Add(-time.Duration(db.opts.RetentionDuration)*time.Millisecond).UnixNano() / int64(time.Millisecond)
+	// Limit the querer to the retention time.
+	if mint < retentionCutoff {
+		mint = retentionCutoff
+	}
+	if maxt < retentionCutoff {
```
Bplotka: Can we return an error in this case then? or immediately return the empty result? If you ask the data before retention.

juliusv: Querying data that is not present is never an error, so empty result would be the right thing.

db.go
```diff
@@ -750,6 +750,16 @@ func (db *DB) Snapshot(dir string, withHead bool) error {
 // Querier returns a new querier over the data partition for the given time range.
 // A goroutine must not handle more than one open Querier.
 func (db *DB) Querier(mint, maxt int64) (Querier, error) {
+	// Substract the retention from the current time and convert it to milliseconds so we can compare it with mint, maxt.
+	retentionCutoff := time.Now().Add(-time.Duration(db.opts.RetentionDuration)*time.Millisecond).UnixNano() / int64(time.Millisecond)
+	// Limit the querer to the retention time.
```
juliusv: querer -> querier

db.go
```diff
@@ -750,6 +750,16 @@ func (db *DB) Snapshot(dir string, withHead bool) error {
 // Querier returns a new querier over the data partition for the given time range.
 // A goroutine must not handle more than one open Querier.
 func (db *DB) Querier(mint, maxt int64) (Querier, error) {
+	// Substract the retention from the current time and convert it to milliseconds so we can compare it with mint, maxt.
```
juliusv: Substract -> Subtract

db.go
Bplotka: Yea, The issue is that code could potentially still return single sample even when retention is `T-10` and we ask for `T-1000 - T-100`, because we will overwrite mint and maxt with `T-10`

db.go
```diff
@@ -750,6 +750,16 @@ func (db *DB) Snapshot(dir string, withHead bool) error {
...
+	retentionCutoff := time.Now().Add(-time.Duration(db.opts.RetentionDuration)*time.Millisecond).UnixNano() / int64(time.Millisecond)
```
juliusv: With the `UnixNano()` we are leaving the domain of the specific encoding of `time.Duration`, so it'd be better to divide by `1e6` instead of `time.Millisecond` (whose int value depends on the `time.Duration` encoding implementation details).

db.go
juliusv: Yep, good point, so we should just return an empty result here (but not an error).

juliusv: (Well, we have to return a querier, but then it should be one that doesn't produce any data.)

fabxc:
FWIW, I'm still not convinced that this is actually a bug.

juliusv:
@fabxc I don't have a strong opinion on that either. Will let others decide...

juliusv:
@brian-brazil I don't think there would be any negative interaction with `read_recent`... if it's set, we read from the remote source in any case and have less local data to merge in, if it's not set, we already have a [4h overlap buffer](https://github.com/prometheus/prometheus/blob/master/cmd/prometheus/main.go#L545-L546) based on the actual oldest sample in the local TSDB.

```
   actual start time
       |                 |--->... time range assumed to be in local storage
       v                 v
local: |---2h---|---2h---|---2h---|---2h---|--->...
                         ^
                         |
                  adjusted start time (+2x2h)
```

When we drop older blocks faster, it just means that the adjusted start time becomes even younger and we query a bit more from the remote storage.

brian-brazil:
> if it's not set, we already have a 4h overlap buffer based on the actual oldest sample in the local TSDB.

This PR could make part of that buffer inaccessible though.

juliusv:
Ah hurdur, forgot that of course chunks are not dropped quicker now, didn't sleep well tonight.

The computed base start time value could be limited to retention cutoff as well?

brian-brazil:
That sounds about right.

fabxc:
Based on the email thread the reporter also didn't report it because he wants to get a fake-retention back but because he doesn't want a single second more data being hold than defined. That's not what this PR solves either – not like that's something we can reasonably do anyway.

krasi-georgiev:
@fabxc if you think this is the expected behaviour and it doesn't solve the reported use case I agree that we shouldn't add this additional check.

@shirpx could you add a bit more context why do you need 1h retention so it helps us understand whether this PR will fix your use case.

shirpx:
It's not only 1h retention that doesn't work. it's any retention that i set, for example 24h:
https://screenshot.net/e52kksl
https://screenshot.net/07l33hw

and the reason i want lower retention than the default 15 days is that want to use thanos that allows to query older data from a bucket when data is missing locally.

krasi-georgiev:
@shirpx now this is unexpected. I will try to replicate locally and will let you know.

krasi-georgiev:
I confirm that I was able to replicate locally.
The retention doesn't seem to affect the returned metrics so will work on this tomorrow.

juliusv:
It still doesn't seem like we have consensus that this is a needed/wanted feature.

@shirpx I do not understand the Thanos use case. If Prometheus has data locally, that will be merged in, otherwise not, but that doesn't impede additional usage of Thanos's long-term store.

krasi-georgiev:
@juliusv did you see the screenshot - retention 1d, but the graph shows 2d(or more) of data.
the only explanation for this is if the largest block is more than 1d so that it still overlaps with the retention and doesn't get deleted.

@shirpx you can confirm this by running the tsdb scan tool or attaching the meta.json for all blocks to see the time ranges.
tsdb ls -h /path/to/data

the tsdb is attached if you want to use from here or build it from prometheus/tsdb/cmd/tsdb
[tsdb.zip](https://github.com/prometheus/tsdb/files/2178117/tsdb.zip)

juliusv:
@krasi-georgiev Yes, I saw that. But what I am missing is an explanation of why that is a problem (other than using up space, which isn't solved by this PR). I do not see how it is a problem for using Thanos to return more data locally than the configured retention time.

krasi-georgiev:
@juliusv Yeah I don't understand what is the problem there as well.

Still curious to see the block timings just to confirm that there isn't some bug in the retention logic.

eladamitpxi:
for us the reason (partially) is also protecting prometheus from OOM
i.e. I don't mind if a query node fails due to OOM due to a bad query since they are stateless and easily scalable, prometheus on the other hand has no memory consumption protection
so there's that and the storage costs which are a driver for having retention work as expected

juliusv:
@eladamitpxi 

> i.e. I don't mind if a query node fails due to OOM due to a bad query since they are stateless and easily scalable, prometheus on the other hand has no memory consumption protection

Ah. I guess with "query node" you mean Thanos query nodes? So are these OOM problems happening in combination with Thanos where the Thanos querier is asking a Prometheus server for the total query time range and Prometheus has more history than expected and thus it ends up trying to build a huge remote read response to Thanos and OOMs? I guess I can see that being a problem (esp. since remote read protos are not a very memory-friendly representation of sample data), though wondering whether that could/should be solved on the Thanos end. Since Thanos should know what local retention is set to, it could be changed to never request more than that from the local Prometheus server. But honestly I don't have a strong opinion or well-thought out reasoning about this :)

> so there's that and the storage costs which are a driver for having retention work as expected

Note that this PR does not change anything about storage costs because it only changes what data is returned in queries, not when it is actually deleted from disk.

eladamitpxi:
yes, that's why a proper fix of the retention flag would be nicer, but i'll take what i can get for the first issue :)

Bplotka:
Yea @eladamitpxi If it relates to Thanos, it will be easier for all here to know about it, so please mention it next time, so we will have a full understanding of what is needed to be fixed.

The main question is here:
A) Does the query part OOMs the Prometheus? If yes - then we can maybe mitigate that in Thanos (although it's not that trivial, and we can talk it in details here https://github.com/improbable-eng/thanos/issues/82 or on Thanos slack). I agree it is a tricky decision if should we fix it in TSDB/Prometheus or not, because you are asking for this time range and TSDB gives what it has. Should it put some virtual limit just because of retention setting? Not sure. 

B) Do you just ingest too much data, so it fits perfectly within retention but no byte more - that's another story.

Let's maybe jump to https://github.com/improbable-eng/thanos/issues/82 and explore options that will help you @eladamitpxi I think we can solve it using `flags` API and Thanos `Info` endpoint (that propagates minTime)

eladamitpxi:
for the query / OOM bit, yes let's move the discussion
for the retention bit in general, the retention flag should work regardless of this (i.e. the storage bit)

shirpx:
for summary, i want the retention to work since thanos queries both bucket and prometheus for the same data and i want to get it only from the bucket.

Bplotka:
@shirpx lets say we land this and this gives only something newer than retention time. Then you would like to put what retention exactly? I heard that you need 1h is that correct?

Bplotka:
if you have 2h block specified (and you should) and desire to backup these data to S3 or GCS in that case you will have lot's of other issues with 1h retention - for example Thanos will never be able to upload the 2h block. Because your retention is 1h. 

Let me know if I am wrong and you actually want to setup something else.

shirpx:
the 1h retention was an example, no retention works, even when i put 24h

roidelapluie:
I do not think that this is a relevant. We already had the case where we were really happy to be able to query data a bit further than the actual retention time.

Bplotka:
@roidelapluie 
> We already had the case where we were really happy to be able to query data a bit further than the actual retention time.

What do you mean?

> the 1h retention was an example, no retention works

@shirpx It works, just it does not work exactly within the second to the specified retention. It can vary with ~2h in worst case AFAIK because retention is applied on a block level, not on sample level. 

**To sum up:**
 We can artificially be strict in retention for query layer, however:
- there is a risk involved with this change, for example affecting `read_recent` as Brian mentioned.
- there is a high risk that someone already bases on this "bug" (or feature?) and will be surprised.
- technically samples are still there in TSDB so limiting just query can give a fake feeling of data not being stored.

I would actually state clearly in retention flag and documentation that retention is not accurate to the second, just to the ~2h unless that logic is really blocking anyone. From what I get you are not yet blocked @shirpx, we are discussion about theoritcal issues, or am I missing something?

BTW @shirpx your actual desired retention matters from Thanos perspective though. If you want a "sane" retention like 24h (sane from Thanos perspective), we can (and should) focus to make `thanos-query` fetch the most from bucket and rest from sidecar. If that does not work as expected, feel free to create issue in Thanos project (: Plus, 15d retention does not make sense with Thanos, you are welcome to join our slack for more details (TLDR we don't allow local compaction)

roidelapluie:
we have 100d of retentions and I remember that we needed data 102 d ago. As they were stinn on TSDB, we could easily get access to that data. I think we should not artificially limit that ability.

eladamitpxi:
it feels like this thread is starting to lose it's focus (at least in terms of what we feel is important for both thanos and prometheus)
no matter what we set as the retention period in prometheus, it is ignored (be it 1, 3 or 24 hours) that is a bug which should be fixed (if it is indeed a bug and not something in our configuration)
we are discussing the thanos angles on the thanos slack and github issues if someone wants to join :)

juliusv:
@eladamitpxi 

> it feels like this thread is starting to lose it's focus (at least in terms of what we feel is important for both thanos and prometheus)
no matter what we set as the retention period in prometheus, it is ignored

So far my assumption was that it was just not deleted immediately, so you still saw some data past the retention period. Do you even when you set retention to 1h, the data *never* disappears? I.e. can you still see 1d-old data with a 1h retention? I would expect data to disappear / be deleted after a couple of hours if you set a 1h retention.

Bplotka:
Basically, there is a difference between:
A) User set retention to 24h and she/he can still view data from 2 days ago.
B) User set retention to 24h and she/he can still view data from 25h ago but not 26h ago.

First one indicates quite a big bug, second is just maybe not clear flag description. I think it is acceptable and it is how the retention logic is designed to not be strict to the second. It can vary with maximum ~2h. 

I am pretty sure we can observe only B and this is what this issue is all about or am I wrong (should we be strict to the very second, or are we fine with the current logic). Let us know if "A" is actually the case here.

juliusv:
> It can vary with maximum ~2h.

To be clear, blocks can be compacted into sizes of [`cfg.tsdb.Retention / 10`](https://github.com/prometheus/prometheus/blob/master/cmd/prometheus/main.go#L205-L207), so for longer retention times, the compacted blocks can cover much larger time ranges and thus also need more than ~2h to be deleted after some of the data in them goes over the retention time.

There's also another limit placed on the maximum block size by https://github.com/prometheus/prometheus/blob/master/storage/tsdb/tsdb.go#L134, but that seems to be a very large limit of 1640.25 days in the last compaction round if I'm not mistaken. That seems a bit crazy high actually.

juliusv:
(that's all still as expected though, and not a bug)

Bplotka:
Ah right, forgot about local compaction part, good point Julius, too much time spent on Thanos (: 

So now the question is: should retention be more strict or not from Prometheus perspective.
And if it should be strict, how to do it (: Because this PR fixes only querier part, not how much is actually stored.

juliusv:
@Bplotka Yep. And I don't see a clear enough reason to artificially limit queried data yet without actually solving disk space issues.

I can see that in the Thanos+Prometheus combination you don't want to put unnecessary load on Prometheus's local TSDB when Thanos already also has most of the requested range, but that part might be best solved on the Thanos side?

Bplotka:
> I can see that in the Thanos+Prometheus combination you don't want to put unnecessary load on Prometheus's local TSDB when Thanos already also has most of the requested range, but that part might be best solved on the Thanos side?

Yes, added quick win tracking issue for it: https://github.com/improbable-eng/thanos/issues/437

eladamitpxi:
@juliusv i wish :( we set the retention flag in our staging env when we started integrating thanos a couple of weeks ago, and it only now started affecting data retention

eladamitpxi:
> There's also another limit placed on the maximum block size by https://github.com/prometheus/prometheus/blob/master/storage/tsdb/tsdb.go#L134, but that seems to be a very large limit of 1640.25 days in the last compaction round if I'm not mistaken. That seems a bit crazy high actually.
@juliusv  is there a minimum block size for rollover to happen (i.e. starting to think that we just arent hitting that in our staging env which is why it might be compacting to one block and not deleting)?

juliusv:
@eladamitpxi So by default, blocks are compacted in 10 steps of 3x, like this:

2h (initial size) -> 6h -> 18h -> 54h -> 162h -> ...(x3 every time)... -> 1640.25 days

So the longer your retention time, the more really long blocks you'll have (cause you need a lot of history before you can end up with the big blocks), and the longer it'll take for them to get dropped because the entire block has to fall outside of the retention window for it to be dropped. However, Prometheus also limits this block size to 1/10 of the total retention time, so the maximum "overhang" of old data you should have at any given time is 1/10 of the total history. If you really want, you can also limit the maximum block size manually by setting the hidden flag `--storage.tsdb.max-block-duration=4h` or something, but every block is its own TSDB with index and you'll have many blocks to query then for long queries.

PR #363, opened 2018-07-14
Expose prometheus_tsdb_lowest_timestamp metric

bobmshannon:
Closes https://github.com/prometheus/prometheus/issues/2988

### DIFF
--- a/db.go
+++ b/db.go
@@ -164,13 +164,10 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
 	}, func() float64 {
 		db.mtx.RLock()
 		defer db.mtx.RUnlock()
-		startTime := time.Now().Unix()
-		for _, b := range db.Blocks() {
-			if b.meta.MinTime < startTime {
-				startTime = b.meta.MinTime
-			}
+		if len(db.blocks) == 0 {
+			return float64(db.head.minTime)
 		}
-		return float64(startTime)
+		return float64(db.blocks[0].meta.MinTime)
 	})
 	m.tombCleanTimer = prometheus.NewHistogram(prometheus.HistogramOpts{
 		Name: "prometheus_tsdb_tombstone_cleanup_seconds",
--- a/db.go
+++ b/db.go
@@ -123,7 +123,7 @@ type dbMetrics struct {
 	compactionsTriggered prometheus.Counter
 	cutoffs              prometheus.Counter
 	cutoffsFailed        prometheus.Counter
-	startTime            prometheus.Gauge
+	startTime            prometheus.GaugeFunc
 	tombCleanTimer       prometheus.Histogram
 }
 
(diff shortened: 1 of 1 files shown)


krasi-georgiev:
Is the original use case still valid?
@bkupidura did you find a workaround for your use case or you still need this metric?

bkupidura:
Hey, im no longer working on this. I will ping right people to give You answer.

@ildarsv @mescanef @dkalashnik

krasi-georgiev:
Thanks
If the workaround works well we can close it and revisit if needed.

btw what was the workaround?

dkalashnik:
@krasi-georgiev small proxy on top of several Prometheuses with data merge.

simonpasquier:
Hello @dkalashnik @bkupidura :smile: 
@gouthamve mentioned that reporting the oldest timestamp in the db is a nice-to-have and I kind of agree, especially if it's cheap and easy.

db.go
```diff
@@ -157,6 +158,20 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
 		Name: "prometheus_tsdb_retention_cutoffs_failures_total",
 		Help: "Number of times the database failed to cut off block data from disk.",
 	})
+	m.startTime = prometheus.NewGaugeFunc(prometheus.GaugeOpts{
+		Name: "prometheus_tsdb_start_time_seconds",
+		Help: "Oldest timestamp stored in the database.",
+	}, func() float64 {
+		db.mtx.RLock()
+		defer db.mtx.RUnlock()
+		startTime := time.Now().Unix()
```
gouthamve: There is no reason to believe that the time stored is going to be the current time. For example, it could `10, 20, 30` or even time in the future. It's best initialise `startTime` as `math.MaxInt64`, and if there is no data in the block, set it to `0`.

Further, the blocks are sorted by time. So doing just `db.blocks[0].MinTime` should suffice. And if `len(db.blocks) == 0`, then we should look at `HeadBlock` mintime.

db.go
```diff
@@ -157,6 +158,17 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
 		Name: "prometheus_tsdb_retention_cutoffs_failures_total",
 		Help: "Number of times the database failed to cut off block data from disk.",
 	})
+	m.startTime = prometheus.NewGaugeFunc(prometheus.GaugeOpts{
+		Name: "prometheus_tsdb_start_time_seconds",
```
gouthamve: `prometheus_tsdb_lowest_timestamp` makes more sense here, as start_time could be confused with the process start time and it may or may not be seconds :)

simonpasquier: I think we should keep the unit suffix and enforce the metric's value to be in seconds though.

gouthamve: No. From a general TSDB perspective, there is no reason that the time passed into `.Append(t int64, v float64)`, is a wall time. It could be anything, and people may even pass nano-seconds.

simonpasquier: Hmm. Does it make sense to have an "opaque" metric then? Wouldn't it be better to expose the raw value to the caller (eg Prometheus) and let it compute the metric?

bobmshannon: If Prometheus appends millisecond timestamps to the TSDB, then this metric as collected might be slightly more difficult to work with using something like the built-in `time()` function for example. For that reason I agree that it would be nice to adhere to metric naming best practices and include the base unit as a suffix in the metric name, converting the millisecond timestamp to a unix timestamp if necessary. As @simonpasquier mentioned I would think that instrumenting this metric directly in Prometheus instead would allow for this.

With that being said perhaps there are other consumers of the TSDB where exposing this lowest timestamp metric directly would still be useful.

simonpasquier:
:+1:

krasi-georgiev:
@simonpasquier , @gouthamve so finally are we adding this?

gouthamve:
👍

PR #364, opened 2018-07-23
Reword chunk references, LSB usually means 'bits'.

cpatulea:
https://en.wikipedia.org/wiki/Bit_numbering#Least_significant_bit


cpatulea:
Added DCO, PTAL.

krasi-georgiev:
LGTM 

@gouthamve PTAL

PR #365, opened 2018-08-05
Handle a bunch of unchecked errors

juliusv:
As discovered by "gosec".

Signed-off-by: Julius Volz <julius.volz@gmail.com>

### DIFF
--- a/block.go
+++ b/block.go
@@ -483,10 +483,13 @@ Outer:
 func (pb *Block) CleanTombstones(dest string, c Compactor) (*ulid.ULID, error) {
 	numStones := 0
 
-	pb.tombstones.Iter(func(id uint64, ivs Intervals) error {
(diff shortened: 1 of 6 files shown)


juliusv:
There's also these two Sync/Close calls in the segment truncation code where I'm not sure if at least one of them should be checked too? (there are a bunch of unchecked `Close()` calls and the like in general, but mostly we cannot do anything when they fail)

juliusv:
Meant to include link in the previous comment: https://github.com/prometheus/tsdb/blob/master/wal.go#L403-L404

krasi-georgiev:
ping @juliusv

juliusv:
@krasi-georgiev Currently traveling, but I think it should be fine to merge as is, and adding additional error logs for `Sync()` and `Close()` calls can still be done separately if people feel it's a good idea.

krasi-georgiev:
Thanks!

PR #366, opened 2018-08-07
Fixes for 32bit archs

fabxc:
@brian-brazil this addresses the build failure for 32bit and moves one `int` to `int64` for safety. 
Still have to debug the prometheus/prometheus tests getting stuck.


brian-brazil:
:+1:

Issue #279, opened 2018-02-08
Add a method for getting all label names


shubheksha:
This has to be implemented in the Querier, like https://github.com/prometheus/tsdb/blob/master/querier.go#L34-L35.  
It's needed to implement https://github.com/prometheus/prometheus/issues/2038.

Gouthamve:
We have the info in `reader.labels` map. We can just check for the separator and if there is none, add it to the list.

sipian:
Is there anyone working on this issue? If not I would like to take this up.

PR #297, opened 2018-03-13
add function to get all label names in db

sipian:
Fixes #279 

This PR fixes for the test cases that were failing in https://github.com/prometheus/tsdb/pull/296

Issue #26, opened 2017-03-29
Label Values composite index


gouthamve:
I see from [this](https://docs.google.com/document/d/13uK6HoEXzX67DuWsJ2PKo8zjpqlgpQ0SIHz2nvMa7f0/edit#heading=h.zzd5egnpdp5) doc that we are going to have composite indexes. But it is still un-implemented. 

@fabxc I am assuming this is a TODO item and you have not dropped the idea of a composite index.

fabxc:
Yes, this is a to-do that should be possible to add seemlessly at a later
point. The label value index interface shows support for it already.

It is useful for advanced autocompletion in Prometheus. For example, if we
want to suggest values for a certain label, it makes sense to only suggest
those that exist for the given metric name (__name__ label). The composite
index provides the required lookup.

codesome:
@krasi-georgiev I was about to implement this in my GSoC. But as there were not many use cases with this, it was decided that we don't implement it for now.

krasi-georgiev:
yeah that is what I wanted to know. 
Autocompletion sounds very useful so really surprising why it didn't get more interest.

gouthamve:
Yes. Also @codesome has a branch somewhere that does this, I think (:

PR #369, opened 2018-08-28
LabelNames() method to get all unique label names

codesome:
This is required to address https://github.com/prometheus/prometheus/issues/2038 and fixes https://github.com/prometheus/tsdb/issues/279.

This PR would supersede https://github.com/prometheus/tsdb/pull/297.

Currently the unit test hangs at `db.Close()`, not sure why, help appreciated.


block.go
```diff
@@ -84,6 +84,9 @@ type IndexReader interface {
 	// LabelIndices returns a list of string tuples for which a label value index exists.
 	LabelIndices() ([][]string, error)
 
+	// LabelNames returns all the unique label names present in the index.
+	LabelNames() []string
```
gouthamve: Should return an `error`

db.go
```diff
@@ -852,6 +852,39 @@ func (db *DB) CleanTombstones() (err error) {
...
+
+	for _, name := range headIndexr.LabelNames() {
+		labelNamesMap[name] = struct{}{}
+	}
+
+	db.mtx.RLock()
+	for _, b := range db.blocks {
+		blockIndexr, err := b.Index()
+		if err != nil {
+			level.Error(db.logger).Log("msg", "block index reader fail", "err", err)
```
gouthamve: Specially here, we need to return this error rather than log it.

head.go
```diff
@@ -959,6 +959,18 @@ func (h *headIndexReader) LabelValues(names ...string) (index.StringTuples, erro
 	return index.NewStringTuples(sl, len(names))
 }
 
+// LabelNames returns all the unique label names present in the head.
+func (h *headIndexReader) LabelNames() []string {
+	var labelNames []string
+	for name := range h.head.values {
```
gouthamve: Hmm, are you sure there are no races here?

db.go
```diff
@@ -852,6 +852,39 @@ func (db *DB) CleanTombstones() (err error) {
 	return errors.Wrap(db.reload(), "reload blocks")
 }
 
+// LabelNames returns all the unique label names present in the DB in sorted order.
+func (db *DB) LabelNames() []string {
+	headIndexr, _ := db.head.Index()
```
gouthamve: `headIndexr.Close()` after using it.

db.go
```diff
@@ -852,6 +852,39 @@ func (db *DB) CleanTombstones() (err error) {
...
+	labelNamesMap := make(map[string]struct{})
+
+	for _, name := range headIndexr.LabelNames() {
+		labelNamesMap[name] = struct{}{}
+	}
+
+	db.mtx.RLock()
+	for _, b := range db.blocks {
+		blockIndexr, err := b.Index()
```
gouthamve: `blockIndexr.Close()` after using it.

db.go
```diff
@@ -852,6 +852,53 @@ func (db *DB) CleanTombstones() (err error) {
...
+func (db *DB) LabelNames() (labelNames []string, err error) {
+	headIndexr, _ := db.head.Index()
+	defer headIndexr.Close()
+	labelNamesMap := make(map[string]struct{})
+
+	names, err := headIndexr.LabelNames()
+	if err != nil {
+		return nil, err
+	}
+	for _, name := range names {
```
gouthamve: Could you close `headIndexr` here itself, as it's not being used again. Further, it returns an error which should be returned.

Also, could you wrap the errors with appropriate messages like: `errors.Wrap(err, "close headIndexReader")`

db.go
```diff
@@ -852,6 +852,53 @@ func (db *DB) CleanTombstones() (err error) {
...
+	}
+
+	db.mtx.RLock()
+	for _, b := range db.blocks {
+		blockIndexr, e := b.Index()
+		if e != nil {
+			err = e
+			break
+		}
+		defer blockIndexr.Close()
```
gouthamve: Same as `headIndexr`.

db_test.go
```diff
@@ -1301,3 +1301,80 @@ func TestInitializeHeadTimestamp(t *testing.T) {
...
+			for _, tuple := range tst.sampleLabels {
+				label := labels.FromStrings(tuple[0], tuple[1])
+				_, err := app.Add(label, i*blockRange, 0)
+				testutil.Ok(t, err)
+			}
+		}
+
+		err := app.Commit()
+		testutil.Ok(t, err)
+
```
gouthamve: Stick this here too:
```
labelNames, err := db.LabelNames()
testutil.Ok(t, err)
testutil.Equals(t, tst.exp, labelNames)
```

Would help test the block + head merge.

db.go
```diff
@@ -852,6 +852,63 @@ func (db *DB) CleanTombstones() (err error) {
...
+	}
+	for _, name := range names {
+		labelNamesMap[name] = struct{}{}
+	}
+	err = headIndexr.Close()
+	if err != nil {
+		return nil, errors.Wrap(err, "close head IndexReader")
+	}
+
+	db.mtx.RLock()
```
gouthamve:
```
db.mtx.RLock()
blocks := db.blocks[:]
db.mtx.RUnlock()

for _, b := range blocks {
}
```

Would avoid all of the hoop jumping :)

gouthamve:
@fabxc @krasi-georgiev This is good to go IMO, and this fixes a couple of hacks that current auto-complete systems do to get around this.

This is simple and would appreciate another set of eyes to see if it fits in here.

db_test.go
```diff
@@ -1301,3 +1301,92 @@ func TestInitializeHeadTimestamp(t *testing.T) {
 		testutil.Equals(t, int64(15000), db.head.MaxTime())
 	})
 }
+
+func TestDB_LabelNames(t *testing.T) {
```
krasi-georgiev: wouldn't something like this cover the same test cases, but a a bit more simplified?

```
func TestDB_LabelNames(t *testing.T) {
	tests := []struct {
		title        string
		sampleLabels [][2]string
		exp          [][]string
	}{
		{
			title: "Labels order",
			sampleLabels: [][2]string{
				[2]string{"name1", ""},
				[2]string{"name3", ""},
				[2]string{"name2", ""},
				[2]string{"name4", ""},
				[2]string{"name1", ""},
			},
			exp: [][]string{
				{"name"},
				{"name1", "name3"},
				{"name1", "name2", "name3"},
				{"name1", "name2", "name3", "name4"},
				{"name1", "name2", "name3", "name4"},
			},
		},
	}

	for _, tst := range tests {
		t.Run(tst.title, func(t *testing.T) {
			db, close := openTestDB(t, nil)
			defer close()
			defer db.Close()

			blockRange := DefaultOptions.BlockRanges[0]
			app := db.Appender()
			for i, tuple := range tst.sampleLabels {
				label := labels.FromStrings(tuple[0], tuple[1])
				_, err := app.Add(label, int64(i)*blockRange, 0)
				testutil.Ok(t, err)
				err = app.Commit()
				testutil.Ok(t, err)
				labelNames, err := db.LabelNames()
				testutil.Ok(t, err)
				testutil.Equals(t, tst.exp[i], labelNames)

			}

			// Check the labels after persisting to disk.
			_, err := db.compact()
			testutil.Ok(t, err)
			labelNames, err := db.LabelNames()
			testutil.Ok(t, err)
			testutil.Equals(t, tst.exp[len(tst.exp)-1], labelNames)
		})
	}
}

```

codesome: Here we are not checking the case where samples are there in both head and disk. So I had kept 2 sets: 1 set in head, 1 set on disk. I will see if there is any better way to test it.

krasi-georgiev: ok ping me when ready.

codesome: @krasi-georgiev I made the changes to make it look more clean. I would still like to have 2 sets of data so that testing head+disk phase is confusion free.

krasi-georgiev: yes it is good to test these separately, but how do you know that when calling `db.Labels()` these come from `blockIndexr.LabelNames()` or `headIndexr.LabelNames()`. Why not call these directly and than also call `db.LabelNames()` to make sure it returns the merged results?

db_test.go
```diff
@@ -1301,3 +1301,88 @@ func TestInitializeHeadTimestamp(t *testing.T) {
...
+				[2]string{"name0", ""},
+			},
+			exp1: []string{"name1", "name2"},
+			exp2: []string{"name0", "name1", "name2", "name6"},
+		},
+	}
+
+	blockRange := DefaultOptions.BlockRanges[0]
+	// Appends samples into the database.
+	appendSamples := func(db *DB, mint, maxt int64, sampleLabels [][2]string) {
```
krasi-georgiev: should be marked as a helper func so if it fails it shows the relevant test line number
https://www.tobstarr.com/2017/06/16/better-test-helpers-in-go/

db_test.go
```diff
@@ -1301,3 +1301,88 @@ func TestInitializeHeadTimestamp(t *testing.T) {
...
+				label := labels.FromStrings(tuple[0], tuple[1])
+				_, err := app.Add(label, i*blockRange, 0)
+				testutil.Ok(t, err)
+			}
+		}
+		err := app.Commit()
+		testutil.Ok(t, err)
+	}
+	// Tests for expected label names.
+	testEquals := func(db *DB, exp []string) {
```
krasi-georgiev: should be marked as a helper func so if it fails it shows the relevant test line number
https://www.tobstarr.com/2017/06/16/better-test-helpers-in-go/

codesome: Didn't know about that, thanks! I added.

db_test.go
krasi-georgiev: yes that is important , but in the current test how do you know that the labels come from the head or the disk when calling db.LabelNames() ? 
Why don't we call explicitly headIndexr.LabelNames(), blockIndexr.LabelNames() and db.LabelNames() at the end. 
This way maybe you can apply the same logic even to the simplified test version I suggested.

db_test.go
krasi-georgiev: yeah @simonpasquier did this in one PR I reviewed.

db_test.go
codesome: I had split the set in such a way that, each set gives new label names. So when called `db.LabelNames()` (for head+disk case), 1 set of unique names should come from head, and other from disk.

Checking directly from head and block is also good idea. I will add it.

krasi-georgiev: yes, but again this is only guessing since you call db.LabelNames() in both cases so best to be more explicit and call directly the head and the block readers.

codesome: I modified the test (with some renaming and sorting at places). Now I am checking head, block, db separately.

codesome:
@gouthamve @krasi-georgiev thoughts on this?

krasi-georgiev:
I agree with @brancz this would probably be the first thing we would need so it won't be premature to add it now.

db_test.go
krasi-georgiev: Thanks!

sampleLabels1, sampleLabels2 and exp1,exp2 is still bugging me, but right now can't think how to simplify.

codesome:
I want to know what kind of matchers are we looking at as label names is just a slice of strings. Is it like label names from all the series for the matchers?

brancz:
An array or variadic parameter of `labels.Matcher` elements would probably be appropriate.

codesome:
@brancz How would this work?
1. We get all the labels as usual and filter using the given matcher.
2. We return all the label names for the series returned by the matcher.

Or something else.

krasi-georgiev:
I would think so. 
Haven't looked in the code implementation yet, but whichever makes more sence.

gouthamve:
2. Would be it.

But I don't know the feasibility, I need to look at the code.
Thanks,
Goutham.

codesome:
After looking into the code and how index is stored, implementing option 2 would indeed be not feasible (it will require something similar to querying, please correct me if I am wrong). 

Currently index stores possible values only for single label name. https://github.com/prometheus/tsdb/issues/26 would be required to be addressed in order for it to be feasible. Still, we cannot have all combination of label-names stored in index.

index/index.go
```diff
@@ -38,6 +38,8 @@ const (
 
 	indexFormatV1 = 1
 	indexFormatV2 = 2
+
+	labelNameSeperator = "\xff"
```
krasi-georgiev: if you export this, there are view more places where you can use it.

gouthamve:
I don't have time to look at this right now, but at first glance `1` doesn't make sense at all. A matcher is from a particular `labelName` to a pattern on `labelValue`. I can't see how you can filter on labelnames using matchers.

brancz:
Agreed with @gouthamve .

index/index.go
codesome: Other than few test files, where I see it being used is `labels/labels.go` and `chunkenc/xor.go`.
In both places they are being used as different data type, and also in different context. So I would like to keep this index specific thing unexposed as those other packages are not related to index directly.

I would like to know where did you plan to reuse this.

krasi-georgiev:
IIUC option 1 can be implemented right now, but it would be slower right?
If that is the case I would say lets go for it and will optimise later.

Maybe try it and post the results and if the performance it dreadful will think again.

codesome:
Option 1 would not be slow. As number of label names wont be too high, filtering them using the matchers wont slow it down.

krasi-georgiev:
Problem solved than!
Rather than block it now, If/when option 1 would be possible can refactor.

codesome:
Done.

brancz:
hmm .. the way I read the code we are still loading all label names and then filter them, why not do it one layer further down, that sounds like it would make more sense. As in change the `LabelValues` function on the `IndexReader` interface to also receive the matchers in as a parameter and filter them on that level. Otherwise we allocate a large list with the intention to immediately filter it again, that seems unnecessary.

codesome:
Sounds fair enough, I will make that change.

db_test.go
```diff
@@ -1303,6 +1303,114 @@ func TestInitializeHeadTimestamp(t *testing.T) {
...
+			},
+			exp1: []string{"name1", "name2"},
+			exp2: []string{"name0", "name1", "name2", "name6"},
+		},
+	}
+
+	m, err := labels.NewRegexpMatcher("", ".*2|.*6")
+	testutil.Ok(t, err)
+	tests[1].ms = []labels.Matcher{m}
+	tests[1].expMs = []string{"name2", "name6"}
```
krasi-georgiev: instead of this you can add it in the test cases as:
```
exp1:  []string{"name1", "name2"},
exp2:  []string{"name0", "name1", "name2", "name6"},
ms:    []labels.Matcher{labels.NewMustRegexpMatcher("", ".*2|.*6")},
expMs: []string{"name2", "name6"},
```

krasi-georgiev:
@brancz I just had another look and the selector matcher is using the name, but in general it feels like trouble. I will think about it the next few days.

https://github.com/prometheus/tsdb/blob/cf2af2b371154209bb7b6b82bca20a16d040fca3/labels/selector.go#L25-L32

codesome:
The name in matcher would come into picture when we have option 2 for filtering label names. So I think it's fine to use that here to keep it future proof. Currently we would just ignore the name in the matcher.

brancz:
I think filtering label names based on full label matchers is fine, this is Howe all of Prometheus work a today. I get that filtering labelnames by some matcher could be desirable, but I think the matchers might be wildly different (as were not filtering label-value pairs but just label names). For now just filtering on all given full matchers is fine in my opinion.

fabxc:
Doesn't `LabelIndices()` also give you all label names? Given that it's a bit of an awkward signature to accommodate features we never ended up adding, maybe we can at least replace one with the other.

I'm personally against passing any matchers here. How many unique label names are we really looking at – maybe 1000? The caller can just loop over them and filter with whatever they want.

```go
names, _ := x.LabelNames()
filtered := names[:0]

for _, n := range names {
    if match(n) {
        filtered = append(filtered, n)
    }
}
```

Doesn't seem complex enough to justify first-class support.

krasi-georgiev:
+1 on replacing LabelIndices() with LabelNames(). this will yield a new release since we are breaking an api.

On a second thought I tend to lean towards NO marchers as well and filter on user side.

codesome:
`LabelIndices()` would give composite names `[][]string`. But `LabelNames()` is just aimed to give a single slice of unique label names. I will keep `LabelNames()` functionality unless any opposition.

krasi-georgiev:
>>  Oh, I don't know how I missed LabelIndices(). I can re-use it in db.LabelNames().

> My point is, that those will do literally the same thing. No need to keep around both of them.

my logic tells me that the proper way is to keep both(LabelNames,LabelIndices) and comment the `LabelIndices` as depreciated and tag a new minor release.
When we reach a point that need to break an existing API than we tag  a new major release and remove all depriciated APIs.

@fabxc , @gouthamve how does that sound?

krasi-georgiev:
@codesome 	Unless you have some other idea I think we should continue with my suggestion from the last comment.

codesome:
I will keep `LabelIndices` and mark it as deprecated.

krasi-georgiev:
ok ping me when ready for a review.

codesome:
@krasi-georgiev ping.

db.go
```diff
@@ -878,6 +878,53 @@ func (db *DB) CleanTombstones() (err error) {
...
+
+	labelNames := make([]string, 0, len(labelNamesMap))
+	for name := range labelNamesMap {
+		labelNames = append(labelNames, name)
+	}
+	sort.Slice(labelNames, func(i, j int) bool {
+		return labelNames[i] < labelNames[j]
+	})
+
+	return labelNames, nil
```
krasi-georgiev: you can use direclty the strings sort func.

```suggestion
	return sort.Strings(labelNames), nil
```

head.go
```diff
@@ -969,6 +969,23 @@ func (h *headIndexReader) LabelValues(names ...string) (index.StringTuples, erro
...
+		if name == "" {
+			continue
+		}
+		labelNames = append(labelNames, name)
+	}
+	sort.Slice(labelNames, func(i, j int) bool {
+		return labelNames[i] < labelNames[j]
+	})
+	return labelNames, nil
```
krasi-georgiev: you can use directly the strings sort func.

```suggestion
	return sort.Strings(labelNames), nil
```

index/index.go
```diff
@@ -935,6 +934,35 @@ func (r *Reader) SortedPostings(p Postings) Postings {
...
+	for name := range labelNamesMap {
+		labelNames = append(labelNames, name)
+	}
+	sort.Slice(labelNames, func(i, j int) bool {
+		return labelNames[i] < labelNames[j]
+	})
+	sort.Slice(labelNames, func(i, j int) bool {
+		return labelNames[i] < labelNames[j]
+	})
+	return labelNames, nil
```
krasi-georgiev: you can use directly the strings sort func.

```suggestion
	return sort.Strings(labelNames), nil
```

db_test.go
```diff
@@ -1303,6 +1303,97 @@ func TestInitializeHeadTimestamp(t *testing.T) {
 	})
 }
 
+func TestDB_LabelNames(t *testing.T) {
+	tests := []struct {
+		sampleLabels1 [][2]string // For checking head and disk.
+		// sampleLabels2 should have atleast 1 label name which is not there in sampleLabels1.
```
krasi-georgiev: don't quite understand what you mean here.

db_test.go
```diff
@@ -1303,6 +1303,97 @@ func TestInitializeHeadTimestamp(t *testing.T) {
...
+				[2]string{"name0", ""},
+			},
+			exp1: []string{"name1", "name2"},
+			exp2: []string{"name0", "name1", "name2", "name6"},
+		},
+	}
+
+	blockRange := DefaultOptions.BlockRanges[0]
+	// Appends samples into the database.
+	appendSamples := func(db *DB, mint, maxt int64, sampleLabels [][2]string) {
```
krasi-georgiev: I see few other test funcs use something like this.
Can we make this func more generic so we can reuse it more tests?

`populateAppender(app Appender,mint, maxt int64, labels) err`

db_test.go
```diff
@@ -1303,6 +1303,97 @@ func TestInitializeHeadTimestamp(t *testing.T) {
...
+		testutil.Ok(t, err)
+		testutil.Equals(t, tst.exp1, labelNames)
+		testutil.Ok(t, headIndexr.Close())
+
+		// Testing disk.
+		err = db.compact()
+		testutil.Ok(t, err)
+		// All blocks have same label names, hence check them individually.
+		// No need to aggregrate and check.
+		for _, b := range db.blocks {
```
```suggestion
		for _, b := range db.Blocks() {
```

querier_test.go
```diff
@@ -1493,10 +1493,16 @@ func (m mockIndex) Series(ref uint64, lset *labels.Labels, chks *[]chunks.Meta)
...
-
 	return res, nil
 }
+
+func (m mockIndex) LabelNames() ([]string, error) {
+	labelNames := make([]string, 0, len(m.labelIndex))
+	for name := range m.labelIndex {
+		labelNames = append(labelNames, name)
+	}
+	return labelNames, nil
```
krasi-georgiev:
```suggestion
	return sort.Strings(labelNames), nil
```

db_test.go
codesome: I see that most of them are adding samples in a different way. Only maybe 1 or 2 were adding sample in the way it is here (and had offsets of `+j` too). I think making this generalised and taking outside wont add much. 

I can add if you feel that would be better.

codesome:
@krasi-georgiev ping.

index/index.go
```diff
@@ -882,14 +883,12 @@ type emptyStringTuples struct{}
 func (emptyStringTuples) At(i int) ([]string, error) { return nil, nil }
 func (emptyStringTuples) Len() int                   { return 0 }
 
-// LabelIndices returns a for which labels or label tuples value indices exist.
+// LabelIndices returns a slice of label names for which labels or label tuples value indices exist.
+// NOTE: This is depricated. Use `LabelNames()` instead.
```
krasi-georgiev:
```suggestion
// NOTE: This is deprecated. Use `LabelNames()` instead.
```

db_test.go
```diff
@@ -1303,6 +1303,101 @@ func TestInitializeHeadTimestamp(t *testing.T) {
 	})
 }
 
+func TestDB_LabelNames(t *testing.T) {
+	tests := []struct {
+		// Add 'sampleLabels1' -> Test Head -> Compact -> Test Disk ->
+		//              -> Add 'sampleLabels2' -> Test Head+Disk
```
```suggestion
		// -> Add 'sampleLabels2' -> Test Head+Disk
```

db_test.go
```diff
@@ -1303,6 +1303,101 @@ func TestInitializeHeadTimestamp(t *testing.T) {
...
+
+		sampleLabels1 [][2]string // For checking head and disk.
```
```suggestion
		sampleLabels1 [][2]string // For checking head and disk separately.
```

db_test.go
```diff
@@ -1303,6 +1303,101 @@ func TestInitializeHeadTimestamp(t *testing.T) {
...
+		// To test Head+Disk, sampleLabels2 should have
+		// atleast 1 unique label name which is not there in sampleLabels1.
+		sampleLabels2 [][2]string // For check head+disk.
```
```suggestion
		sampleLabels2 [][2]string // For checking head and disk together.
```

db_test.go
```diff
@@ -1303,6 +1303,101 @@ func TestInitializeHeadTimestamp(t *testing.T) {
...
 }
 
+func TestDB_LabelNames(t *testing.T) {
+	tests := []struct {
+		// Add 'sampleLabels1' -> Test Head -> Compact -> Test Disk ->
+		//              -> Add 'sampleLabels2' -> Test Head+Disk
+
+		sampleLabels1 [][2]string // For checking head and disk.
+		// To test Head+Disk, sampleLabels2 should have
+		// atleast 1 unique label name which is not there in sampleLabels1.
```
```suggestion
		// at least 1 unique label name which is not in sampleLabels1.
```

krasi-georgiev:
@codesome  Thanks!

@gouthamve 	thanks for the pointer we can optimise later if needed.

PR #371, opened 2018-09-04
add wrapper for migrate wal (MigrateWalIfExists)

cstyan:
When reading the new WAL code I found the naming of this functionality a bit confusing. I've simply added a wrapper that does the check for an existing WAL in `dir` to `MigrateWalIfExists`, which in turn can call `migrateWal`.

Feel free to close if you prefer the old naming, I assume there may be a reason for not naming it this way in the first place.

### DIFF
--- a/db.go
+++ b/db.go
@@ -192,7 +192,7 @@ func Open(dir string, l log.Logger, r prometheus.Registerer, opts *Options) (db 
 	if err := repairBadIndexVersion(l, dir); err != nil {
 		return nil, err
 	}
-	// Migrate old WAL.
+	// Migrate old WAL if one exists.
 	if err := MigrateWAL(l, filepath.Join(dir, "wal")); err != nil {
 		return nil, errors.Wrap(err, "migrate WAL")
 	}
--- a/wal.go
+++ b/wal.go
@@ -1212,38 +1212,44 @@ func (r *walReader) decodeDeletes(flag byte, b []byte, res *[]Stone) error {
 	return nil
 }
 
-// MigrateWAL rewrites the deprecated write ahead log into the new format.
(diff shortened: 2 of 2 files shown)


krasi-georgiev:
yeah that func is rather long and it would be more readable if broken down a bit, but I would prefer something like:
```
func MigrateWAL(){
    if err := deprecatedWALExists(); err != nil{
       return err
    }

   // migrate....
}
```

this keeps the func single purpose by delegating other logic to the other funcs,

cstyan:
Yeah I originally started with something like that. I'll have another look at splitting the logic up more today.

krasi-georgiev:
thanks

cstyan:
Wait @krasi-georgiev that's the opposite of what I wanted to solve. I wanted the name to be more descriptive, so that now from `db.go` we know we're only migrating the WAL if one exists in the deprecated format. The diff isn't the easiest to read, but all I've done is move the wal detection out into another function `MigrateWalIfExists` that calls `migrateWal` if a deprecated WAL actually does exist.

krasi-georgiev:
I think this is a bit unnecessary. An additional comment would be sufficient.

gouthamve:
Yup, `MigrateWalIfExists` is unnecessary. I think `MigrateWAL` makes it clear (atleast for me) that it'll check for the existing WAL first. A comment would clarify that sufficiently.

cstyan:
Hmm, personal preference I guess. I would prefer the more descriptive name to a comment in cases like this.

Anyways, I'll swap the names around in the morning to match the `MigrateWAL` `deprecatedWALExists` format.

krasi-georgiev:
Thanks!

PR #372, opened 2018-09-04
clean up after running repair tests

cstyan:
I noticed that some files are changed and others are created when running tsdb tests

### DIFF
--- a/fileutil/fileutil.go
+++ b/fileutil/fileutil.go
@@ -1,108 +1,137 @@
-package tsdb
+// Package fileutil provides utility methods used when dealing with the filesystem in tsdb.
+// It is largely copied from github.com/coreos/etcd/pkg/fileutil to avoid the
+// dependency chain it brings with it.
+// Please check github.com/coreos/etcd for licensing information.
+package fileutil
 
 import (
+	"io/ioutil"
 	"os"
-	"testing"
-
-	"github.com/prometheus/tsdb/chunks"
-	"github.com/prometheus/tsdb/fileutil"
-	"github.com/prometheus/tsdb/index"
-	"github.com/prometheus/tsdb/labels"
-	"github.com/prometheus/tsdb/testutil"
+	"path/filepath"
+	"sort"
+	"strings"
 )
 
-func TestRepairBadIndexVersion(t *testing.T) {
-	// The broken index used in this test was written by the following script
-	// at a broken revision.
-	//
-	// func main() {
-	// 	w, err := index.NewWriter("index")
-	// 	if err != nil {
-	// 		panic(err)
-	// 	}
-	// 	err = w.AddSymbols(map[string]struct{}{
-	// 		"a": struct{}{},
-	// 		"b": struct{}{},
-	// 		"1": struct{}{},
-	// 		"2": struct{}{},
-	// 	})
-	// 	if err != nil {
-	// 		panic(err)
-	// 	}
-	// 	err = w.AddSeries(1, labels.FromStrings("a", "1", "b", "1"))
-	// 	if err != nil {
-	// 		panic(err)
-	// 	}
-	// 	err = w.AddSeries(2, labels.FromStrings("a", "2", "b", "1"))
-	// 	if err != nil {
-	// 		panic(err)
-	// 	}
-	// 	err = w.WritePostings("b", "1", index.NewListPostings([]uint64{1, 2}))
-	// 	if err != nil {
-	// 		panic(err)
-	// 	}
-	// 	if err := w.Close(); err != nil {
-	// 		panic(err)
-	// 	}
-	// }
-	const dbDir = "testdata/repair_index_version/01BZJ9WJQPWHGNC2W4J9TA62KC"
-	tmpDir := "testdata/repair_index_version/copy"
-	tmpDbDir := tmpDir + "/3MCNSQ8S31EHGJYWK5E1GPJWJZ"
-
-	// Check the current db.
-	// In its current state, lookups should fail with the fixed code.
-	meta, err := readMetaFile(dbDir)
-	testutil.NotOk(t, err)
-
-	// Touch chunks dir in block.
-	os.MkdirAll(dbDir+"/chunks", 0777)
-	defer os.RemoveAll(dbDir + "/chunks")
-
-	r, err := index.NewFileReader(dbDir + "/index")
-	testutil.Ok(t, err)
-	p, err := r.Postings("b", "1")
-	testutil.Ok(t, err)
-	for p.Next() {
-		t.Logf("next ID %d", p.At())
-
-		var lset labels.Labels
-		testutil.NotOk(t, r.Series(p.At(), &lset, nil))
+// CopyDirs copies all directories, subdirectories and files recursively including the empty folders.
+// Source and destination must be full paths.
+func CopyDirs(src, dest string) error {
+	if err := os.MkdirAll(dest, 0777); err != nil {
+		return err
 	}
-	testutil.Ok(t, p.Err())
-	testutil.Ok(t, r.Close())
-
-	// Create a copy DB to run test against.
-	if err = fileutil.CopyDirs(dbDir, tmpDbDir); err != nil {
-		t.Fatal(err)
-	}
-	// defer os.RemoveAll(tmpDir)
-	// On DB opening all blocks in the base dir should be repaired.
-	db, err := Open(tmpDir, nil, nil, nil)
-	testutil.Ok(t, err)
-	db.Close()
-
-	r, err = index.NewFileReader(tmpDbDir + "/index")
-	testutil.Ok(t, err)
-	p, err = r.Postings("b", "1")
-	testutil.Ok(t, err)
-	res := []labels.Labels{}
-
-	for p.Next() {
-		t.Logf("next ID %d", p.At())
-
-		var lset labels.Labels
-		var chks []chunks.Meta
-		testutil.Ok(t, r.Series(p.At(), &lset, &chks))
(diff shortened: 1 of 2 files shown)


krasi-georgiev:
Yeah I noticed that too, but I think the problem is with the test itself. It should revert the state after completing. Otherwise it doesn't allow running it twice and confuses git as well

I also  prefer to use  `.git/info/exclude` for such edge use-cases and other user specific workflows.

cstyan:
@krasi-georgiev updated so that we're cleaning up after the repair tests, reverted the gitignore changes. It's a bit ugly atm but I need to sleep. I'll clean it up it in the morning.

krasi-georgiev:
thanks , ping me when ready and I will review again.

cstyan:
@krasi-georgiev ping :)

krasi-georgiev:
I just had a look at the test I did for another PR and here is a lot easyer way to achieve the same:

```
// Create and work on the db copy so we can run the test many times.
tmpdir := "testdata/repair_index_chunks/snapshot"
defer os.RemoveAll(tmpdir)
testutil.Ok(t, db.Snapshot(tmpdir, false))
dbCopy, err := Open(tmpdir, nil, nil, DefaultOptions)
testutil.Ok(t, err)
```

cstyan:
@krasi-georgiev yeah I tried that initially, we need tombstones in order to create a snapshot. I don't we want to deal with generating tombstones for this test.

krasi-georgiev:
even if you need to put an empty tombstones file it would be a much cleaner solution without requiring so many lines of additional code.

cstyan:
if all we needed was an empty tombstones file I would have just done that :) let me have another look

cstyan:
@krasi-georgiev we can't use snapshot here anyways, it's the call to `db, err := Open("testdata/repair_index_version/", nil, nil, DefaultOptions)` that actually modifies the existing files.

krasi-georgiev:
I made it work using:
```
bl := Block{
   meta: BlockMeta{
	ULID: ulid.ULID{},
   },
   dir: "testdata/repair_index_version/01BZJ9WJQPWHGNC2W4J9TA62KC",
}
copy(bl.meta.ULID[:], bl.dir)

tmpdir := "testdata/repair_index_chunks/snapshot"
testutil.Ok(t, bl.Snapshot(tmpdir))
```

I think we would need to do this quite often form now on so  why don't we make some more generic func which takes filepaths as a param and recursively makes hard links using `os.Link` and put it in:
`fileutil/fileutil.go`

Something like:
```
func CopyFiles( files ...string, destDir string){
  //  iterate and os.Link all files
}
```
than we can have: `CopyFile(ReadDir(dirpath),destDir)`

if you don't have time for the extra effort let me know and I can take over.

cstyan:
Thanks @krasi-georgiev, I got it working via the block snapshotting. 

Unfortunately I don't understand this generic copy files helper, how do hardlinks help us here?

krasi-georgiev:
look at the snapshot implementation it uses hard links instead of copying files.

krasi-georgiev:
hardlinks is essentially like copying , but without even opening the files.

cstyan:
Right, but it's linked to the same underlying data on disk. So modifying the file at the hardlink name modifies the file at the original name, because they're essentially the same file.

Looking at the snapshot code, I don't understand how using the hardlinks gets us a copy of the DB files that we can modify without making modifications to the originals.

krasi-georgiev:
`CloneFiles` is another name that comes to mind.

krasi-georgiev:
ooh yeah I missunderstood the hardlinks a bit. 
Lets use the original idea with  `os.Open/io.Copy`, but just make the func something like: `CopyFile(ReadDir(dirpath),destDir)` as I have another test that would use it right away.

repair_test.go
```diff
@@ -1,16 +1,43 @@
...
 	"testing"
 
 	"github.com/prometheus/tsdb/chunks"
+	"github.com/prometheus/tsdb/fileutil"
 
 	"github.com/prometheus/tsdb/index"
 	"github.com/prometheus/tsdb/labels"
 )
 
+func copyFiles(tmpIndex, tmpMeta *os.File, dir string) error {
```
krasi-georgiev: don't need this anymore.

cstyan:
`fileutil.ReadDir` just returns the names as strings, we'd need the path to the files. Also we need to handle the case where some contents of `srcDir` are directories even if we don't also want to recursively copy from sub-directories.

If `func CopyFiles( files ...string, destDir string)` works better for your use case I can make that change, but I don't think passing `fileutil.ReadDir(srcDir)` is useful unless srcdir happens to also be `os.Getwd()`.

krasi-georgiev:
```
// Reads files in the src directory and all sub directories recursively. 
func ReadDirs(dir) map[path]name, error{
  recursively walk all folders including the root dir and build the  `map[path]name` using the `func ReadDir(...)`
}

CopyFiles(ReadDirs(srcDir),dstDir)
```

something like this should do the trick. 

Than I will try if it will work for my usecase as well.

cstyan:
@krasi-georgiev imo it's easier if `ReadDirs` returns `[]string`, but how do you want it to handle empty directories? For example for this repair index test, we need the `chunks` dir to exist in the copy of the tsdb, but we just create an empty db in the real dir before copying it. I can just add another mkdir in the copy, or make `ReadDirs` and `CopyFiles` handle empty dirs as well.

krasi-georgiev:
I would say lets create dirs regardless of whether they will include any files. 
This solves our use case so it is a good start. We can always update later if needed.

krasi-georgiev:
for `[]string` vs `map[path]name` just do whichever would require less lines of code.

fileutil/fileutil.go
```diff
@@ -5,11 +5,101 @@
...
+	}
+
+	if copied, err := io.Copy(temp, f); err != nil || copied != stat.Size() {
+		return err
+	}
+	if err = os.Rename(temp.Name(), dest); err != nil {
+		return err
+	}
+	return nil
+}
```
krasi-georgiev: why not something more simple like:

```
 data, err := ioutil.ReadFile(src)
 if err != nil {
     return err
  }

 err = ioutil.WriteFile(dst, data, 0644)
 if err != nil {
      return err
 }
```

fileutil/fileutil.go
```diff
@@ -5,11 +5,102 @@
...
+	// path/copy/to/file, this function should return /copy
+	dir, _ := filepath.Split(fullPath)
+	chunks := strings.Split(dir, remove)
+	if len(chunks) != 2 {
+		return ""
+	}
+	return chunks[1]
+}
+
+func copyFile(dest, src string) error {
```
krasi-georgiev: Usually the order is src, dst

fileutil/fileutil.go
```diff
@@ -5,11 +5,102 @@
...
+func getExtraPath(fullPath, remove string) string {
+	// If we have /path/to/file and we want to copy it to
+	// path/copy/to/file, this function should return /copy
+	dir, _ := filepath.Split(fullPath)
+	chunks := strings.Split(dir, remove)
+	if len(chunks) != 2 {
+		return ""
+	}
+	return chunks[1]
+}
```
krasi-georgiev: Can you explain why do we need this func, can't understand from the comments.

krasi-georgiev:
> see the latest commit, we also needed to grab extra portions of the path for files we're creating copies of so that the end up in the right destination location

this looks hacky, can you try to simplify.

cstyan:
yes

cstyan:
@krasi-georgiev simplified a bit. We still need a way to remove a prefix from the files we find when using `filepath.Walk`, but we can use `strings.TrimPrefix` rather than the function I had written.

krasi-georgiev:
can you also try the simplified version I suggested for  `copyFile`.

cstyan:
@krasi-georgiev yep, just slipped my mind earlier. 

Want me to squash commits now?

krasi-georgiev:
no need I will squash when I merge it.


moved few bits around and updated few comment.

```
// CopyDirs copies all directories, subdirectories and files recursively including the empty folders.
// Source and destination must be full paths.
func CopyDirs(src, dest string) error {
	if err := os.MkdirAll(dest, 0777); err != nil {
		return err
	}
	files, err := readDirs(src)
	if err != nil {
		return err
	}

	for _, f := range files {
		dp := filepath.Join(dest, f)
		sp := filepath.Join(src, f)

		stat, err := os.Stat(sp)
		if err != nil {
			return err
		}

		// Empty directories are also created.
		if stat.IsDir() {
			if err := os.MkdirAll(dest, 0777); err != nil {
				return err
			}
			continue
		}

		if err := copyFile(sp, dp); err != nil {
			return err
		}
	}
	return nil
}

func copyFile(src, dest string) error {
	data, err := ioutil.ReadFile(src)
	if err != nil {
		return err
	}

	err = ioutil.WriteFile(dest, data, 0644)
	if err != nil {
		return err
	}
	return nil
}

// readDirs reads the source directory recursively
// returns relative paths to all files and empty directories.
func readDirs(src string) ([]string, error) {
	var files []string
	var err error

	err = filepath.Walk(src, func(path string, f os.FileInfo, err error) error {
		relativePath := strings.TrimPrefix(path, src)
		if len(relativePath) > 0 {
			files = append(files, relativePath)
		}
		return nil
	})
	if err != nil {
		return nil, err
	}
	return files, nil
}
```

cstyan:
:+1: changes lgtm, pushed them

krasi-georgiev:
resolve the conflict and LGTM

krasi-georgiev:
Thanks for spending the time on this.

Issue #309, opened 2018-03-22
compact/write: Don't write down empty blocks (?)


bwplotka:
I had no samples for a while in my Prometheus (because of wrong scraping configuration?). TSDB dumped empty block into my disk which looks like this:
```
01C6XA7C2K2EJKV6GGWQCGNYTC/
-- index (39.53 KB)
-- meta.json (408 B)
-- chunks/
     < empty dir >
```
Meta file:
```
{
    "version": 1,
    "ulid": "01C6XA7C2K2EJKV6GGWQCGNYTC",
    "minTime": 1519236000000,
    "maxTime": 1519243200000,
    "stats": {},
    "compaction": {
        "level": 1,
        "sources": [
            "01C6XA7C2K2EJKV6GGWQCGNYTC"
        ]
    },
}
```
This causes some tiny problems in [Thanos](https://github.com/improbable-eng/thanos) that we need to workaround, but my main questions are:
1) Have you seen that before? How often can that happen?
2) **Is that necessary to produce those?**

In my personal opinion, it can cause lot's of unnecessary compactions (and downsampling for Thanos) and index lookups. It gets even worse for Thanos or other long storage solutions, because these lookups & compactions are bit more expensive. 

Do you think we can just don't write empty blocks down?

The only negative consequence I can think of is the inability to tell from the storage perspective (looking on the blocks) if the Prom/TSDB was down or just Prometheus did not scrape anything (no samples in TSDB).

gouthamve:
Even I think that writing empty blocks can be removed.

See: https://github.com/prometheus/tsdb/pull/245

PR #374, opened 2018-09-08
Don't write empty blocks

codesome:
Fixes https://github.com/prometheus/tsdb/issues/309
Supersedes https://github.com/prometheus/tsdb/pull/245


compact.go
```diff
@@ -508,8 +508,13 @@ func (c *LeveledCompactor) write(dest string, meta *BlockMeta, blocks ...BlockRe
 	df = nil
 
 	// Block successfully written, make visible and remove old ones.
-	if err := renameFile(tmp, dir); err != nil {
-		return errors.Wrap(err, "rename block dir")
+	if meta.Stats.NumSamples > 0 {
```
krasi-georgiev: why not do this check at the right top of the func  and just return early.

krasi-georgiev:
LGTM

compact.go
krasi-georgiev: yep you are correct.

krasi-georgiev:
actually I just thought it will be good to add a test for this with some comments why we don't want empty blocks.

gouthamve:
Yup, a test for this would be good. LGTM otherwise.

codesome:
And the test found out a bug. The compaction of head was going on in an infinite loop, but not sure why.

One workaround that I tested (which worked properly) was that, send the dir of empty block in `Plan()`. And in compaction delete the blocks which are empty.

Does this sound fine?

krasi-georgiev:
can you add the test so I can check what is happening.

codesome:
@krasi-georgiev I added the test case.

gouthamve:
> when adding empty sample the head should not change the min/maxtime

We never add "empty" sample :) We can delete samples though. It's kinda hard to set the mintime during `delete`.

krasi-georgiev:
@codesome still stuck with the test?

codesome:
@krasi-georgiev Stuck with exams :stuck_out_tongue:. I will be back on it tomorrow.

codesome:
@gouthamve @krasi-georgiev 
The thing that is broken in this PR feels not straight forward to fix (will dig in more), so I have come up with this alternative approach.

https://github.com/prometheus/tsdb/compare/master...codesome:remove-empty-block-in-compact?expand=1

This allows you to write empty block during compact, but removes it immediately after. More like, empty blocks might appear during `compact()`, but not after it has finished. How does this look?

codesome:
@krasi-georgiev Actually the test is not broken, the code is broken. I actually found a fix for it (tested, working), I will push it soon.

db_test.go
```diff
@@ -1301,3 +1301,53 @@ func TestInitializeHeadTimestamp(t *testing.T) {
...
+
+	_, err = db.compact()
+	testutil.Ok(t, err)
+	testutil.Equals(t, 1, len(db.blocks))
+
+	// No blocks after deleting all samples from disk.
+	testutil.Ok(t, db.Delete(math.MinInt64, math.MaxInt64, labels.NewEqualMatcher("foo", "bar")))
+	_, err = db.compact()
+	testutil.Ok(t, err)
+	testutil.Equals(t, 0, len(db.blocks))
```
codesome: Test would fail here. https://github.com/prometheus/tsdb/pull/385 needs to get merged for this to be fixed. Also after that this might need some changes to get the block selected for compaction, will look at it after #385 is merged.

codesome: This no longer needs to wait for the above.

codesome:
@krasi-georgiev This is good for anther review.
/cc @gouthamve

compact.go
```diff
@@ -471,17 +486,24 @@ func (c *LeveledCompactor) write(dest string, meta *BlockMeta, blocks ...BlockRe
...
-	}
-
 	if err = chunkw.Close(); err != nil {
 		return errors.Wrap(err, "close chunk writer")
 	}
 	if err = indexw.Close(); err != nil {
 		return errors.Wrap(err, "close index writer")
 	}
 
+	if meta.Stats.NumSamples == 0 {
```
krasi-georgiev: small comment: // Populated block is empty so cleanup and exit.

db.go
```diff
@@ -383,6 +383,22 @@ func (db *DB) compact() (changes bool, err error) {
...
+		db.mtx.RLock()
+		l := len(db.blocks)
+		db.mtx.RUnlock()
+		if l == 0 {
+			err = db.head.Truncate(maxt)
+			if err != nil {
+				return changes, errors.Wrap(err, "head truncate failed (in compact)")
+
+			}
+		}
```
krasi-georgiev: why not move this logic inside:

```
func (db *DB) reload() (err error) {
....
if len(blocks) == 0 {
        // Truncate head and return
}
```

db_test.go
```diff
@@ -1301,3 +1301,62 @@ func TestInitializeHeadTimestamp(t *testing.T) {
...
+		testutil.Ok(t, err)
+	}
+	err = app.Commit()
+	testutil.Ok(t, err)
+
+	testutil.Ok(t, db.Delete(math.MinInt64, math.MaxInt64, labels.NewEqualMatcher("foo", "bar")))
+	_, err = db.compact()
+	testutil.Ok(t, err)
+	// No blocks created.
+	testutil.Equals(t, 0, len(db.blocks))
```
krasi-georgiev: shouldn't we first create a block test that `len(db.blocks) > 0` and than run the Delete logic to test that the empty block was deleted?

db_test.go
```diff
@@ -1301,3 +1301,62 @@ func TestInitializeHeadTimestamp(t *testing.T) {
...
+
+	// Test no blocks remaining after small samples are deleted from disk.
```
krasi-georgiev: what do you mean by `small samples` ?

db_test.go
```diff
@@ -1301,3 +1301,62 @@ func TestInitializeHeadTimestamp(t *testing.T) {
...
+			_, err := app.Add(label, i*blockRange+j, 0)
+			testutil.Ok(t, err)
+		}
+	}
+	err = app.Commit()
+	testutil.Ok(t, err)
+
+	_, err = db.compact()
+	testutil.Ok(t, err)
+	testutil.Assert(t, len(db.blocks) > 0, "No blocks created")
```
krasi-georgiev: something more descriptive: "No blocks created when compacting with more than 0 samples."

db_test.go
```diff
@@ -1301,3 +1301,62 @@ func TestInitializeHeadTimestamp(t *testing.T) {
...
+
+	testutil.Ok(t, db.Delete(math.MinInt64, math.MaxInt64, labels.NewEqualMatcher("foo", "bar")))
+
+	// Mimicking small part of compaction.
```
krasi-georgiev: I am bit confused here. Can you give a bit more details what is happening and what are testing for.

codesome: So we want to test 2 cases here
1. When head is being compacted to create new block. Here if it results in 0 samples then we write no blocks.
2. (which you asked) When compacting a block or multiple blocks, if it results in an empty block, we should test that. Here I am deleting all samples so that no blocks should remain. While testing, `db.compact()` was not selecting the blocks for compaction (basically `Plan()` in compactor was not selecting those directories). So here I am manually mimicking `Plan()` (selecting all blocks) and calling the compactor on those blocks.

db.go
codesome: To do this we need to pass the time at which head has to be truncated. I did that first by having something like a wrapper around `reload()` which takes the time. But making that change just for a special case in compaction looked like overkill.

Do you still want me to include it inside `reload()`?

db_test.go
krasi-georgiev: yep just saw it.

db.go
krasi-georgiev: can't you use `mint, maxt := rangeForTimestamp(db.head.MinTime(), db.opts.BlockRanges[0])` within  `reload()` ?

db_test.go
krasi-georgiev: yep I understand now. so we need some better comments here than.

Something to describe the workflow without looking at each func implementation.

// Build a list of block paths to pass to the compactor.
...
// The compactor should mark all empty blocks as deletable.
...
// The reload should remove all deletable blocks.

db_test.go
```diff
@@ -1301,3 +1301,62 @@ func TestInitializeHeadTimestamp(t *testing.T) {
 		testutil.Equals(t, int64(15000), db.head.MaxTime())
 	})
 }
+
+func TestNoEmptyBlock(t *testing.T) {
```
krasi-georgiev: not a deal breaker, but I prefer this plural - `TestNoEmptyBlocks`

krasi-georgiev:
lets also update the comment for the `Compactor interface`

```
// Compact runs compaction against the provided directories. Must
// only be called concurrently with results of Plan().
// Compaction resulting in an empty block marks all parents as deletable in its meta data..
Compact(dest string, dirs ...string) (ulid.ULID, error)
``

File: db.go (hunk 4 above)
Comment (codesome) [reply db.go]: There is possibility of `reload()` being called in other places too. In that case if blocks were 0, we would end up compacting the head unnecessarily, don't you think so?

File: db_test.go (hunk 8 above)
Comment (codesome) [reply db_test.go]: Updated the comments, with few more checks.

File: block.go [hunk 10]
```diff
@@ -178,6 +178,10 @@ type BlockMetaCompaction struct {
 	Level int `json:"level"`
 	// ULIDs of all source head blocks that went into the block.
 	Sources []ulid.ULID `json:"sources,omitempty"`
+	// As we dont write empty blocks, we need this to mark
+	// if the block is deletable if in compaction this block
```
krasi-georgiev: if ~in~  during compaction

db.go
```diff
@@ -383,6 +383,22 @@ func (db *DB) compact() (changes bool, err error) {
 		if err := db.reload(); err != nil {
 			return changes, errors.Wrap(err, "reload blocks")
 		}
+		// After this is fixed https://github.com/prometheus/tsdb/issues/309,
+		// if the compaction of head did not create any new blocks (possible
+		// if the samples were deleted in that range), and if number of
+		// blocks are 0, db.reload() does not truncate head. Hence MinTime
+		// of head is not set properly. So only for that situation, we need
+		// to call db.head.Truncate(..) manually.
```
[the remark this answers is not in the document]
codesome: This comment was a bit confusing to me. I will take goods of both and combine :)

block.go
```diff
@@ -179,6 +179,10 @@ type BlockMetaCompaction struct {
 	Level int `json:"level"`
 	// ULIDs of all source head blocks that went into the block.
 	Sources []ulid.ULID `json:"sources,omitempty"`
+	// As we dont write empty blocks, we need this to mark
+	// if the block is deletable if during compaction this block
+	// resulted in an empty block.
```
krasi-georgiev: // Indicates that during compaction it resulted in a block without any samples 
// so it should be deleted on the next reload.

krasi-georgiev:
@gouthamve one final look?

codesome:
Fixed it. It was a minor addition.

db.go
```diff
@@ -401,7 +401,8 @@ func (db *DB) compact() (err error) {
 			// from the block interval here.
 			maxt: maxt - 1,
 		}
-		if _, err = db.compactor.Write(db.dir, head, mint, maxt, nil); err != nil {
+		var uid ulid.ULID
+		if uid, err = db.compactor.Write(db.dir, head, mint, maxt, nil); err != nil {
 			return errors.Wrap(err, "persist head block")
 		}
```
krasi-georgiev: I prefer this as 

```
uid, err := db.compactor.Write(db.dir, head, mint, maxt, nil)
if err != nil {
	return errors.Wrap(err, "persist head block")
}
```

codesome:
@krasi-georgiev I have already added the test. It was running into infinite loop at `db.compact()` before this fix.
https://github.com/prometheus/tsdb/blob/d89e3c7950af9bb02dbe8329b02a09e2c0b63db8/db_test.go#L1348-L1357

krasi-georgiev:
weird the test is failing only with go1.10.4

codesome:
I tested again with `go 1.10.4` in my system and it seems to be working fine. Do we need to re-run this test?

krasi-georgiev:
yeah I just triggered another run, but if it failed once probably it will be flaky.

krasi-georgiev:
still failing on the same one.

codesome:
Which is very strange. I will have a look.

krasi-georgiev:
maybe another rebase?

codesome:
I already rebased yesterday (just checked, up to date with master). Apparently it's not related to it. I am not sure why is the error, tests doesn't seem to fail in my system.

Can you test and let me know if you get the same error?

krasi-georgiev:
passes on my system as well.

krasi-georgiev:
it passed 3 times and now  it fails every time again as if it keeps some state.

krasi-georgiev:
@codesome the problem was the when you open a db and add random samples this triggers the auto compaction and writes the bocks before you have the chance to delete them.

so with the changes in https://github.com/prometheus/tsdb/pull/412 and adding `db.DisableCompactions()` this will fix he tests.

db_test.go
```diff
@@ -1303,6 +1303,97 @@ func TestInitializeHeadTimestamp(t *testing.T) {
...
 }
 
+func TestNoEmptyBlocks(t *testing.T) {
+	db, close := openTestDB(t, nil)
+	defer close()
+	defer db.Close()
+
+	// Test no blocks after compact with empty head.
+	testutil.Ok(t, db.compact())
+	testutil.Equals(t, 0, len(db.blocks))
```
krasi-georgiev: use `len(db.Blocks())` everywhere to avoid races

krasi-georgiev:
merged the PR for disabling compaction so you resolve the conflict and fix the tests.

codesome:
Will do it in couple of days, or by this weekend, caught up with exams at my University.

codesome:
Done. Got some free time.
I will review the code again to see if any cases are missing.

codesome:
The test has passed this time even without disabling compactions manually. Still, will disable.

db.go
krasi-georgiev: yeah i still think we should avoid this edge case somehow. the size based retention PR https://github.com/prometheus/tsdb/pull/343 simplifies quite a few places so lets wait for that one to be merged and will revisit.

db_test.go
```diff
@@ -1303,6 +1303,100 @@ func TestInitializeHeadTimestamp(t *testing.T) {
...
+		testutil.Ok(t, err)
+	}
+	testutil.Ok(t, app.Commit())
+
+	testutil.Ok(t, db.Delete(math.MinInt64, math.MaxInt64, labels.NewEqualMatcher("foo", "bar")))
+	uid, err := db.compactor.Write(db.dir, db.head, db.head.MinTime(), db.head.MaxTime(), nil)
+	testutil.Ok(t, err)
+	testutil.Equals(t, ulid.ULID{}, uid)
+	testutil.Ok(t, db.reload())
+	testutil.Ok(t, db.head.Truncate(db.head.MaxTime()))
```
krasi-georgiev: why do we need to truncate? the tests pass even without this.

db_test.go
```diff
@@ -1303,6 +1303,100 @@ func TestInitializeHeadTimestamp(t *testing.T) {
...
+	// No blocks created.
+	testutil.Equals(t, 0, len(db.Blocks()))
```
krasi-georgiev: don't you think it would be better to actually check the blocks on disk rather than relying on db.Blocks()

```
bb, err := blockDirs(db.Dir())
testutil.Ok(t, err)
testutil.Equals(t, 0, len(bb))
```

db_test.go
```diff
@@ -1303,6 +1303,100 @@ func TestInitializeHeadTimestamp(t *testing.T) {
...
+	app = db.Appender()
+	for i := int64(7); i < 25; i++ {
+		for j := int64(0); j < 10; j++ {
+			_, err := app.Add(label, i*blockRange+j, 0)
+			testutil.Ok(t, err)
+		}
+	}
+	testutil.Ok(t, app.Commit())
+
+	testutil.Ok(t, db.compact())
```
krasi-georgiev: is there any reason why we are not using 
`db.compactor.Write(db.dir, db.head, db.head.MinTime(), db.head.MaxTime(), nil)` everywhere? I would rather avoid using block ranges as it is a bit unclear why we need something like `_, err := app.Add(label, i*blockRange+j, 0)` to trigger a new block creation. Also depending on `db.head.MaxTime()-db.head.MinTime() <= db.opts.BlockRanges[0]/2*3` will make the tests fragile if this logic changes.

krasi-georgiev:
> The test has passed this time even without disabling compactions manually. Still, will disable.

yeah that was flaky even before. Sometimes it would pass, but disabling compaction definitely fixes that.

db_test.go
codesome: If we compact head manually using Write, we cannot come across many edge cases which I came across while using `db.compact()` (example running into infinite loop, not truncating head properly, etc). I would prefer keeping it like this, as this is the one called.

And I think block range is the range of time for a block? So if the samples in the head span multiple blocks, then head would be compacted. If in future with any changes new blocks were not created, then we would know by the assert below.

db_test.go
codesome: I think `db.reload()` takes care of that. `db.blocks` is updated with the blocks on disk in `db.reload()`.
https://github.com/prometheus/tsdb/blob/38a2c6b6bd454443896a401c236fdea8383441d5/db.go#L534-L554

db_test.go
codesome: I tried to keep the test as if it's going ahead in time. So after compact all the samples in head are removed, so we truncate till that time (which is the max time). Test might pass, but this way looked better to me.

codesome: Basically head is truncated till the point it is compacted.

compact.go
```diff
@@ -413,6 +427,10 @@ func (c *LeveledCompactor) Write(dest string, b BlockReader, mint, maxt int64, p
 		return uid, err
 	}
 
+	if meta.Stats.NumSamples == 0 {
+		return ulid.ULID{}, nil
```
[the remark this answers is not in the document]
codesome: I updated the comments. And about CHANGELOG, I added the `[CHANGE]` info in the PR description so that later when new version will be tagged, it will be easy to add in that.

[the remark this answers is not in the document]
krasi-georgiev: no I meant to actually create a change log file and add it there.

db_test.go
krasi-georgiev: if there is a bug in db.reload and Blocks() might return incorrect blocks as well.

db_test.go
krasi-georgiev: in that case we should refactor something so that we don't have to rely in internal implementation and just test the behaviour.

krasi-georgiev: or maybe test one level up in the api.

codesome: There is a check after this function call to make sure that new block was created. Would that be sufficient?

db_test.go
```diff
@@ -1309,6 +1309,112 @@ func TestInitializeHeadTimestamp(t *testing.T) {
...
+	// Test no blocks after compact with empty head.
+	testutil.Ok(t, db.compact())
+	bb, err := blockDirs(db.Dir())
+	testutil.Ok(t, err)
+	testutil.Equals(t, len(db.Blocks()), len(bb))
+	testutil.Equals(t, 0, len(bb))
+
+	// Test no blocks after deleting all samples from head.
+	blockRange := DefaultOptions.BlockRanges[0]
+	label := labels.FromStrings("foo", "bar")
```
krasi-georgiev: can you rename this to `defaultLabel` and use it everywhere as `labels.NewEqualMatcher(defaultLabel[0].Name, defaultLabel[0].Value))

db_test.go
krasi-georgiev: maybe to make it more explicit use what I ended up doing in
https://github.com/prometheus/tsdb/blob/bac9cbed2eb417e0637ba2bed59d070b37deed44/db_test.go#L1468

db_test.go
```diff
@@ -1309,6 +1309,112 @@ func TestInitializeHeadTimestamp(t *testing.T) {
...
+	testutil.Ok(t, app.Commit())
+	testutil.Ok(t, db.head.Delete(math.MinInt64, math.MaxInt64, labels.NewEqualMatcher("foo", "bar")))
+	testutil.Ok(t, db.compact())
+	testutil.Equals(t, oldBlocks, db.Blocks())
+
+	// Test no blocks remaining after deleting all samples from disk.
+	testutil.Ok(t, db.Delete(math.MinInt64, math.MaxInt64, labels.NewEqualMatcher("foo", "bar")))
+
+	// Mimicking Plan() of compactor and getting list
+	// of all block directories to pass for compaction.
```
krasi-georgiev: might be better to use openTestDB with approproate opts so that db.compact() results in proper compaction instead of mimicking it. and using the helper `createBlock(..`  to create some test blocks or `app.Add`

db_test.go
krasi-georgiev: yep I see what you mean. So why not use `db.compact()` here as well which already includes this `Truncating` than we can also check that the head was Truncated as expected.

db_test.go
```diff
@@ -1309,6 +1309,112 @@ func TestInitializeHeadTimestamp(t *testing.T) {
...
+	plan := []string{}
+	for _, b := range db.Blocks() {
+		plan = append(plan, b.Dir())
+	}
+
+	// Blocks are not set deletable before compaction.
```
krasi-georgiev: I don't we need the meta structs. Just checking that compaction results in no new blocks and deleting the old ones is fine.

codesome:
Was not able to address this yet: https://github.com/prometheus/tsdb/pull/374#discussion_r248280494, rest all reviews have been addressed.

db_test.go
codesome: Tried, but not able to tune in properly to select all blocks for compaction. Help needed.

codesome:
ping @krasi-georgiev @gouthamve

krasi-georgiev:
LGTM

compact.go
```diff
@@ -186,19 +191,18 @@ func (c *LeveledCompactor) plan(dms []dirMeta) ([]string, error) {
 		return res, nil
 	}
 
-	// Compact any blocks that have >5% tombstones.
+	// Compact any blocks with big enough time range that have >5% tombstones.
 	for i := len(dms) - 1; i >= 0; i-- {
 		meta := dms[i].meta
 		if meta.MaxTime-meta.MinTime < c.ranges[len(c.ranges)/2] {
-			break
+			continue
```
gouthamve: https://github.com/prometheus/tsdb/pull/374/files#diff-43297cfd6441b52c54c50d646a564339R178

The above ensures that we can break here. Basically we're going from big blocks to small blocks and if we find a small block, we can be assured that further blocks will be the same size if not smaller.

compact.go
```diff
@@ -186,19 +191,18 @@ func (c *LeveledCompactor) plan(dms []dirMeta) ([]string, error) {
...
 		}
-
 		if float64(meta.Stats.NumTombstones)/float64(meta.Stats.NumSeries+1) > 0.05 {
-			return []string{dms[i].dir}, nil
+			res = append(res, dms[i].dir)
```
gouthamve: This is technically going to compact blocks together that should instead be rewritten. For example, the ranges that should not be compacted together will be. Returning a single block is the right thing to do.

krasi-georgiev:
all done

PR #375, opened 2018-09-08
Added metric for symbol table size

codesome:
Closes https://github.com/prometheus/tsdb/issues/244, Supersedes https://github.com/prometheus/tsdb/pull/272

`prometheus_tsdb_symbol_table_size` will show the total size of symbol table of persisted blocks. 

### DIFF
--- a/db.go
+++ b/db.go
@@ -119,6 +119,7 @@ type DB struct {
 
 type dbMetrics struct {
 	loadedBlocks         prometheus.GaugeFunc
+	symbolTableSize      prometheus.GaugeFunc
 	reloads              prometheus.Counter
 	reloadsFailed        prometheus.Counter
 	compactionsTriggered prometheus.Counter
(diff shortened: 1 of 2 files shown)


db.go
```diff
@@ -138,6 +139,19 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
 		defer db.mtx.RUnlock()
 		return float64(len(db.blocks))
 	})
+	m.symbolTableSize = prometheus.NewGaugeFunc(prometheus.GaugeOpts{
+		Name: "prometheus_tsdb_symbol_table_size",
```
gouthamve: See: https://prometheus.io/docs/practices/naming/#metric-names

Should be `prometheus_tsdb_symbol_table_size_bytes`

brian-brazil:
Yes, it's a way to spot high cardinality issues.

gouthamve:
Note that this is `the total` bytes. I'm wondering if a histogram would make more sense.

brian-brazil:
I think the total's fine.

gouthamve:
Thanks for the input!

PR #377, opened 2018-09-12
fixed refactoring for `serializedStringTuples1`

### LINKED ISSUE #241: Refactor `serializedStringTuples` used in the index
Refactor the type along with its methods. Make them more readable, rename variables currently, it's confusing. https://github.com/prometheus/tsdb/blob/master/index/index.go#L929-L933

knrt10:
closes #241. @krasi-georgiev would you please review and tell is this what you want?

### DIFF
--- a/index/index.go
+++ b/index/index.go
@@ -868,9 +868,9 @@ func (r *Reader) LabelValues(names ...string) (StringTuples, error) {
 		return nil, errors.Wrap(d.err(), "read label value index")
 	}
 	st := &serializedStringTuples{
-		l:      nc,
-		b:      d.get(),
-		lookup: r.lookupSymbol,
+		length:  nc,
+		breadth: d.get(),
+		lookup:  r.lookupSymbol,
 	}
 	return st, nil
 }
@@ -929,8 +929,8 @@ func (r *Reader) Postings(name, value string) (Postings, error) {
 
 // SortedPostings returns the given postings list reordered so that the backing series
 // are sorted.
-func (r *Reader) SortedPostings(p Postings) Postings {
-	return p
+func (r *Reader) SortedPostings(post Postings) Postings {
+	return post
 }
 
 type stringTuples struct {
@@ -973,23 +973,23 @@ func (t *stringTuples) Less(i, j int) bool {
 }
 
 type serializedStringTuples struct {
-	l      int
-	b      []byte
-	lookup func(uint32) (string, error)
+	length  int
+	breadth []byte
+	lookup  func(uint32) (string, error)
 }
 
 func (t *serializedStringTuples) Len() int {
-	return len(t.b) / (4 * t.l)
+	return len(t.breadth) / (4 * t.length)
 }
 
 func (t *serializedStringTuples) At(i int) ([]string, error) {
-	if len(t.b) < (i+t.l)*4 {
+	if len(t.breadth) < (i+t.length)*4 {
 		return nil, errInvalidSize
 	}
-	res := make([]string, 0, t.l)
+	res := make([]string, 0, t.length)
 
-	for k := 0; k < t.l; k++ {
-		offset := binary.BigEndian.Uint32(t.b[(i+k)*4:])
+	for k := 0; k < t.length; k++ {
+		offset := binary.BigEndian.Uint32(t.breadth[(i+k)*4:])
 
 		s, err := t.lookup(offset)
 		if err != nil {
--- a/index/index.go
+++ b/index/index.go
@@ -392,7 +392,7 @@ func (w *Writer) WriteLabelIndex(names []string, values []string) error {
 	w.buf2.putBE32int(valt.Len())
 
 	// here we have an index for the symbol file if v2, otherwise it's an offset
-	for _, v := range valt.s {
+	for _, v := range valt.entries {
 		index, ok := w.symbols[v]
 		if !ok {
 			return errors.Errorf("symbol entry for %q does not exist", v)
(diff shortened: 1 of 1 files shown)


index/index.go
```diff
@@ -929,8 +929,8 @@ func (r *Reader) Postings(name, value string) (Postings, error) {
 
 // SortedPostings returns the given postings list reordered so that the backing series
 // are sorted.
-func (r *Reader) SortedPostings(p Postings) Postings {
-	return p
+func (r *Reader) SortedPostings(post Postings) Postings {
+	return post
```
krasi-georgiev: when we use a variable in the next 1-2 lines it is fine to have it as a single letter.

index/index.go
```diff
@@ -868,9 +868,9 @@ func (r *Reader) LabelValues(names ...string) (StringTuples, error) {
 		return nil, errors.Wrap(d.err(), "read label value index")
 	}
 	st := &serializedStringTuples{
-		l:      nc,
-		b:      d.get(),
-		lookup: r.lookupSymbol,
+		length:  nc,
+		breadth: d.get(),
```
krasi-georgiev: what does `breadth` stand for?

index/index.go
```diff
@@ -868,9 +868,9 @@ func (r *Reader) LabelValues(names ...string) (StringTuples, error) {
 		return nil, errors.Wrap(d.err(), "read label value index")
 	}
 	st := &serializedStringTuples{
-		l:      nc,
-		b:      d.get(),
-		lookup: r.lookupSymbol,
+		length:  nc,
```
krasi-georgiev: I need to read the code deeper, but I think this represents a the 32 bit big Indian version and not a length.

index/index.go
knrt10: Ok, will change it.

index/index.go
knrt10: I was confused as [here](https://github.com/prometheus/tsdb/pull/377/files#diff-71ebe2bcf31a915b1fa3b3b289d5d31dR977) in the definition of struct **serializedStringTuples** it was previously slice of bytes and didn't knew what it should be. What do you suggest should I change it to?

index/index.go
knrt10: I thought as according to [here](https://github.com/prometheus/tsdb/blob/3df5bcf39e9fb4dea3f70f9ec2bcb77c23271268/index/index.go#L937). What should I change it too?

index/index.go
knrt10: @krasi-georgiev after reading code deeper and some [go doc](https://golang.org/pkg/encoding/binary/#example_ByteOrder_get) I found that its not `breadth` its just **slices of binary data**. So instead of binary I changes it to *binSlice*. And *l* can be used as length because it is just size of serializedString tuple. I am saying beacause [here](https://github.com/prometheus/tsdb/blob/master/index/index.go#L864) **nc** is being initialized as `be32int` that is an integer returned from [here](https://github.com/prometheus/tsdb/blob/master/index/encoding_helpers.go#L78). And also we are using **l** just to iterate [here](https://github.com/prometheus/tsdb/blob/master/index/index.go#L991). So as per me *lenght* is just fine. If you suggest anything else I can change it to that. I have pushed latest changes please review.

knrt10: @krasi-georgiev updated.

krasi-georgiev:
@knrt10 while we are on the topic , should we also update the `stringTuples` struct field names?

knrt10:
> @knrt10 while we are on the topic , should we also update the `stringTuples` struct field names?

Yes on it.

krasi-georgiev:
not sure , look at the implementation.

knrt10:
> not sure , look at the implementation.

Ok let me check

knrt10:
@krasi-georgiev I have changed it to 
```
type stringTuples struct {
	tupLength int      // tuple length
	tupSlice  []string // flattened tuple entries
}
```
is it ok?

krasi-georgiev:
I think so , 
now we don't even need the additional comments as the var name is good enough to understand its purpose.

krasi-georgiev:
Now that I see it stutters a bit. Lets try as:
```
type stringTuples struct {
	length int      // tuple length
	entries  []string // flattened tuple entries
}
```

knrt10:
Do i have to leave the comment as it is or remove them?

krasi-georgiev:
lets leave it.

knrt10:
Updated

krasi-georgiev:
shall we also change `idsSlice -> idsEntries` to make it consistent?

knrt10:
Yep, updated

krasi-georgiev:
@gouthamve what do you think does this change makes it more clear?

knrt10:
@krasi-georgiev any updates on this?

krasi-georgiev:
still waiting for @gouthamve to confirm.
This will  take few days as we have some more urgent  bugfixes.

krasi-georgiev:
ping @gouthamve

krasi-georgiev:
seems @gouthamve is busy putting out fires in other places :stuck_out_tongue: 
@codesome could you take a look

index/index.go
```diff
@@ -973,23 +973,23 @@ func (t *stringTuples) Less(i, j int) bool {
 }
 
 type serializedStringTuples struct {
-	l      int
-	b      []byte
-	lookup func(uint32) (string, error)
+	idsCount   int
+	idsEntries []byte
```
codesome: As this is not exactly entries but the bytes containing them, I would ideally name them `idsBytes` (previous naming was `b`, I guess it meant bytes).

It was not totally clear why prefix is `ids` (later saw the comments). Can you add a comment saying these bytes contains the ids pointing to actual string in the lookup table.

krasi-georgiev: yep +1 on both comments.

knrt10: So @krasi-georgiev what should I do? Please don't mind me asking as I don't quite fully understand.

krasi-georgiev:
```
idsBytes []byte // bytes containing the ids pointing to the string in the lookup table.
```

knrt10: Updated

PR #383, opened 2018-09-17
Remove `prometheus_` prefix from metrics

beorn7:
This can now be added by users of the library as needed with the new
https://godoc.org/github.com/prometheus/client_golang/prometheus#WrapRegistererWithPrefix

(The latter to be done in the prometheus/prometheus repo separately, going along with a vendoring update.)

@gouthamve @brancz 

Signed-off-by: beorn7 <beorn@soundcloud.com>

### DIFF
--- a/compact.go
+++ b/compact.go
@@ -84,30 +84,30 @@ func newCompactorMetrics(r prometheus.Registerer) *compactorMetrics {
 	m := &compactorMetrics{}
 
 	m.ran = prometheus.NewCounter(prometheus.CounterOpts{
-		Name: "prometheus_tsdb_compactions_total",
+		Name: "tsdb_compactions_total",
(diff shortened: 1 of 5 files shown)


beorn7:
And also @grobie as he chimed in on the old PR.

brancz:
lgtm :+1:

grobie:
As long as the metric names in Prometheus don't change, totally cool with this change.

gouthamve:
👍

brancz:
> As long as the metric names in Prometheus don't change, totally cool with this change.

That's the idea.

beorn7:
Follow-up PR for prometheus/prometheus is being worked on.

brian-brazil:
I believe we should revert this PR, "tsdb_" is a very generic prefix as there are many many tsdb out there.

Issue #384, opened 2018-09-18
Incorrect NumTombstones after block.Delete(...)


codesome:
Found this while working on https://github.com/prometheus/tsdb/pull/374

Apparently updating `NumTombstones` only [here](https://github.com/prometheus/tsdb/blob/a8966cb53d39747c5d7b7ba787873b52899881d0/block.go#L487) is incorrect. For example, if currently there were no stones and we want to add new stones in `Delete(...)`, `NumTombstones` is nowhere updated.

Please correct me if I am wrong. I will open a PR for this soon.

gouthamve:
Yes you are right. This was introduced in this: https://github.com/prometheus/tsdb/pull/212

Please fix it and add a test for the same :)

PR #385, opened 2018-09-18
Fix updating of NumTombstones in block.Delete(..)

codesome:
Fixes https://github.com/prometheus/tsdb/issues/384

### DIFF
--- a/block.go
+++ b/block.go
@@ -484,7 +484,6 @@ Outer:
 	err = pb.tombstones.Iter(func(id uint64, ivs Intervals) error {
 		for _, iv := range ivs {
 			stones.addInterval(id, iv)
-			pb.meta.Stats.NumTombstones++
 		}
 		return nil
 	})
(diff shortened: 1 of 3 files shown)


krasi-georgiev:
I would say something like this would be a bit more clear and better to avoid races:

```

// TombstoneReader gives access to tombstone intervals by series reference.
type TombstoneReader interface {
	...
        // Total returns the total count of Tombstones.
	Total() uint64
}


func (t *memTombstones) Total() uint64 {
	t.mtx.RLock()
	defer t.mtx.RUnlock()

	var total uint64
	for _, intvs := range t.intvlGroups {
		total += uint64(len(intvs))
	}
	return total
}


func (pb *Block) Delete
....
pb.meta.Stats.NumTombstones = pb.tombstones.Total()

```

codesome:
There wont be any race condition in the current change, but what you proposed looks good. I will add that function.

krasi-georgiev:
LGTM 

ping @gouthamve

krasi-georgiev:
Thanks!

PR #387, opened 2018-09-18
return ErrOutOfOrderSample even on the same appender

krasi-georgiev:
supersedes: https://github.com/prometheus/tsdb/pull/258

ping @gouthamve 

any pointers how to run some good benchmarking?

I have tried `tsdb bench write` which shows very mixed results. 

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -301,14 +301,14 @@ func TestOutOfOrderDatapointCausesError(t *testing.T) {
 	app := db.Appender()
 	_, err := app.Add(labels.Labels{}, 1, 0)
 	testutil.Ok(t, err)
-	_, err = app.Add(labels.Labels{}, 1, 0)
+	_, err = app.Add(labels.Labels{}, 0, 0)
 	testutil.Equals(t, ErrOutOfOrderSample, err)
 
 	testutil.Ok(t, app.Commit())
 
 	// Also test that it errors on a new appender as well.
 	app = db.Appender()
-	_, err = app.Add(labels.Labels{}, 1, 0)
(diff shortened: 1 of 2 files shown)


gouthamve:
Hmm, this fails for `Rollback`. Basically if there is `Rollback` the `lastTime` would be an invalid one.

krasi-georgiev:
updated.

What worries is me that it brakes a test that specifically checks that adding an ooo sample doesn't cause an error as if this is desired behaviour.

https://github.com/prometheus/tsdb/blob/d1a6edae6f790fd286a3f5bf01274da9e5f53111/db_test.go#L351-L357

krasi-georgiev:
ping @gouthamve

krasi-georgiev:
@gouthamve any memories regarding https://github.com/prometheus/tsdb/pull/387#issuecomment-422594066 ?

krasi-georgiev:
@fabxc do you remember why out of order sample are allowed to be added to the head?

fabxc:
What do you mean? They aren't – hence the error.

krasi-georgiev:
before the change in this PR adding  ooo samples don't return an error and they are just dropped.

there is even a test that specifically checks that adding ooo samples don't return an error.

https://github.com/prometheus/tsdb/blob/d1a6edae6f790fd286a3f5bf01274da9e5f53111/db_test.go#L351-L357

fabxc:
Oh, that's what you mean. Yea, probably would be better to return some
PartialError that indicates how many samples have been dropped.
All callers would need to be adapted accordingly of course.

krasi-georgiev:
@fabxc  ^

PR #388, opened 2018-09-18
Fix assert order being backwards in a few places

csmarchbanks:
In a few places testutil.Equals arguments are in the wrong order (supposed to be expected, result). It was a little confusing when I got an incorrect message when a test failed.

Signed-off-by: Chris Marchbanks <csmarchbanks@gmail.com>


krasi-georgiev:
LGTM on green, Thanks!
ping @gouthamve

krasi-georgiev:
:+1:

PR #389, opened 2018-09-19
Make sure WAL Repair can handle wrapped errors

gouthamve:
Fixes https://github.com/prometheus/prometheus/issues/4603

### DIFF
--- a/wal/wal.go
+++ b/wal/wal.go
@@ -255,15 +255,17 @@ Loop:
 
 // Repair attempts to repair the WAL based on the error.
 // It discards all data after the corruption.
-func (w *WAL) Repair(err error) error {
+func (w *WAL) Repair(origErr error) error {
 	// We could probably have a mode that only discards torn records right around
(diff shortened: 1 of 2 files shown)


wal/wal.go
```diff
@@ -261,6 +261,8 @@ func (w *WAL) Repair(err error) error {
 	// But that's not generally applicable if the records have any kind of causality.
 	// Maybe as an extra mode in the future if mid-WAL corruptions become
 	// a frequent concern.
+	err = errors.Cause(err) // So that we can pick up errors even if wrapped.
+
 	cerr, ok := err.(*CorruptionErr)
 	if !ok {
 		return errors.New("cannot handle error")
```
fabxc: This should probably show the original error as well, i.e. `errors.Wrap(origErr, "cannot handle error")`

gouthamve:
@fabxc Yep, that would have made it easier to debug. Done.

PR #392, opened 2018-09-19
Fix filehandling for windows

gouthamve:
Windows requires it's files to be closed before being removed. We've been bitten by this before. 

Having tests doesn't help if we don't run them :/ See: https://github.com/prometheus/prometheus/issues/4622

### DIFF
--- a/compact.go
+++ b/compact.go
@@ -452,6 +452,7 @@ func (c *LeveledCompactor) write(dest string, meta *BlockMeta, blocks ...BlockRe
 	if err != nil {
 		return errors.Wrap(err, "open chunk writer")
 	}
+	defer chunkw.Close()
 	// Record written chunk sizes on level 1 compactions.
 	if meta.Compaction.Level == 1 {
 		chunkw = &instrumentedChunkWriter{
@@ -466,6 +467,7 @@ func (c *LeveledCompactor) write(dest string, meta *BlockMeta, blocks ...BlockRe
 	if err != nil {
 		return errors.Wrap(err, "open index writer")
 	}
+	defer indexw.Close()
 
 	if err := c.populateBlock(blocks, meta, indexw, chunkw); err != nil {
 		return errors.Wrap(err, "write compaction")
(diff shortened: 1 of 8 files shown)


wal.go
```diff
@@ -1272,7 +1272,6 @@ func MigrateWAL(logger log.Logger, dir string) (err error) {
 	if err != nil {
 		return errors.Wrap(err, "open old WAL")
 	}
-	defer w.Close()
```
brian-brazil: If there's an error, below removing this could mean the wal is left open

wal/wal.go
```diff
@@ -311,16 +311,18 @@ func (w *WAL) Repair(origErr error) error {
 	if err != nil {
 		return errors.Wrap(err, "open segment")
 	}
-	defer f.Close()
```
brian-brazil: Same here

gouthamve:
@brian-brazil @fabxc @krasi-georgiev PTAL. These are non-trivial and might break things, I'm working on testing this now.

gouthamve:
WAL migration and compactions work on Windows now. Tested with `min-block-duration` set to `2m`.

gouthamve:
And checkpointing works on Windows! I think this is good to go!

wal.go
```diff
@@ -735,10 +742,12 @@ func (w *SegmentWAL) Close() error {
 	// On opening, a WAL must be fully consumed once. Afterwards
 	// only the current segment will still be open.
 	if hf := w.head(); hf != nil {
-		return errors.Wrapf(hf.Close(), "closing WAL head %s", hf.Name())
+		if err := hf.Close(); err != nil {
+			return errors.Wrapf(hf.Close(), "closing WAL head %s", hf.Name())
```
brian-brazil: Why are you calling hf.Close twice?

gouthamve: Typo, fixed.

brian-brazil:
:+1:

compact.go
```diff
@@ -475,6 +477,10 @@ func (c *LeveledCompactor) write(dest string, meta *BlockMeta, blocks ...BlockRe
 		return errors.Wrap(err, "write merged meta")
 	}
 
+	// We are explicitly closing them here to check for error even
+	// though these are covered under defer. This is because in Windows,
+	// you cannot delete these unless there is they are closed and we want to close them
+	// in case of any errors above.
```
krasi-georgiev: > This is because in Windows,
you cannot delete these unless there is they are closed and we want to close them
in case of any errors above.

I don't understand this comment

gouthamve:
I've let this run overnight on Darwin, Linux and Windows. I've checked the migration again on windows and darwin. Everything runs pretty well, so this is good to go!

Issue #373, opened 2018-09-07
Metrics for WAL truncation failure


gouthamve:
From: https://github.com/prometheus/prometheus/issues/4578

> I could not find any metric that got incremented for a failure to truncate the WAL. That would at least allow to alert on a non-truncating WAL before the disk runs full.

There should be metrics for both the ignored (rather, just logged) errors here:
https://github.com/prometheus/tsdb/blob/master/checkpoint.go#L266-L277

PR #396, opened 2018-09-25
Add new metrics.

codesome:
Fixes https://github.com/prometheus/tsdb/issues/373.

This is the best I could think of to prevent passing entire `head` or `headMetrics` into `Checkpoint(..)`.

### DIFF
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -23,10 +23,7 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/go-kit/kit/log"
-	"github.com/go-kit/kit/log/level"
 	"github.com/pkg/errors"
-	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/tsdb/fileutil"
 	"github.com/prometheus/tsdb/wal"
 )
@@ -101,12 +98,7 @@ const checkpointPrefix = "checkpoint."
 // segmented format as the original WAL itself.
 // This makes it easy to read it through the WAL package and concatenate
 // it with the original WAL.
-//
-// Non-critical errors are logged and not returned.
-func Checkpoint(logger log.Logger, w *wal.WAL, m, n int, keep func(id uint64) bool, mint int64, checkpointDeleteFail prometheus.Counter) (*CheckpointStats, error) {
-	if logger == nil {
-		logger = log.NewNopLogger()
-	}
+func Checkpoint(w *wal.WAL, m, n int, keep func(id uint64) bool, mint int64) (*CheckpointStats, error) {
 	stats := &CheckpointStats{}
 
 	var sr io.Reader
@@ -272,19 +264,6 @@ func Checkpoint(logger log.Logger, w *wal.WAL, m, n int, keep func(id uint64) bo
 	}
 	if err := closeAll(closers...); err != nil {
 		return stats, errors.Wrap(err, "close opened files")
-	}
-	if err := w.Truncate(n + 1); err != nil {
-		// If truncating fails, we'll just try again at the next checkpoint.
-		// Leftover segments will just be ignored in the future if there's a checkpoint
-		// that supersedes them.
-		level.Error(logger).Log("msg", "truncating segments failed", "err", err)
-	}
-	if err := DeleteCheckpoints(w.Dir(), n); err != nil {
-		// Leftover old checkpoints do not cause problems down the line beyond
-		// occupying disk space.
-		// They will just be ignored since a higher checkpoint exists.
-		level.Error(logger).Log("msg", "delete old checkpoints", "err", err)
-		checkpointDeleteFail.Add(float64(1))
 	}
 	return stats, nil
 }
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -26,6 +26,7 @@ import (
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/tsdb/fileutil"
 	"github.com/prometheus/tsdb/wal"
 )
@@ -102,7 +103,7 @@ const checkpointPrefix = "checkpoint."
 // it with the original WAL.
 //
 // Non-critical errors are logged and not returned.
-func Checkpoint(logger log.Logger, w *wal.WAL, m, n int, keep func(id uint64) bool, mint int64) (*CheckpointStats, error) {
+func Checkpoint(logger log.Logger, w *wal.WAL, m, n int, keep func(id uint64) bool, mint int64, checkpointDeleteFail prometheus.Counter) (*CheckpointStats, error) {
 	if logger == nil {
 		logger = log.NewNopLogger()
 	}
@@ -283,6 +284,7 @@ func Checkpoint(logger log.Logger, w *wal.WAL, m, n int, keep func(id uint64) bo
 		// occupying disk space.
 		// They will just be ignored since a higher checkpoint exists.
 		level.Error(logger).Log("msg", "delete old checkpoints", "err", err)
+		checkpointDeleteFail.Add(float64(1))
 	}
 	return stats, nil
 }
--- a/head.go
+++ b/head.go
@@ -76,19 +76,20 @@ type Head struct {
 }
 
 type headMetrics struct {
-	activeAppenders     prometheus.Gauge
-	series              prometheus.Gauge
-	seriesCreated       prometheus.Counter
-	seriesRemoved       prometheus.Counter
-	seriesNotFound      prometheus.Counter
-	chunks              prometheus.Gauge
-	chunksCreated       prometheus.Counter
-	chunksRemoved       prometheus.Counter
-	gcDuration          prometheus.Summary
-	minTime             prometheus.GaugeFunc
-	maxTime             prometheus.GaugeFunc
-	samplesAppended     prometheus.Counter
-	walTruncateDuration prometheus.Summary
+	activeAppenders      prometheus.Gauge
+	series               prometheus.Gauge
+	seriesCreated        prometheus.Counter
+	seriesRemoved        prometheus.Counter
+	seriesNotFound       prometheus.Counter
+	chunks               prometheus.Gauge
+	chunksCreated        prometheus.Counter
+	chunksRemoved        prometheus.Counter
+	gcDuration           prometheus.Summary
+	minTime              prometheus.GaugeFunc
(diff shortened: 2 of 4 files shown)


krasi-georgiev:
why not just put these  in
https://github.com/prometheus/tsdb/blob/a971f52ac8c41ecf8ca54325b312ddab97b1f3aa/wal.go#L58-L61

codesome:
Thanks, didn't know about that. That would cover `prometheus_tsdb_wal_truncate_fail`, we might still want to pass `prometheus_tsdb_checkpoint_delete_fail` to the `Checkpoint(...)`.

krasi-georgiev:
yes I am still looking through the code to see if we can simplify that part as well.

simonpasquier:
I'd add another metric (eg `prometheus_tsdb_checkpoints_failed_total`) in [`head.go`](https://github.com/prometheus/tsdb/blob/a971f52ac8c41ecf8ca54325b312ddab97b1f3aa/head.go#L472-L474) that would be increased whenever `Checkpoint()` returns a non-nil error.

head.go
```diff
@@ -150,6 +152,14 @@ func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
 		Name: "prometheus_tsdb_head_samples_appended_total",
 		Help: "Total number of appended samples.",
 	})
+	m.checkpointDeleteFail = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "prometheus_tsdb_checkpoint_deletions_failed_total",
+		Help: "Total number of checkpoint deletions that failed.",
+	})
+	m.checkpointsFail = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "prometheus_tsdb_checkpoints_failed_total",
+		Help: "Total number of create checkpoint that failed.",
```
simonpasquier: (nit) "Total number of checkpoints that failed."

codesome: Done.

codesome:
I don't see anything wrong in what you proposed. Should I wait for another ack or shall I go ahead with this?

krasi-georgiev:
it is an easy change so I would say go ahead and if anyone objects it is easy to revert.

krasi-georgiev:
ping me when ready for another review.

wal/wal.go
```diff
@@ -530,13 +535,15 @@ func (w *WAL) Segments() (m, n int, err error) {
...
 	if err != nil {
+		w.truncateFail.Inc()
 		return err
 	}
 	for _, r := range refs {
 		if r.n >= i {
 			break
 		}
 		if err := os.Remove(filepath.Join(w.dir, r.s)); err != nil {
+			w.truncateFail.Inc()
```
krasi-georgiev: not a deal breaker, but to make it more bulletproof against further code changes how about adding this somewhere in the beginning:
```
defer func() {
    if err!=nil{
       w.truncateFail.Inc()
    }
}()
```


this way even if we add more blocks that might return an error wouldn't have to change the logic for the metrics.

head.go
```diff
@@ -150,6 +152,14 @@ func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
...
 		Help: "Total number of appended samples.",
 	})
+	m.checkpointDeleteFail = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "prometheus_tsdb_checkpoint_deletions_failed_total",
+		Help: "Total number of checkpoint deletions that failed.",
+	})
+	m.checkpointsFail = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "prometheus_tsdb_checkpoints_failed_total",
+		Help: "Total number of checkpoints that failed.",
+	})
```
krasi-georgiev: I don't think we have a use case that required this metric and I prefer to add lines of code only when we need to.

checkpoint.go
```diff
@@ -272,17 +271,5 @@ func Checkpoint(logger log.Logger, w *wal.WAL, m, n int, keep func(id uint64) bo
...
-		// Leftover segments will just be ignored in the future if there's a checkpoint
-		// that supersedes them.
-		level.Error(logger).Log("msg", "truncating segments failed", "err", err)
-	}
-	if err := DeleteCheckpoints(w.Dir(), n); err != nil {
-		// Leftover old checkpoints do not cause problems down the line beyond
-		// occupying disk space.
-		// They will just be ignored since a higher checkpoint exists.
-		level.Error(logger).Log("msg", "delete old checkpoints", "err", err)
-	}
```
krasi-georgiev: now that we have moved these should also update the Checkpoint func comment
remove - // Non-critical errors are logged and not returned.

wal/wal.go
codesome: Thought about that :) I modified now.

checkpoint.go
codesome: Removed. Also removed logger as an argument as it's no longer needed.

head.go
simonpasquier: The initial [request](https://github.com/prometheus/prometheus/issues/4578) from @beorn7 was

> I could not find any metric that got incremented for a failure to truncate the WAL. That would at least allow to alert on a non-truncating WAL before the disk runs full.

This is what the new `prometheus_tsdb_wal_truncate_failed_total` metric is about but it doesn't cover for errors in `head.Truncate()`. So I would say that we need only one additional metric `prometheus_tsdb_head_truncate_total` that is incremented every time `head.Truncate()` fails. The exact cause (checkpoint deletion failed or any other file system error) can be determined by looking at the logs.

codesome: If I understand correctly, we are looking at only 2 metrics in whole, `prometheus_tsdb_wal_truncate_failed_total` and `prometheus_tsdb_head_truncate_failed_total`, correct?

But another thing is that, failing of `DeleteCheckpoints(...)` is not considered as head truncate fail, so do we ignore it?

simonpasquier: > But another thing is that, failing of DeleteCheckpoints(...) is not considered as head truncate fail, so do we ignore it?

Right, I missed that part. Then we can keep `prometheus_tsdb_checkpoint_deletions_failed_total`.

codesome: @krasi-georgiev I have removed `prometheus_tsdb_checkpoints_failed_total`, rather added `prometheus_tsdb_head_truncate_failed_total`. Now these 3 metrics should cover `head.Truncate(...)` entirely.

krasi-georgiev: @simonpasquier The main idea is to know when we have a failure that results in increasing the used hdd space which is when a checkpoint deletion fails or when the wal truncation fails.
In this regard why do you think we need to know about `prometheus_tsdb_head_truncate_total`? 
What could we use this metric for?

gouthamve:
Sorry, for jumping in late for the review, this generally LGTM, but I think we need to add the total number of attempts for each failure too. Metrics are cheap, and shouldn't be a problem to add a couple of more if they are going to make debugging easy.

See: https://prometheus.io/docs/practices/instrumentation/#failures
> When reporting failures, you should generally have some other metric representing the total number of attempts. This makes the failure ratio easy to calculate.

Further, we don't have a metric for see the number of failures for creating a `Checkpoint` and I think that should be added too. I am not fully awake yet, so, feel free to let me know if I got something wrong anywhere.

simonpasquier:
:+1: to what @gouthamve says. As I've written before, it is not just about checkpoint deletions that fail. We have blind spots on `head.Truncate()` and this PR needs to solve them all,

simonpasquier:
@codesome yes. By convention, metric names are `foo_failed_total` and `foo_total` (instead of `foo_attempt_total`).

* `prometheus_tsdb_wal_truncations_total`
* `prometheus_tsdb_head_truncations_total`
* `prometheus_tsdb_checkpoint_creations_failed_total`
* `prometheus_tsdb_checkpoint_creations_total`

codesome:
Updated. Now it contains metrics for
* Head truncation fail and total
* WAL truncation fail and total
* Checkpoint creation fail and total
* Checkpoint deletion fail and total

head.go
```diff
@@ -150,6 +156,30 @@ func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
...
+	m.checkpointDeleteFail = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "prometheus_tsdb_checkpoint_deletions_failed_total",
+		Help: "Total number of checkpoint deletions that failed.",
+	})
+	m.checkpointDeleteTotal = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "prometheus_tsdb_checkpoint_deletions_total",
+		Help: "Total number of checkpoint deletions attempted.",
+	})
+	m.checkpointCreationFail = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "prometheus_tsdb_checkpoint_creation_failed_total",
```
simonpasquier: (nit) `prometheus_tsdb_checkpoint_creations_failed_total`

head.go
```diff
@@ -150,6 +156,30 @@ func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
...
+		Help: "Total number of checkpoint creations that failed.",
+	})
+	m.checkpointCreationTotal = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "prometheus_tsdb_checkpoint_creation_total",
```
simonpasquier: (nit) `prometheus_tsdb_checkpoint_creations_total`

simonpasquier:
2 small comments otherwise :+1:

krasi-georgiev:
LGTM
@gouthamve or @simonpasquier feel free to merge once your comments are addressed.

codesome:
I was wondering if there is some bug in travis test. Error says `./db_test.go:1323:9: cannot assign 1 values to 2 variables`, but I can see only 1304 lines in [db_test.go](https://github.com/prometheus/tsdb/blob/61b000ee0e6a6095ef194e18f47a42ebb51c2acb/db_test.go).

It is failing similarly in https://github.com/prometheus/tsdb/pull/398

PS: Test passes in my system with no build fails.

krasi-georgiev:
when I clone the pr using 
https://help.github.com/articles/checking-out-pull-requests-locally/

I do see 1323, but it is as it should be : `err := db.compact()`

maybe try rebasing?

codesome:
Yup rebasing fixed it. Didn't know that tests are performed this way.

simonpasquier:
@codesome thanks for the work!

checkpoint.go
fabxc: Why did those get moved? This is an exported API with other tools depending on this to be in place.

codesome: @krasi-georgiev I think we need to move it back to `Checkpoint`.

Issue #395, opened 2018-09-24
Missing unit on lowest timestamp metric.


SuperQ:
In https://github.com/prometheus/tsdb/pull/363 a new metric, `prometheus_tsdb_lowest_timestamp`, was added.

It seems like this is milliseconds, but is missing the unit.

Maybe we should convert this to seconds and make it `prometheus_tsdb_lowest_timestamp_seconds`.

simonpasquier:
See https://github.com/prometheus/tsdb/pull/363#discussion_r204637082 for the rationale. We have the same situation with the `prometheus_tsdb_head_max_time` and `prometheus_tsdb_head_min_time` metrics.

SuperQ:
I don't think that rationale makes sense. If it's a timestamp, it has to have a unit, which we should know by the datatype.

brian-brazil:
The unit is defined by the code that uses it though, not the tsdb. To the TSDB this is just a number.

SuperQ:
Maybe the TSDB shouldn't expose this metric, but we have Prometheus create the metric instead so we have an opinion on what the type is.

krasi-georgiev:
What are we doing with this one?

PR #401, opened 2018-10-02
add a metric for the last time a sample was logged to the WAL

cstyan:
As part of changing Prometheus' remote write path to read data from the WAL rather than gathering it from scrapes, it will be useful to compare the last time data was written to the WAL vs. the last time each remote write config. successfully sent samples to it's remote storage.

### DIFF
--- a/wal/wal.go
+++ b/wal/wal.go
@@ -164,6 +164,7 @@ type WAL struct {
 	pageCompletions prometheus.Counter
 	truncateFail    prometheus.Counter
 	truncateTotal   prometheus.Counter
+	lastLog         prometheus.Gauge
 }
 
 // New returns a new WAL over the given directory.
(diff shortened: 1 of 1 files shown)


cstyan:
@gouthamve @krasi-georgiev 
cc: @fabxc

krasi-georgiev:
I think checkpointing was implemented to solve this problem, no?

krasi-georgiev:
based on the  discussion in https://github.com/prometheus/tsdb/issues/395 I think this is best to instrument in Prometheus itself so that you can also add the metric name sec/milsec etc.

krasi-georgiev:
while you are at it can also add the mint to Prometheus itself so we can close https://github.com/prometheus/tsdb/issues/395 as well.

cstyan:
We don't really have the same issue as #395, Gauge `SetToCurrentTime` is the timestamp in seconds. Either way, I've opened a PR https://github.com/prometheus/prometheus/pull/4750

krasi-georgiev:
will continue tracking in https://github.com/prometheus/prometheus/pull/4750

PR #402, opened 2018-10-05
add a function to head/db to return a list of all series ID's in Head

cstyan:
As part of rewriting the remote write (https://github.com/prometheus/prometheus/pull/4588) I'm keeping track of series records that we read out of the WAL so we can use the reference ID's when reading samples records to build the full metric for sending. However, at the moment that map of reference ID's to labels could grow endlessly. Within TSDB's Head, when truncation it trims series references that are no longer relevant. In the remote write changes, I can detect when a checkpoint is created, and make the call to get the series ID's that this PR adds.

### DIFF
--- a/db.go
+++ b/db.go
@@ -279,6 +279,11 @@ func (db *DB) Dir() string {
 	return db.dir
 }
 
+// AllSeriesIDs returns a slice of all the series ID's currently in Head.
+func (db *DB) AllSeriesIDs() []uint64 {
+	return db.head.allSeriesIDs()
+}
+
 func (db *DB) run() {
 	defer close(db.donec)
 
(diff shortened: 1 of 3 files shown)


cstyan:
@gouthamve @krasi-georgiev 
cc: @fabxc

cstyan:
@fabxc any thoughts here? Not sure if making this data available outside Head in this way is okay with you. Same with #401 

Is any of the work for this: https://docs.google.com/document/d/1TEqqE_Stq04drhjSU1I7Ctmuy0dpsvlPL1AKxqEQoSg/edit?disco=AAAACHODOcM&ts=5b8826c0 going to be open source

Issue #381, opened 2018-09-17
Document and follow release process that is compatible with the go modules.


krasi-georgiev:
In relation to the discussion in the mailing list.

https://groups.google.com/forum/#!topic/prometheus-developers/UMaSwEjasPA

krasi-georgiev:
We have already started doing tags and on top will also maintain a changelog file similar to the on in the Prometheus repo. This should be enough to keep the process lightweight and still document API changes.

PR #406, opened 2018-10-11
Bring back old functionality of Checkpoint()

codesome:
As pointed by @fabxc here https://github.com/prometheus/tsdb/pull/396#discussion_r224167073, `Checkpoint` functionality should not be changed as it is exposed. 

In order to preserve the new metrics added in https://github.com/prometheus/tsdb/pull/396, there are 2 new arguments for `Checkpoint(...)` (can be `nil`).

/cc @krasi-georgiev 

### DIFF
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -23,7 +23,10 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
 	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/tsdb/fileutil"
 	"github.com/prometheus/tsdb/wal"
 )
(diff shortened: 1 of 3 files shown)


brian-brazil:
TSDB is an internal library, we don't make any promises about API stability on those.

fabxc:
> TSDB is an internal library, we don't make any promises about API stability on those.

We never formalized rules but we know that our libraries have external users and that breaking exported APIs causes friction. So I think it's fair to at least require a strong-ish reason to break.

What bothered me in this case was that semantics where changed significantly while code will keep compiling as-is – so undetectable with fatal-ish consequences.
Restoring the old behavior in this PR of course caused another breaking change and the signature is quite a bit more convoluted.

I'm on the fence about it – happy to hear other opinions.

krasi-georgiev:
> What bothered me in this case was that semantics where changed significantly while code will keep compiling as-is – so undetectable with fatal-ish consequences.

my bad, sorry!

in regards to changing exported APIs for packages , what about making a new release when we need to change an api describing the change so that people vendoring the package would at least read the release notes?

simonpasquier:
When I reviewed #396, it crossed my mind that changing the internals of `Checkpoint()` was modifying the API contract but I wasn't sure what is the level of stability promised by `tsbd`. Somehow this relates with #381: if we're better at producing `tsdb` releases, it would both help Prometheus and external consumers of the library.

fabxc:
> my bad, sorry!

No need to apologize – it's a valid refactoring in general, given something had to be broken either way for the new metrics. Maybe it's alright since the API is rather fresh and not super widely used yet.

Proper library releases with tags and notes sound like a good idea. I guess we are not doing that fur libraries in prometheus/prometheus. But arguably the consequences there are nowhere as bad as for TSDB.

krasi-georgiev:
ok , lets leave the current implementation as is and will try with the  releases when we have exposed API changes. 

With the first one will also make a short doc in the wiki(so it is easy to update) for the release process  and once it works well will add it somewhere in the repo as well.

PR #407, opened 2018-10-11
log an error instead of blocking a wal truncation on checkpoint gaps.

krasi-georgiev:
missing wal segment blocks truncation and the wal grows indefinitely.  

An error log message should be enough to indicate a problem.

related to:  https://github.com/prometheus/prometheus/issues/4695

### DIFF
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -112,9 +112,6 @@ func Checkpoint(w *wal.WAL, from, to int, keep func(id uint64) bool, mint int64)
 		}
 		last := idx + 1
 		if err == nil {
-			if from > last {
-				return nil, fmt.Errorf("unexpected gap to last checkpoint. expected:%v, requested:%v", last, from)
-			}
 			// Ignore WAL files below the checkpoint. They shouldn't exist to begin with.
 			from = last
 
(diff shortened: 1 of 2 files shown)


gouthamve:
Hmm, this means we could lose metrics. We need to come up with a policy about when to break and when to potentially lose metrics before we merge this in.

krasi-georgiev:
The other alternative is rebuilding it from the HEADas you suggested. I looked into it, but it is not trivial.

krasi-georgiev:
@fabxc what is your comment for this one?

fabxc:
From the linked issue I read that it is not clear what went wrong in that users setup, right? The problem is that any checkpoint with missing segments since the last checkpoint will potentially or even likely be corrupted.
The missing segments may contain series records and all subsequent samples for that series will fail. One could argue that we can't do anything else anyway and this is thus an acceptable tradeoff. But it also feels a bit wrong to deliberately continue with corrupted data.

It would certainly be great to understand what caused this.
From the user's logs, it appears that this has been an issue at startup already. Maybe a decent intermediate step would be to bail on startup if we detect this issue?

krasi-georgiev:
yes I understand the logic now and agree with you.

Since the original issue was never confirmed I am closing this until we confirm the actual bug if any.

PR #409, opened 2018-10-15
fileutil: remove dependency on prometheus/prometheus

simonpasquier:
`fileutil` tests depended on https://github.com/prometheus/prometheus/util/testutil for creating temporary directories. This change copies the code to the tsdb repository instead of importing it and creating this weird dependency chain. Especially since it isn't used by https://github.com/prometheus/prometheus directly itself.

This is a straight copy of the code but it could even be stripped down as not all exported identifiers are needed by the tsdb code.

cc @krasi-georgiev

### DIFF
--- a/fileutil/flock_test.go
+++ b/fileutil/flock_test.go
@@ -18,7 +18,7 @@ import (
 	"path/filepath"
 	"testing"
 
-	"github.com/prometheus/prometheus/util/testutil"
+	"github.com/prometheus/tsdb/testutil"
 )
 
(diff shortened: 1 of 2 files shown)


krasi-georgiev:
I think @gouthamve  and @fabxc  are both in favour of having no external dependencies so no objections from me.
 if it is an exact copy feel free to merge.

simonpasquier:
Thanks for the quick review! I'll give a bit of time for others to review and merge tomorrow if no one complains.

PR #410, opened 2018-10-15
Add function to get list of all ref ID -> labels from Head

cstyan:
Using this as part of https://github.com/prometheus/prometheus/pull/4588 (remote write via WAL)

 @gouthamve @krasi-georgiev 
cc: @fabxc

### DIFF
--- a/db.go
+++ b/db.go
@@ -279,6 +279,11 @@ func (db *DB) Dir() string {
 	return db.dir
 }
 
+// GetAllLabels returns all the labels currently in the data of the head by series reference ID.
+func (db *DB) GetAllLabels() map[uint64]labels.Labels {
+	return db.head.getAllLabels()
+}
+
 func (db *DB) run() {
 	defer close(db.donec)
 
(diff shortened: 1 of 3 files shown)


krasi-georgiev:
can you summarise the use case to save me reading the comments and PR code? 

btw there is already an interface to get all label values and a pr for the label names https://github.com/prometheus/tsdb/pull/369

cstyan:
@krasi-georgiev Yeah sorry I should have done that already. It's essentially the same as the other PR to get all reference ID's. As far as I can tell, Head is the source of truth for which series references are active and what their labels are. If Prometheus is started from an existing WAL, Head will rebuild the structures containing the series references and labels that it knows about. We can just ask Head for that info rather than rereading the WAL again in the new remote code. The other PR, that just adds a function to get all the series references. I want to use this when the WAL is truncated and a checkpoint is created, since in the new remote write code we're caching the series id's and labels as well.

#369 gets me the labels values, but I also need #402 for the truncate case.

PR #412, opened 2018-10-16
Allow manual compaction for tests when compaction is disabled globally.

krasi-georgiev:
for tests we need to control when a compaction happens so with this
change automated compaction can be disabled, but allow to run it
manually it tests.

fixes failing tests in : https://github.com/prometheus/tsdb/pull/374

cc @gouthamve , @codesome 

### DIFF
--- a/compact_test.go
+++ b/compact_test.go
@@ -19,10 +19,13 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/go-kit/kit/log"
 	"github.com/pkg/errors"
+	dto "github.com/prometheus/client_model/go"
 	"github.com/prometheus/tsdb/chunks"
+	"github.com/prometheus/tsdb/labels"
 	"github.com/prometheus/tsdb/testutil"
 )
 
@@ -681,4 +684,60 @@ func TestCompaction_populateBlock(t *testing.T) {
 			return
 		}
 	}
+}
+
+// TestDisableAutoCompactions checks that we can
+// disable and enable the auto compaction.
+// This is needed for unit tests that rely on
+// checking state before and after a compaction.
+func TestDisableAutoCompactions(t *testing.T) {
+	db, close := openTestDB(t, nil)
+	defer close()
+	defer db.Close()
+
+	blockRange := DefaultOptions.BlockRanges[0]
+	label := labels.FromStrings("foo", "bar")
+
+	// Trigger a compaction to check that it was skipped and
+	// no new blocks were created when compaction is disabled.
+	db.DisableCompactions()
+	app := db.Appender()
+	for i := int64(0); i < 3; i++ {
+		_, err := app.Add(label, i*blockRange, 0)
+		testutil.Ok(t, err)
+		_, err = app.Add(label, i*blockRange+1000, 0)
+		testutil.Ok(t, err)
+	}
+	testutil.Ok(t, app.Commit())
+
+	select {
+	case db.compactc <- struct{}{}:
+	default:
+	}
+
+	m := &dto.Metric{}
+	for x := 0; x < 10; x++ {
+		db.metrics.compactionsSkipped.Write(m)
+		if *m.Counter.Value > float64(0) {
+			break
+		}
+		time.Sleep(10 * time.Millisecond)
+	}
+
+	testutil.Assert(t, *m.Counter.Value > float64(0), "No compaction was skipped after the set timeout.")
+	testutil.Equals(t, 0, len(db.blocks))
+
+	// Enable the compaction, trigger it and check that the block is persisted.
+	db.EnableCompactions()
+	select {
+	case db.compactc <- struct{}{}:
+	default:
+	}
+	for x := 0; x < 10; x++ {
+		if len(db.Blocks()) > 0 {
+			break
+		}
+		time.Sleep(30 * time.Millisecond)
+	}
+	testutil.Assert(t, len(db.Blocks()) > 0, "No block was persisted after the set timeout.")
 }
--- a/db.go
+++ b/db.go
@@ -113,8 +113,8 @@ type DB struct {
 	stopc    chan struct{}
 
 	// cmtx is used to control compactions and deletions.
-	cmtx               sync.Mutex
-	compactionsEnabled bool
+	cmtx            sync.Mutex
+	autoCompactions bool
 }
 
 type dbMetrics struct {
@@ -123,6 +123,7 @@ type dbMetrics struct {
 	reloads              prometheus.Counter
 	reloadsFailed        prometheus.Counter
 	compactionsTriggered prometheus.Counter
+	compactionsSkipped   prometheus.Counter
 	cutoffs              prometheus.Counter
 	cutoffsFailed        prometheus.Counter
 	startTime            prometheus.GaugeFunc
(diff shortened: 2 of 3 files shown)


db.go
```diff
@@ -298,6 +298,10 @@ func (db *DB) run() {
 			default:
 			}
 		case <-db.compactc:
+			if !db.compactionsEnabled {
+				continue
+			}
+
 			db.metrics.compactionsTriggered.Inc()
```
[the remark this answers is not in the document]
krasi-georgiev: hm , this doesn't make sense , why would we increment it if the compaction is disabled. probably this was a bug from before.
@gouthamve can you think of any reason for this behaviour?

db.go
```diff
@@ -298,6 +298,10 @@ func (db *DB) run() {
 			default:
 			}
 		case <-db.compactc:
+			if !db.compactionsEnabled {
```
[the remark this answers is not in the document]
krasi-georgiev: aah , nice catch. I don't think we even need a lock in `.compact()` , right?

and I can add another func like
```
if !db.compactionsEnabled() {
	continue
}
func ... compactionsEnabled(){
    db.cmtx.Lock()
    defer db.cmtx.Unlock()
    return db.compactionsEnabled
}
```

db.go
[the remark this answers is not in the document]
codesome: Yup, probably a bug.

db.go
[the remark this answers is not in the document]
codesome: If we consider : `db.compactionsEnabled()` -> `db.DisableCompactions()` -> `db.compact()`, would that still be a race?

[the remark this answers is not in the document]
codesome: It is also used in Delete, so we need to lock it in `compact()`.

krasi-georgiev:
updated

db.go
```diff
@@ -298,6 +298,10 @@ func (db *DB) run() {
 			default:
 			}
 		case <-db.compactc:
+			if !db.compactionsIsEnabled() {
```
codesome: If we consider : `db.compactionsEnabled()` -> `db.DisableCompactions()` -> `db.compact()`, would that still be a race?

db.go
```diff
@@ -367,13 +371,6 @@ func (a dbAppender) Commit() error {
 // Old blocks are only deleted on reload based on the new block's parent information.
 // See DB.reload documentation for further information.
 func (db *DB) compact() (err error) {
-	db.cmtx.Lock()
```
codesome: `db.cmtx` is also used in Delete, we should not remove it from here. Comments say `cmtx is used to control compactions and deletions.`.

db.go
krasi-georgiev: not sure what you mean , each one of these has a mutex  lock

codesome: After successful `DisableCompactions` call, we should not be doing any compactions. So a `compact` call after disabling compactions doesn't sound right. Though we can logically assume that `compact` happened before in this case.

db.go
krasi-georgiev: yes I see it now, these shouldn't happen at the same time.

db_test.go
```diff
@@ -1337,3 +1338,50 @@ func TestCorrectNumTombstones(t *testing.T) {
...
+// TestDisableCompactions checks that we can disable the
+// auto compaction and run compactions on demand.
+// This is needed for unit tests that rely on
+// checking state before and after a compaction.
+func TestDisableAutoCompactions(t *testing.T) {
+	db, close := openTestDB(t, nil)
+	defer close()
+	defer db.Close()
+
+	db.DisableCompactions()
```
codesome: For more comprehensive test, should we also check
* Auto compaction is happening before disabling.
* Enable auto compactions again at the end and check if auto compactions are happening.

(Enabled -> Disabled -> Enabled)

codesome:
ping @gouthamve. With this change, you wont be able to disable compactions altogether, but only disable auto compactions. But given that `compact()` is not exposed, and it is not called anywhere other than during auto compactions, it will effectively behave the same.

db.go
```diff
@@ -731,21 +739,21 @@ func (db *DB) Close() error {
 	return merr.Err()
 }
 
-// DisableCompactions disables compactions.
+// DisableCompactions disables auto compactions.
 func (db *DB) DisableCompactions() {
```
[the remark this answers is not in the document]
krasi-georgiev: yes lets keep it as is for now and will change if/when necessary.
There is no notion for manual compaction in the api so don't think there will be any confusion.

db.go
```diff
@@ -112,9 +112,13 @@ type DB struct {
...
-	// cmtx is used to control compactions and deletions.
-	cmtx               sync.Mutex
-	compactionsEnabled bool
+	// cmtx ensures that compactions and deletions don't run simultaneously.
+	cmtx sync.Mutex
+
+	// autoCompactMtx ensures that no compaction gets triggered while
+	// changing the autoCompact var.
+	autoCompactMtx sync.Mutex
+	autoCompact    bool
```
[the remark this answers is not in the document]
krasi-georgiev: I think it is clear enough, but if you insist I don't mind changing it.

db_test.go
krasi-georgiev: I refactored it a bit to better cover `Disabled -> Enabled` and unless you can think of some immediate bug that that is not covered by this logic. I think it should be ok.

db.go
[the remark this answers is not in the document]
codesome: I think `autoCompact` is fine.

db_test.go
```diff
@@ -1337,3 +1338,50 @@ func TestCorrectNumTombstones(t *testing.T) {
...
+	// check that no new blocks were created.
+	m := &dto.Metric{}
+	for x := 0; x < 3; x++ {
+		db.metrics.compactionsSkipped.Write(m)
+		if *m.Counter.Value > float64(0) {
+			break
+		}
+		time.Sleep(10 * time.Millisecond)
+	}
+	if *m.Counter.Value == float64(0) {
```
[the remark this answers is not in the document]
codesome: It *may* have been possible that compaction didn't happen while it looped 3 times above. The new addition should work.

db.go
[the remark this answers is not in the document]
krasi-georgiev: I think I understand why it was there. Most likely to check that the logic for when to trigger a compaction is working as expected. Regardless of whether a compaction actually runs.

krasi-georgiev:
@fabxc 	, @gouthamve do you 	mind have a quick look as it is blocking another bug fix in https://github.com/prometheus/tsdb/pull/374

db.go
```diff
@@ -112,9 +112,13 @@ type DB struct {
...
 
-	// cmtx is used to control compactions and deletions.
-	cmtx               sync.Mutex
-	compactionsEnabled bool
+	// cmtx ensures that compactions and deletions don't run simultaneously.
+	cmtx sync.Mutex
+
+	// autoCompactMtx ensures that no compaction gets triggered while
+	// changing the autoCompact var.
+	autoCompactMtx sync.Mutex
```
gouthamve: Do we really need another mutex for this? Can we not use `cmtx` itself?

krasi-georgiev: Yeah I think blocking is the right way.
Imagine a processes that relies on the fact that after `DisableCompaction()`  no data changes will happen which won't be the case if we don't block. Otherwise if the process doesn't care than can always run `DisableCompaction()` in the background to continue the other executions. 
So in other words blocking leaves that decision to the caller which is always a good thing.

compact_test.go
```diff
@@ -682,3 +685,59 @@ func TestCompaction_populateBlock(t *testing.T) {
...
+	}
+	testutil.Ok(t, app.Commit())
+
+	select {
+	case db.compactc <- struct{}{}:
+	default:
+	}
+
+	m := &dto.Metric{}
+	for x := 0; x < 10; x++ {
```
gouthamve: Can we simplify this by just sleeping for `100ms`?

compact_test.go
```diff
@@ -682,3 +685,59 @@ func TestCompaction_populateBlock(t *testing.T) {
...
+	testutil.Equals(t, 0, len(db.blocks))
+
+	// Enable the compaction, trigger it and check that the block is persisted.
+	db.EnableCompactions()
+	select {
+	case db.compactc <- struct{}{}:
+	default:
+	}
+	for x := 0; x < 10; x++ {
+		if len(db.Blocks()) > 0 {
```
gouthamve: Same here?

compact_test.go
krasi-georgiev: yeah, but that could complete a lot faster depending on the hardware where it runs so I thought this is the best way to make the test run as short as possible. Not a big deal, but every little helps right :question:  :smile:

krasi-georgiev: I am merging as is and will revisit about this one if needed.

PR #415, opened 2018-10-17
Delete temp checkpoint folder on error.

krasi-georgiev:


krasi-georgiev:
cc @codesome

block.go
```diff
@@ -212,7 +213,7 @@ func writeMetaFile(dir string, meta *BlockMeta) error {
 
 	// Make any changes to the file appear atomic.
 	path := filepath.Join(dir, metaFilename)
-	tmp := path + ".tmp"
+	tmp := path + fileutil.TmpExt
```
codesome: Instead of manually appending `fileutil.TmpExt` everywhere, would it be better to have a function call like `fileutil.TmpDir` which would take the `dir` and append `fileutil.TmpExt` to it? In case we want to change the logic of naming temp dir, this would come handy (example `ioutil.TempDir`).
```go
package fileutil

func TmpDir(dir string) string {
    return dir + TmpExt
}
```
```go
tmp := fileutil.TmpDir(path)
```

`fileutil.TmpExt` would still be present if it is needed somewhere.

krasi-georgiev: yeah I like the idea, but I prefer to be more explicit and `tmp := fileutil.TmpDir(path)` might be a bit strange. What does it do?  create a tmp dir? copies to a tmp dir ? 
refactoring is easy enough so lets keep as is and will change if needed.

checkpoint.go
```diff
@@ -141,12 +141,15 @@ func Checkpoint(w *wal.WAL, from, to int, keep func(id uint64) bool, mint int64)
...
-	cpdir := filepath.Join(w.Dir(), fmt.Sprintf("checkpoint.%06d", to))
-	cpdirtmp := cpdir + ".tmp"
+	cpdir := filepath.Join(w.Dir(), fmt.Sprintf(checkpointPrefix+"%06d", to))
+	cpdirtmp := cpdir + fileutil.TmpExt
 
 	if err := os.MkdirAll(cpdirtmp, 0777); err != nil {
 		return nil, errors.Wrap(err, "create checkpoint dir")
 	}
+	// Ensure that an early return doesn't leave any tmp folder.
+	defer os.RemoveAll(cpdirtmp)
```
gouthamve: Pretty sure this will break in Windows (atleast for the test given below). When we open the `wal` below, the chances are there are open references to the files inside the directory that we are not closing. Don't have access to windows machine now to test this though.


Also don't have a good idea as to how to fix this.

krasi-georgiev:
after having another look it is a terrible solution and very likely to brake but will keep it open as a reminder to think how ti fix properly for linux and windows.

krasi-georgiev:
Refactored with a different approach. `wal.Close()` is now non blocking so it we just call it again and remove all tmp folder.

krasi-georgiev:
cc @codesome

codesome:
LGTM

PR #417, opened 2018-10-17
Reduce allocations for queries on `HEAD`

jacksontj:
This was growing the size of the `[]string` while iterating over, which in practive ends up being multiple allocations. Since we know the max number that could be there, we should instead right-size allocate the slice. In my local testing, here are the results (from the benchmark included):

```
$ benchcmp baseline after
benchmark                             old ns/op     new ns/op     delta
BenchmarkHeadPostingForMatchers-4     74452         63514         -14.69%

benchmark                             old allocs     new allocs     delta
BenchmarkHeadPostingForMatchers-4     20             13             -35.00%

benchmark                             old bytes     new bytes     delta
BenchmarkHeadPostingForMatchers-4     5425          3137          -42.18%
```

### DIFF
--- a/head_bench_test.go
+++ b/head_bench_test.go
@@ -1,0 +1,59 @@
+package tsdb
+
+import (
+	"strconv"
+	"sync/atomic"
+	"testing"
+
+	"github.com/prometheus/tsdb/labels"
+	"github.com/prometheus/tsdb/testutil"
+)
+
+func BenchmarkHeadStripeSeriesCreate(b *testing.B) {
+	// Put a series, select it. GC it and then access it.
+	h, err := NewHead(nil, nil, nil, 1000)
+	testutil.Ok(b, err)
+	defer h.Close()
+
+	for i := 0; i < b.N; i++ {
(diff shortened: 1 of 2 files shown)


jacksontj:
@gouthamve updated!

head_bench_test.go
```diff
@@ -0,0 +1,61 @@
...
+	h, err := NewHead(nil, nil, nil, 1000)
+	testutil.Ok(b, err)
+	defer h.Close()
+
+	var count int64
+
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			i := atomic.AddInt64(&count, 1)
+			s, _ = h.getOrCreate(uint64(i), labels.FromStrings("a", strconv.Itoa(int(i))))
```
krasi-georgiev: this one has races when run in parallel.

btw why it is here since the PR changes seem unrelated to this benchmark.

krasi-georgiev:
Thanks @jacksontj , appreciated :+1:

PR #419, opened 2018-10-21
refactor util funcs to allow re-usage.

krasi-georgiev:
I am trying to organise the test and helper funcs  to that the tests end up in the correct locations and also can reuse some helper funcs.

This will allow the `testeMergeChunk` from https://github.com/prometheus/tsdb/pull/397 to be put in `chunks_test.go` instead of keeping  it in `compact_test.go`

https://github.com/prometheus/tsdb/pull/397/files#diff-ac389a623f099fc727fc1f3edca6853fR764

cc @csmarchbanks

### DIFF
--- a/tsdbutil/buffer.go
+++ b/tsdbutil/buffer.go
@@ -2,13 +2,25 @@ package tsdbutil
 
 import (
 	"math"
-
-	"github.com/prometheus/tsdb"
 )
+
+// SeriesIterator iterates over the data of a time series.
+type SeriesIterator interface {
+	// Seek advances the iterator forward to the given timestamp.
+	// If there's no value exactly at t, it advances to the first value
+	// after t.
+	Seek(t int64) bool
+	// At returns the current timestamp/value pair.
+	At() (t int64, v float64)
+	// Next advances the iterator by one.
+	Next() bool
+	// Err returns the current error.
+	Err() error
+}
 
 // BufferedSeriesIterator wraps an iterator with a look-back buffer.
 type BufferedSeriesIterator struct {
-	it  tsdb.SeriesIterator
+	it  SeriesIterator
 	buf *sampleRing
 
 	lastTime int64
@@ -16,7 +28,7 @@ type BufferedSeriesIterator struct {
 
 // NewBuffer returns a new iterator that buffers the values within the time range
 // of the current element and the duration of delta before.
-func NewBuffer(it tsdb.SeriesIterator, delta int64) *BufferedSeriesIterator {
+func NewBuffer(it SeriesIterator, delta int64) *BufferedSeriesIterator {
 	return &BufferedSeriesIterator{
 		it:       it,
 		buf:      newSampleRing(delta, 16),
@@ -31,7 +43,7 @@ func (b *BufferedSeriesIterator) PeekBack() (t int64, v float64, ok bool) {
 }
 
 // Buffer returns an iterator over the buffered data.
-func (b *BufferedSeriesIterator) Buffer() tsdb.SeriesIterator {
+func (b *BufferedSeriesIterator) Buffer() SeriesIterator {
 	return b.buf.iterator()
 }
 
@@ -90,6 +102,14 @@ type sample struct {
 	v float64
 }
 
+func (s sample) T() int64 {
+	return s.t
+}
+
+func (s sample) V() float64 {
+	return s.v
+}
+
 type sampleRing struct {
 	delta int64
 
@@ -112,7 +132,7 @@ func (r *sampleRing) reset() {
 	r.f = 0
 }
 
-func (r *sampleRing) iterator() tsdb.SeriesIterator {
+func (r *sampleRing) iterator() SeriesIterator {
 	return &sampleRingIterator{r: r, i: -1}
 }
 
--- a/db_test.go
+++ b/db_test.go
@@ -242,7 +242,7 @@ Outer:
 		res, err := q.Select(labels.NewEqualMatcher("a", "b"))
 		testutil.Ok(t, err)
 
-		expSamples := make([]Sampler, 0, len(c.remaint))
+		expSamples := make([]Sample, 0, len(c.remaint))
 		for _, ts := range c.remaint {
 			expSamples = append(expSamples, sample{ts, smpls[ts]})
 		}
@@ -469,7 +469,7 @@ Outer:
 		res, err := q.Select(labels.NewEqualMatcher("a", "b"))
 		testutil.Ok(t, err)
 
-		expSamples := make([]Sampler, 0, len(c.remaint))
+		expSamples := make([]Sample, 0, len(c.remaint))
 		for _, ts := range c.remaint {
 			expSamples = append(expSamples, sample{ts, smpls[ts]})
 		}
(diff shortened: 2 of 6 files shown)


tsdbutil/buffer.go
```diff
@@ -2,21 +2,33 @@ package tsdbutil
...
+	it  SeriesIterator
 	buf *sampleRing
 
 	lastTime int64
 }
 
 // NewBuffer returns a new iterator that buffers the values within the time range
 // of the current element and the duration of delta before.
-func NewBuffer(it tsdb.SeriesIterator, delta int64) *BufferedSeriesIterator {
+func NewBuffer(it SeriesIterator, delta int64) *BufferedSeriesIterator {
```
krasi-georgiev: @fabxc this is an API change so I am thinking to try the versioning with this one.

This is a not breakin change so should bump the minor version. Since we don't have any versioning shall we tag this as 2.0.1 ?

this change is needed to avoid dependancy import cycles with `tsdbutil/chunks.go`.

fabxc:
Moving core interface types to a `util` package is problematic IMO – it's now no longer a helper package but describes the core data model.

Why `sample` to `Sampler`? The trailing `r` doesn't seem to make sense.
It also adds one layer of indirection to every single lookup of timestamp and value in all places it is used. Did you check any potential performance impact?

fabxc:
Moving core interface types to a `util` package is problematic IMO – it's now no longer a helper package but describes the core data model.

Why `sample` to `Sampler`? The trailing `r` doesn't seem to make sense.
It also adds one layer of indirection to every single lookup of timestamp and value in all places it is used. Did you check any potential performance impact?

krasi-georgiev:
@fabxc the change is only to facilitate testing and allow re-usage of  `ChunkFromSamples(s []Sampler) chunks.Meta` and `PopulatedChunk(numSamples int, minTime int64)` , and my plan is to move other test helper funcs to `tsdbutil` as well.

because `type sample` is unexplored I had to add the `Sampler` interface so that for example a test in `chunks/chunks_test.go` can reuse these helper funcs. I tried few different ways and this was the only one that required minimal changes.

> Moving core interface types to a util package is problematic IMO – it's now no longer a helper package but describes the core data model.

can you add a bit more info what you mean here. I could try to  put these test helper funcs in `./internal` to make sure these are used only within tsdb.

krasi-georgiev:
@fabxc the change is only to facilitate testing and allow re-usage of  `ChunkFromSamples(s []Sampler) chunks.Meta` and `PopulatedChunk(numSamples int, minTime int64)` , and my plan is to move other test helper funcs to `tsdbutil` as well.

because `type sample` is unexplored I had to add the `Sampler` interface so that for example a test in `chunks/chunks_test.go` can reuse these helper funcs. I tried few different ways and this was the only one that required minimal changes.

> Moving core interface types to a util package is problematic IMO – it's now no longer a helper package but describes the core data model.

can you add a bit more info what you mean here. I could try to  put these test helper funcs in `./internal` to make sure these are used only within tsdb.

krasi-georgiev:
@fabxc the change is only to facilitate testing and allow re-usage of  `ChunkFromSamples(s []Sampler) chunks.Meta` and `PopulatedChunk(numSamples int, minTime int64)` , and my plan is to move other test helper funcs to `tsdbutil` as well.

because `type sample` is unexplored I had to add the `Sampler` interface so that for example a test in `chunks/chunks_test.go` can reuse these helper funcs. I tried few different ways and this was the only one that required minimal changes.

> Moving core interface types to a util package is problematic IMO – it's now no longer a helper package but describes the core data model.

can you add a bit more info what you mean here. I could try to  put these test helper funcs in `./internal` to make sure these are used only within tsdb.

krasi-georgiev:
@fabxc the change is only to facilitate testing and allow re-usage of  `ChunkFromSamples(s []Sampler) chunks.Meta` and `PopulatedChunk(numSamples int, minTime int64)` , and my plan is to move other test helper funcs to `tsdbutil` as well so that we can place tests for subpackages in the corresponding folders and still be able to reuse these helper funcs.

because `type sample` is unexplored I had to add the `Sampler` interface so that for example a test in `chunks/chunks_test.go` can reuse these helper funcs. I tried few different ways and this was the only one that required minimal changes.

> Moving core interface types to a util package is problematic IMO – it's now no longer a helper package but describes the core data model.

can you add a bit more info what you mean here. I could try to  put these test helper funcs in `./internal` to make sure these are used only within tsdb.

krasi-georgiev:
@fabxc the change is only to facilitate testing and allow re-usage of  `ChunkFromSamples(s []Sampler) chunks.Meta` and `PopulatedChunk(numSamples int, minTime int64)` , and my plan is to move other test helper funcs to `tsdbutil` as well so that we can place tests for subpackages in the corresponding folders and still be able to reuse these helper funcs.

because `type sample` is unexplored I had to add the `Sampler` interface so that for example a test in `chunks/chunks_test.go` can reuse these helper funcs. I tried few different ways and this was the only one that required minimal changes.

> Moving core interface types to a util package is problematic IMO – it's now no longer a helper package but describes the core data model.

can you add a bit more info what you mean here. I could try to  put these test helper funcs in `./internal` to make sure these are used only within tsdb.

krasi-georgiev:
@fabxc the change is only to facilitate testing and allow re-usage of  `ChunkFromSamples(s []Sampler) chunks.Meta` and `PopulatedChunk(numSamples int, minTime int64)` , and my plan is to move other test helper funcs to `tsdbutil` as well so that we can place tests for subpackages in the corresponding folders and still be able to reuse these helper funcs.

because `type sample` is unexplored I had to add the `Sampler` interface so that for example a test in `chunks/chunks_test.go` can reuse these helper funcs. I tried few different ways and this was the only one that required minimal changes.

> Moving core interface types to a util package is problematic IMO – it's now no longer a helper package but describes the core data model.

can you add a bit more info what you mean here. I could try to  put these test helper funcs in `./internal` to make sure these are used only within tsdb.

krasi-georgiev:
@fabxc the change is only to facilitate testing and allow re-usage of  `ChunkFromSamples(s []Sampler) chunks.Meta` and `PopulatedChunk(numSamples int, minTime int64)` , and my plan is to move other test helper funcs to `tsdbutil` as well so that we can place tests for subpackages in the corresponding folders and still be able to reuse these helper funcs.

because `type sample` is unexplored I had to add the `Sampler` interface so that for example a test in `chunks/chunks_test.go` can reuse these helper funcs. I tried few different ways and this was the only one that required minimal changes.

> Moving core interface types to a util package is problematic IMO – it's now no longer a helper package but describes the core data model.

can you add a bit more info what you mean here. I could try to  put these test helper funcs in `./internal` to make sure these are used only within tsdb.

fabxc:
> can you add a bit more info what you mean here. I could try to put these test helper funcs in ./internal to make sure these are used only within tsdb.

Looks like you just duplicated the SeriesIterator interface. I misread that you had moved it. So all good I think.

I'd still just name the interface `Sample` maybe. The `r` suffixing like in`http.Handler` or `io.Reader` mostly seems to make sense for things that "do stuff". This just has two distinct getters here and `Sampler` makes me think the underlying object can produce samples for something.

krasi-georgiev:
@fabxc updated, 
I wanted to show that is it an interface , but not that important. I think this change it is a step in the right direction so can optimise as we go.

PR #414, opened 2018-10-16
Fix slice index out of bounds panic

AudriusButkevicius:
r.buf is used for header and the page, if the actual page size is near pageSize buf[:k] goes out of bounds.
Pr produced on a mobile via the web editor, so beware of the formatting issues.

```

goroutine 167 [running]:
github.com/prometheus/prometheus/vendor/github.com/prometheus/tsdb/wal.(*Reader).next(0xc0074e0000, 0xc0135f2000, 0xc0138d3470)
        /go/src/github.com/prometheus/prometheus/vendor/github.com/prometheus/tsdb/wal/wal.go:739 +0x9a6
github.com/prometheus/prometheus/vendor/github.com/prometheus/tsdb/wal.(*Reader).Next(0xc0074e0000, 0x94416)
        /go/src/github.com/prometheus/prometheus/vendor/github.com/prometheus/tsdb/wal/wal.go:706 +0x2f
github.com/prometheus/prometheus/vendor/github.com/prometheus/tsdb.(*Head).loadWAL(0xc0001c0fd0, 0xc0074e0000, 0xc0071b1640, 0x1d1e5a0)
        /go/src/github.com/prometheus/prometheus/vendor/github.com/prometheus/tsdb/head.go:304 +0x266
github.com/prometheus/prometheus/vendor/github.com/prometheus/tsdb.(*Head).Init(0xc0001c0fd0, 0x0, 0x0)
        /go/src/github.com/prometheus/prometheus/vendor/github.com/prometheus/tsdb/head.go:411 +0x28d
github.com/prometheus/prometheus/vendor/github.com/prometheus/tsdb.Open(0x7fff9c72ff70, 0xb, 0x1d1d600, 0xc0002194d0, 0x1d30ec0, 0xc00018c3c0, 0xc000219500, 0xc00075ce80, 0x0, 0x0)
        /go/src/github.com/prometheus/prometheus/vendor/github.com/prometheus/tsdb/db.go:268 +0x55a
github.com/prometheus/prometheus/storage/tsdb.Open(0x7fff9c72ff70, 0xb, 0x1d1d600, 0xc0002194d0, 0x1d30ec0, 0xc00018c3c0, 0xc0003e19b8, 0x0, 0x0, 0x0)
        /go/src/github.com/prometheus/prometheus/storage/tsdb/tsdb.go:143 +0x285
main.main.func20(0x0, 0x0)
        /go/src/github.com/prometheus/prometheus/cmd/prometheus/main.go:555 +0x1e6
[...]

PR #429, opened 2018-10-24
buffer-panic when reading a record after recPageTerm

krasi-georgiev:
fixes: #414  

cc @gouthamve 	, @fabxc 

I want to have a look at it tomorrow again with fresh eyes, but I was able to replicate the panic in a test and the change fixes the panic.

### DIFF
--- a/wal/wal.go
+++ b/wal/wal.go
@@ -92,8 +92,9 @@ func (e *CorruptionErr) Error() string {
 }
 
 // OpenWriteSegment opens segment k in dir. The returned segment is ready for new appends.
-func OpenWriteSegment(dir string, k int) (*Segment, error) {
-	f, err := os.OpenFile(SegmentName(dir, k), os.O_WRONLY|os.O_APPEND, 0666)
+func OpenWriteSegment(logger log.Logger, dir string, k int) (*Segment, error) {
+	segName := SegmentName(dir, k)
+	f, err := os.OpenFile(segName, os.O_WRONLY|os.O_APPEND, 0666)
 	if err != nil {
 		return nil, err
 	}
@@ -108,6 +109,7 @@ func OpenWriteSegment(dir string, k int) (*Segment, error) {
 	// If it was torn mid-record, a full read (which the caller should do anyway
 	// to ensure integrity) will detect it as a corruption by the end.
 	if d := stat.Size() % pageSize; d != 0 {
+		level.Warn(logger).Log("msg", "last page of the wal is torn, filling it with zeros", "segment", segName)
 		if _, err := f.Write(make([]byte, pageSize-d)); err != nil {
 			f.Close()
 			return nil, errors.Wrap(err, "zero-pad torn page")
@@ -225,7 +227,7 @@ func NewSize(logger log.Logger, reg prometheus.Registerer, dir string, segmentSi
 			return nil, err
 		}
 	} else {
-		if w.segment, err = OpenWriteSegment(w.dir, j); err != nil {
+		if w.segment, err = OpenWriteSegment(w.logger, w.dir, j); err != nil {
 			return nil, err
 		}
 		// Correctly initialize donePages.
--- a/wal/wal.go
+++ b/wal/wal.go
@@ -745,6 +745,9 @@ func (r *Reader) next() (err error) {
 
 		// Gobble up zero bytes.
 		if typ == recPageTerm {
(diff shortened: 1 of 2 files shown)


wal/wal.go
```diff
@@ -745,6 +745,9 @@ func (r *Reader) next() (err error) {
 
 		// Gobble up zero bytes.
 		if typ == recPageTerm {
+			// Increase buf to be able to fit an entyre page size minus 1 byte for the rec type.
+			buf = r.buf[1:]
```
fabxc: I have a hard time connecting the dots between this and the linked pull request.
The latter seems to say that we write records of `pageSize` bytes but add the header on top afterwards, which makes the read buffer too small. That's not the case, headers are part of the page, see [here](https://github.com/prometheus/tsdb/blob/master/wal/wal.go#L473).

Could you confirm that this is the actual case the user hit? I'd be curious to understand how it could happen outside of data corruption since we only pad a page with zero bytes (of which `recPageTerm` is the first one), if we can't even fit a header into the page anymore.

krasi-georgiev: I have been staring at the code few hours and can't see how can this happen apart from  data corruption.
Will keep digging and will let you know if I find anything else.

krasi-georgiev: in any case shouldn't the `buf` be able to fit an entire page after a `recPageTerm` ?

fabxc: Yes, your fix is correct. I'm mostly wondering whether this can happen in the non-corruption case.
I had to go back to the code to understand why we need this though. Could we change the comment to:

```
recPageTerm is a single byte that indicates that the rest of the page is padded. If it's the first byte in a page buf is to small and we have to resize buf to fit pageSize-1 bytes.
```

krasi-georgiev: thanks, updated.
Since @AudriusButkevicius won't be available for a while to confirm if this was caused by a data corruption maybe merge this one and will keep his PR open to wait for his reply?

I also opened another PR with to add some more comments if you want to have a look at that one while the topic is fresh.
https://github.com/prometheus/tsdb/pull/430

AudriusButkevicius:
It's possible that the issue was caused by corruption, as the instance in question on quobyte.

krasi-georgiev:
@AudriusButkevicius any chance to send me the wal files (data/wal/*) privately at kgeorgie at redhat.com?

I just had a look at the code again and can't see how a `recPageTerm` rec type can end up written on disk so having the wal would help to confirm if it was indeed corruption.

krasi-georgiev:
partial write or incorrect rec length could be another cause, but this again would be a result of some incorrect file write.

AudriusButkevicius:
So I've seen this again, and ops folks just nuked the wals and restarted like we did last time. Sadly I cannot provide wals due to work policy, but as I said this is quobyte volume on k8s, so I guess corruption is more probable than usual, yet still index error is unusual as it implies a bug, I'd expect a corruption error, with a potential way to recover ideally.

krasi-georgiev:
thanks to https://github.com/prometheus/tsdb/pull/442 see [my comment](https://github.com/prometheus/tsdb/pull/442#issuecomment-436961364). There is  another way to replicate this. Basically if you share the wal dir with another Prometheus and if the  second instance starts before the first closes the wal, the second instance will detect a corruption and will zero pad the wal. 
Usually this should be detected by the data dir locking, but maybe this behaves different with Quobyte. 
I have added an additional log so you can compile Prometheus with this fix and the additional log to confirm my theory. 

cc @fabxc

AudriusButkevicius:
I think we use persistent volume claims, so I suspect it would be hard to have two instances running at once. I guess next time I see corruption I can just see if there are a slur of zeroes somewhere.

AudriusButkevicius:
Abrupt termination is very possible in the scenario I am seeing, we have k8s machines moving around and going up and down all the time, yet I still don't see a reason it should panic with an index error upon starting after abrupt termination. It should potentially discard the corrupt wal log/record and get on with its life. 

Sure, it's data loss, but you cannot recover that data, and you had data loss anyway if the service was down.

Hence I wonder why this is still not merged?

krasi-georgiev:
Waiting for at least one LGTM from another maintainer.


I have opened another PR that addresses the root cause 
https://github.com/prometheus/tsdb/pull/451

ping @fabxc

krasi-georgiev:
merging since Fabian confirmed it is a valid fix in the comment above and no need to postpone it more.

PR #433, opened 2018-10-29
FromMap(), sorts and returns instead of calling New()

nilsocket:
FromMap(), calls New(), which creates slice once again (which isn't necessary).
as we can do the same in FromMap() itself.

Signed-off-by: nilsocket <nilsocket@gmail.com>


krasi-georgiev:
lgtm , thanks

cc @gouthamve

PR #435, opened 2018-10-29
loop from the end, as readDir returns sorted output

nilsocket:
`ReadDir()` returns sorted output, and the next sequence number is going to be from last.

question: Is it better to return, this:
```go
return filepath.Join(dir, "000001"), 1, nil
```

Signed-off-by: nilsocket <nilsocket@gmail.com>

### DIFF
--- a/chunks/chunks.go
+++ b/chunks/chunks.go
@@ -372,20 +372,21 @@ func (s *Reader) Chunk(ref uint64) (chunkenc.Chunk, error) {
 }
 
 func nextSequenceFile(dir string) (string, int, error) {
-	names, err := fileutil.ReadDir(dir)
+	names, err := fileutil.ReadDir(dir) //sorted
 	if err != nil {
 		return "", 0, err
 	}
 
-	i := uint64(0)
-	for _, n := range names {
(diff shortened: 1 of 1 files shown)


krasi-georgiev:
we have this one duplicated in `db.go` as well.
how about moving both to `fileutil/fileutil.go` and adding a test to check for correct behaviour. 

For example using the `LogSamples` and `WriteChunks` to check that a new sequenced file is created.

krasi-georgiev:
sure.

You can use some notes we have in:
https://docs.google.com/document/d/1t5eSLlH36oK1Ex3PWwuc4R58kx4PC7Ago90TRR_7nmo

and maybe add more or suggest changes.

also look at:
https://github.com/prometheus/tsdb/tree/master/docs/format

nilsocket:
Thanks for the doc, why isn't `wal.md` indexed in `format/README.md`.
I read the format.

Thanks.

krasi-georgiev:
just merged a PR to add it there :smile:

krasi-georgiev:
@nilsocket 	seems stale, feel free to reopen if you decide to continue working on this.

Issue #434, opened 2018-10-29
add and fix the tests for `check_license` and `staticcheck`


krasi-georgiev:
currently in travis these are not enabled as the tests fail due to some valid problems.

ideally this should be fixed and enabled in the tests.

https://github.com/prometheus/tsdb/blob/d804a27062fc524a7494592b72cb23cae1f709cc/.travis.yml#L15-L18

PR #439, opened 2018-10-31
fix lint errors

krasi-georgiev:
unexported NewMemTombstones as this was probably exported by mistake since it returns unexported memTombstones type which will not be shown in godoc.
Added missing comments for exported methods.
Removed unused RecordLogger,RecordReader interfaces.

cc @fabxc 	, @gouthamve

### DIFF
--- a/block.go
+++ b/block.go
@@ -457,7 +457,7 @@ func (pb *Block) Delete(mint, maxt int64, ms ...labels.Matcher) error {
 	ir := pb.indexr
 
 	// Choose only valid postings which have chunks in the time-range.
-	stones := NewMemTombstones()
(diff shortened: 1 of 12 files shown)


codesome:
Quoting https://github.com/prometheus/tsdb/issues/434, running `make staticcheck` manually will give some more places where it can be fixed.

krasi-georgiev:
sorry I didn't get what you mean?

krasi-georgiev:
no this fixes some of the error reported by `golint ./...`

these were reported when browsing around with vscode and now when I run the tool manually it found few more, but I will probably fix this in a separate PR.

PR #440, opened 2018-10-31
Improve WAL reading

brian-brazil:
Reading a 1B sample WAL before:
1630.86user 17.04system 3:43.40elapsed 737%CPU (0avgtext+0avgdata 5897432maxresident)k
23424957inputs+90outputs (1031major+2039558minor)pagefaults 0swaps

After:
383.87user 10.86system 1:00.87elapsed 648%CPU (0avgtext+0avgdata 3766796maxresident)k
23304589inputs+54outputs (6major+1150971minor)pagefaults 0swaps

So that's about 3.75x faster wall time, and 4.25x less CPU.

I also tried putting the WAL reads and decodes in separate goroutines, which didn't help. Nor did avoiding the NumSamples calls.

### DIFF
--- a/head.go
+++ b/head.go
@@ -237,7 +237,6 @@ func NewHead(r prometheus.Registerer, l log.Logger, wal *wal.WAL, chunkRange int
 // Samples before the mint timestamp are discarded.
 func (h *Head) processWALSamples(
 	minValidTime int64,
-	partition, total uint64,
 	input <-chan []RefSample, output chan<- []RefSample,
 ) (unknownRefs uint64) {
 	defer close(output)
@@ -249,7 +248,7 @@ func (h *Head) processWALSamples(
 
 	for samples := range input {
 		for _, s := range samples {
-			if s.T < minValidTime || s.Ref%total != partition {
+			if s.T < minValidTime {
 				continue
 			}
 			ms := refSeries[s.Ref]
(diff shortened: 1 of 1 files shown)


brian-brazil:
Found another useful optimisation.

Before:

1708.38user 17.95system 4:05.89elapsed 702%CPU (0avgtext+0avgdata 5887176maxresident)k
23526309inputs+96outputs (831major+2021538minor)pagefaults 0swaps

After:

261.94user 11.54system 1:01.40elapsed 445%CPU (0avgtext+0avgdata 3548828maxresident)k
23106453inputs+60outputs (5major+1092800minor)pagefaults 0swaps

So that's 6.5x less CPU, and 4x faster wall time.

Also, multiple goroutines threads doesn't seem to help.

krasi-georgiev:
@brian-brazil any way to add a bench test as well?

head.go
```diff
@@ -376,8 +378,9 @@ func (h *Head) loadWAL(r *wal.Reader) error {
 				firstInput <- append(buf[:0], samples[:n]...)
 				samples = samples[n:]
 			}
+			samples = s // Keep whole slice for reuse.
```
narqo: Why did you need to keep the original slice here? After you'd fixed `samples` to be local to the whole function (the change on L359), didn't it be enough to fix the issue? I believe, `samples` and `s` are backed by the same array.

brian-brazil: They're backed by the same array, but samples would pointing at the end of it so it'd end up as a memory leak.

PR #441, opened 2018-11-01
Use already open blocks while compacting.

brian-brazil:
This roughly halves the RAM requirements of compaction.

There's not much point in optimising compaction memory itself further, as loading in the new blocks would OOM at reload time. Reducing the memory used by the indexes would be the main hope.

This also makes compaction a little bit faster.

### DIFF
--- a/compact.go
+++ b/compact.go
@@ -59,7 +59,9 @@ type Compactor interface {
 
 	// Compact runs compaction against the provided directories. Must
 	// only be called concurrently with results of Plan().
-	Compact(dest string, dirs ...string) (ulid.ULID, error)
+	// Can optionally pass a list of already open blocks,
+	// to avoid having to reopen them.
+	Compact(dest string, dirs []string, open []*Block) (ulid.ULID, error)
 }
 
 // LeveledCompactor implements the Compactor interface.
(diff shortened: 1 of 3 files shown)


krasi-georgiev:
would you mind opening the same PR on Prometheus so we can run a bench test to see it in action?

bwplotka:
Nice @brian-brazil 

Just curious how you measured the RAM consumption? Profiling, flamegraph? Can see you added timings as well.

Also just wondering in what case Prometheus had those blocks already opened? Opened because what?

brian-brazil:
Heap profiling.

They'll already be loaded by Prometheus as they're normal blocks.

compact.go
```diff
@@ -317,24 +319,39 @@ func compactBlockMetas(uid ulid.ULID, blocks ...*BlockMeta) *BlockMeta {
...
-		defer b.Close()
 
-		meta, err := readMetaFile(d)
-		if err != nil {
-			return uid, err
+		var b *Block
+
+		// Use already open blocks if we can, to avoid
+		// having the index data in memory twice.
+		for _, o := range open {
```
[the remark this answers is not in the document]
brian-brazil: My thinking was there's usually only 10ish open blocks, so no need to optimise.

compact.go
```diff
@@ -59,7 +59,9 @@ type Compactor interface {
 
 	// Compact runs compaction against the provided directories. Must
 	// only be called concurrently with results of Plan().
-	Compact(dest string, dirs ...string) (ulid.ULID, error)
+	// Can optionally pass a list of already open blocks,
+	// to avoid having to reopen them.
+	Compact(dest string, dirs []string, open []*Block) (ulid.ULID, error)
```
krasi-georgiev: Since this is changing public interface. Could you tag a minor version? We don't have any tags yet so maybe a good time to start now.

brian-brazil: 0.1.0?

krasi-georgiev: Yeah , or maybe even 1.1.0 , since it is pretty stable now and I think go modules don't pull anything below 1.0. It would be nice if we could use these tags as well  when we vendor it in Prometheus.

simonpasquier: > I think go modules don't pull anything below 1.0

Go modules do work with v0.x.y tags. Cutting a v1 is a pretty heavy commitment and I tend to think that a v0.something is appropriate here.

brian-brazil: Okay, so I should merge and then add a v0.1.0 tag?

krasi-georgiev: yep lets try something as simple as that, unless @simonpasquier has any other recommendations.

I will keep reading how are we suppose to do this properly, but for now it is ok to start with just this.

brian-brazil: We've changed the API many times before, we have to start somewhere.

PR #442, opened 2018-11-01
Don't panic for index out of bounds in reader.next

cstyan:
catch case in reader.next where we may try to access data outside the buffers slice bounds, avoid panic

This is mostly relevant to https://github.com/prometheus/prometheus/pull/4588, where we read from the WAL segment on every write event. For a series of write events we could read all of the data for subsequent write events in an earlier event, causing all later reads to panic (which at the moment I'm handling with a recover.

@gouthamve @fabxc

### DIFF
--- a/wal/wal.go
+++ b/wal/wal.go
@@ -752,6 +752,9 @@ func (r *Reader) next() (err error) {
 			if k == pageSize {
 				continue // Initial 0 byte was last page byte.
 			}
+			if int(k) > len(buf) {
+				return errors.New("read would result in slice bounds out of range")
+			}
 			n, err := io.ReadFull(r.rdr, buf[:k])
 			if err != nil {
 				return errors.Wrap(err, "read remaining zeros")


krasi-georgiev:
probably a duplicate of https://github.com/prometheus/tsdb/pull/429

can you explain in test how to replicate?

cstyan:
@krasi-georgiev I think this is a separate issue. I can try to replicate via a test but here's an explanation of what I see happening.

In the remote write via WAL PR, I'm listening for FS write and create events in the WAL dir with fsnotify. On write event to the current segment, we read some records. It's seems to be possible that during the read for a given write event, more data will be written to the segment and so the current read will read the data for the next write event as well. Then, when the loop iterates and we start to process that next write event, there isn't more data in the segment to read, so the buffer index would go out of range and cause a panic. Does that make sense?

cstyan:
@krasi-georgiev had a hard time reproducing this as a test. It comes up reliably in my remote write PR when reading the WAL segment for the last write event to a segment before a new segment is created (meaning we read the last of the data on the 2nd last write event). The test here causes a panic about 50% of the time: https://github.com/cstyan/tsdb/commit/f6a0baec03247b75f56c12dba0649c7346eec712 ping me on IRC if you want to talk about it

krasi-georgiev:
I was able to replicate it using your test and my PR fixes the issues.

I will spend some more time to find out what is happening, but in the meanwhile you can use my PR to confirm if it fixes the issue.

krasi-georgiev:
I figured what is happening in your test. There is a race. You are writing to the wal `err := w.Log(recs...)` and at the same time in another go routine calling `nw, err := New(nil, nil, walDir)`. When this happens before the wal page is complete it forces adding 0 padding for the rest of the wal segment file.

https://github.com/prometheus/tsdb/blob/a9470dd8d54f7df3e6212ddee72d014bd63e9b3d/wal/wal.go#L109-L119

On another note in my experience  fsnotify is not a very good solution to monitor file changes. Maybe do it periodically, when the size changes or some other criteria. 

In any case my PR fixes the issue.

I have also added an info log to my PR, that will at least display when this padding happens in  such cases.

cstyan:
@krasi-georgiev I haven't seen this race condition you're referring to, since without the recover we would get the panic when reading near the end of any segment, not just the segment that existed when the call to New was made.

krasi-georgiev:
I spent several hours today and with great certainty can say that this is what is happening :smiley_cat: 
Although my PR fixes the issues, if your implementation for the remote write uses a similar approach this race would cause data loss.

krasi-georgiev:
superseded by https://github.com/prometheus/tsdb/pull/429

PR #443, opened 2018-11-02
Reduce memory taken up by posting/symbol tables.

brian-brazil:
Reuse the string already allocated for symbols
in the posting tables.

Use a slice for symbols in v2 format.

Move symbol size logic into the index code.
Avoid duplication of lookupSymbol logic.

On my test data with 100000 unique values for the label `llllllllllll` this brought index memory usage from 12.1MB  to 8.4MB.

### DIFF
--- a/block.go
+++ b/block.go
@@ -15,7 +15,6 @@
 package tsdb
 
 import (
-	"encoding/binary"
 	"encoding/json"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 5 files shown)


brian-brazil:
Found another improvement, down to 5.9MB so about half the usage now.

brian-brazil:
I have also applied similar to the head posting map, which cut out 2MB of usage (12MB->9.8MB)/

PR #444, opened 2018-11-02
Use sampleBuf instead of maintaining lastValue.

brian-brazil:
This cuts the size of memSize by 8B.

If you had 10M head series, that's a 80MB saving.


gouthamve:
LGTM! 

Curious, can we use the same for `s.maxTime` too?

brian-brazil:
Looks like, I'll do up a PR.

brian-brazil:
So you could do that with s.maxTime, but there's not much point as it's only used in deletion and only saves us an if.

PR #445, opened 2018-11-06
[Proof of Concept - Don't Merge] Benchmark embedded M3DB against TSDB

richardartoul:
After meeting Krasi at OpenFest 2018 we decided to investigate embedded M3DB vs OpenTSDB performance. This P.R modifies the TSDB benchmark to use embedded M3DB to compare.

Results:

TSDB
```
ingestion completed
>> completed stage=ingestScrapes duration=8.910575888s
 > total samples: 60000000
 > samples/sec: 6.733558191366459e+06
>> start stage=stopStorage
```

TSDB Without "AddFast"
```
ingestion completed
>> completed stage=ingestScrapes duration=19.842553198s
 > total samples: 60000000
 > samples/sec: 3.023802862388617e+06
>> start stage=stopStorage
```

M3DB
```
>> start stage=ingestScrapes
ingestion completed
>> completed stage=ingestScrapes duration=7.330143645s
 > total samples: 60000000
 > samples/sec: 8.185342167325305e+06
>> start stage=stopStorage
>> completed stage=stopStorage duration=141.166402ms
```

The benchmarks show that M3DB appears to have a write path that is at least as fast as TSDBs, if not faster. In addition, M3DB is more than twice as fast as TSDB when the "AddFast" path is disabled. M3DB does not have an equivalent path because it is more optimized for receiving traffic over the network than via embedding, but I believe that with about one weeks work we could add some optimizations to M3DB for this use-case which would make it at least 40% faster than even the benchmarks we are seeing here (which are already faster than TSDB in both cases). In addition, M3DB is highly tunable for various workloads and I ran this benchmark with pretty much the default configuration so there is probably a lot of room for improvement simply for making config changes.

That said, these benchmarks are extremely synthetic, the real test would be to try and embed M3DB in Prometheus itself and see what kind of impact that makes. Raw write speed isn't really a good benchmark though because the numbers will change when concurrent reads / compactions / etc are going on. That said, we think that its quite possible M3DB will compare favorably in that situation just because a lot of engineering effort has gone into it (we have a small team that has been developing and maintaining it in production for 3 years), and it doesn't really perform compactions. There may be some minor optimizations we need to make for the embedded use-case, but those should be small. In addition, there is the interesting option of exploring some of M3DB distributed functionality in Prometheus.

Reproduction steps:

1. Pull this branch
2. CD into the directory
3. Run `glide install` (I had to migrate the repo to glide due to some versioning issues that were cropping up as a result of `go get`)
4.  Run `make bench ENGINE=tsdb` or `make bench ENGINE=m3db`

Also I apologize in advance for the really hacky code. I just wanted to get some measurements, but if we wanted to keep this benchmark around I could clean it up and make it nice.


cmd/tsdb/main.go
```diff
@@ -31,19 +31,35 @@ import (
...
 	"github.com/pkg/errors"
 	"github.com/prometheus/tsdb"
 	"github.com/prometheus/tsdb/labels"
 	"gopkg.in/alecthomas/kingpin.v2"
 )
 
+var (
+	defaultM3DBNamespace = ident.StringID("defaultM3DBNamespace")
+	defaultSeriesID      = ident.StringID("default_m3db_id")
+	defaultNumShardsM3DB = 1
```
robskillington: I would set this to `runtime.NumCPU()` to reduce contention as much as possible.

richardartoul: Oh I already updated that, let me push the latest branch (turns out it makes very little diff)

robskillington: I see, yeah looks like most of the IDs are pretty stable? i.e. not a lot of insertions after startup.

RichiH:
Could you sign-off on your commits, please? This does sound interesting even if it's somewhat artificial, atm.

krasi-georgiev:
@RichiH that is not intended for merging, rather to see how feasible it is to use M3DB for Prometheus.

I think it would be good to see some read tests as well.

brancz:
I don’t want to ruin the party, but if we even remotely wanted to use m3db that would need a much larger discussion and in my opinion it has to significantly outperform tsdb in essentially every regard. The current results look pretty nice but in my opinion not good enough to justify adoption. We’re very familiar with tsdb, we have no idea about he m3db codebase and we don’t have control over m3db in regard to compatibility. 

That said at first glance it looks like superb engineering and I’m still very interested in the benchmarks so without merging, I’m in favor of continuing experiments. 

@richie even for benchmark comparison we shouldn’t merge this as it pulls in vendored deps.

krasi-georgiev:
> we don’t have control over m3db in regard to compatibility.
@brancz fair and an important point , on the plus side we would have a full time team behind such an important library for Prometheus.

robskillington:
I would have thought too the binary file compatibility making any change of this nature a major breaking change being a pretty big blocker?

robskillington:
Breaking changes to disk formats is usually a no-no, however if was just in a config option for when running Prometheus then that could make it a little less onerous.

The option could be in the config file, a command line switch, or even a compilation option to make it less likely people will play with it to make the support load manageable.  

My counter-point to this though, is that you'll have to still support this storage option being able to be used long term, and it also means if you want to add a feature that's only doable/support by one storage engine, then it means making that change is harder because you'll have to either stop supporting the non-compatible storage engines or work together to add the features to both at the same time (causing slower progress potentially).

gouthamve:
Hi, I am super interested to see how m3db performs in actual Prometheus workloads. Having said that, unless there is a 2+x improvement across the board. Having said the initial benchmark has piqued my interest.

if anything, this will show us where we can improve the in-mem head block structures :) But again, the memory/CPU usage should also be considered. Right now, in Prometheus 2.0, the ingest is not a bottleneck.

Would be interesting to see how m3db vs TSDB would work with on-disk data, and in general with Prometheus server using it.

cmd/tsdb/main.go
```diff
@@ -188,57 +329,110 @@ func (b *writeBenchmark) ingestScrapes(lbls []labels.Labels, scrapeCount int) (u
...
+		batch := make(commitlog.WritesBatch, len(scrape))
+		for i := 0; i < scrapeCount; i++ {
+			ts += timeDelta
+			id := ident.StringID(string(i))
+			for j, s := range scrape {
+				s.value += 1000
+				batch[j] = commitlog.Write{
+					Series: commitlog.Series{
+						Namespace: defaultM3DBNamespace,
+						ID:        id,
```
fabxc: `s.tags` obivously differs for each here but `ID` stays the same – what does that mean?

Do I read correctly, that all samples of a scrape are appended to the same series with an ID of simply the scrape index?
The tags would obviously be different so that wouldn't seem right. Using `j` would not be correct either since scrapes are not guaranteed to always contain the same series in the same order.

That looks like it is always doing an equivalent of `AddFast`. The main problem in TSDB is really getting from an `s.tags` to an`ID`. That's why `Add` is so much slower (each insert essentially requires a full index lookup) and there's dedicated `AddFast` path.

So I guess my question is whether I misunderstand what `ID` stands for or how we would be able to figure out the correct one if all we have is a set of tags.

cmd/tsdb/main.go
```diff
@@ -188,57 +329,110 @@ func (b *writeBenchmark) ingestScrapes(lbls []labels.Labels, scrapeCount int) (u
...
-				ref, err := app.Add(s.labels, ts, float64(s.value))
-				if err != nil {
-					panic(err)
-				}
-				s.ref = &ref
+				total++
 			}
 
-			total++
+			err := b.m3dbStorage.WriteTaggedBatch(nil, defaultM3DBNamespace, batch)
```
fabxc: Is this using a development version? The godoc of the package only seems to contain non-batched methods.

cmd/tsdb/main.go
```diff
@@ -31,19 +31,37 @@ import (
...
 	"github.com/pkg/errors"
 	"github.com/prometheus/tsdb"
 	"github.com/prometheus/tsdb/labels"
 	"gopkg.in/alecthomas/kingpin.v2"
 )
 
+var (
+	defaultM3DBNamespace = ident.StringID("defaultM3DBNamespace")
+	defaultSeriesID      = ident.StringID("default_m3db_id")
+	defaultNumShardsM3DB = 100
```
fabxc: Does m3db guarantee atomicity across shards? Scrapes are supposed to be atomic, which is why prometheus/tsdb only has a single shard.

To be fair, the actual MVCC isn't implemented yet (what's the status there @gouthamve?). But the resulting slowdown due to locking is already factored in. Would be interesting to get numbers with this set to `1` unless there's atomicity across shards already (or possible in the future).

cmd/tsdb/main.go
richardartoul: Yeah sorry I mentioned this in an email to Krasi but should have posted it here. We actually found a bug in our original benchmark that made us much slower and then I quickly hacked together a development branch to improve our efficiency in this type of embedded workload which is what this is running.

I'm currently in the process of productioning those fixes and they should be merged into master within a week. There was basically just some low hanging fruit that we'd never bothered to do because we hadn't had a reason to optimize for the embedded case until now.

cmd/tsdb/main.go
richardartoul: We don't have any atomicity guarantees (we don't have any concept of transactions within a shard or across shards). If a batch write fails (lets say half the write succeed and half fail for some weird reason), then you can retry the entire thing and if everything succeeds you'll be in the same state as if everything succeeded the first time, but if you don't retry then half of the datapoints will still be present.

I assume thats what you meant by atomic?

We could explore supporting that functionality, but we've just never had a reason for it before. Some data has always been better than no data for us.

cmd/tsdb/main.go
richardartoul: Ah ok I may have done the wrong thing here. I'm still new to Prometheus / TSDB so still trying to figure out the model.

Is the way this is supposed to work is that each individual "scrape" is a different "series"? If so, I can fix the benchmark. The confusion lies around the difference in-between how TSDB and M3DB identify a given series. It seems like in TSDB a series is identified by a combination of tags, whereas in M3DB every series has a unique ID and (optionally) a bunch of tags attached to that ID.

If you write different tags to the same ID the new tags will just be ignored. 

I don't think we're doing the equivalent of AddFast because even with the WriteTaggedBatch API we still have to look up each series in-memory for each individual write. The WriteTaggedBatch is basically just an optimization to avoid a lot of contention on our commitlog queue, but doesn't do any lookup optimization.

In terms of figuring out the correct ID given only a set of tags (for the write path at least) I think we'd need to do something that deterministically generates the same ID from a given set of tags, I think thats what our M3Coordinator service does when its receiving data from Prometheus over the remote read/write endpoint and then inside of M3DB we have some optimizations to re-use the memory that is allocated for the tags within the generated ID so it has minimal impact on memory utilization

richardartoul:
@fabxc Thanks for taking a look at this! I think I have a better understanding of how this all fits together now. In a few days I'll have the development changes we're making to optimized embedded workloads complete and then I can adjust the benchmark to mimic the Prometheus workload more correctly (especially in terms of the mapping between tags and IDs not that I understand that better) and then maybe we can discuss again.

Also I totally agree that this is a very synthetic benchmark and that Prometheus is probably not bottlenecked by the ingest path (neither are our M3DB clusters). The reason I was running this benchmark was that Krasi mentioned that some users were having trouble with compactions with TSDB using too much memory and I mentioned that our design pretty much avoids compactions. You need to do a little bit of tuning to get the blocksize configuration right, but once a workload gets going M3DBs performance is pretty much completely stable with no expensive background processes, but before we wanted to consider doing more advanced comparisons we wanted to make sure that we were at least operating within the same ballpark of performance for the "simple" benchmarks

krasi-georgiev:
I haven't had a chance to look at the code yet, but I think next week would be able to spend more time here. 

@richardartoul in the meanwhile feel free to ping me on the #prometheus-dev channel if you need any pointers.

cmd/tsdb/main.go
fabxc: Each sample in a scrape belongs to a different series in the general case – at least in the way Prometheus does writes.
When we scrape a `/metrics` endpoint, a bunch of tags is all we get to identify what a sample belongs to. The targets have no way to tell us at scrape time what database ID a series they expose map to.
So figuring out what series that has to be written to is the fundamental difficulty and makes up most of the work TSDB does.

A simple way for m3db would probably be to use a sorted and serialized version of the tags/labels as the ID. This is usually rather expensive, which is why Prometheus+TSDB play quite a few tricks with Add and AddFast.

krasi-georgiev:
after some emails decided to not invest anymore time in this and will revisit if needed

Issue #336, opened 2018-05-29
magic number of tombstones in readme


olegtikhonov:
Hi, please correct the magic number of tombstones in tsdb/docs/format/tombstones.md
instead of being a magic(0x130BA30) to be a magic(0x0130BA30)

olegtikhonov:
Great !!! Thanks.

PR #448, opened 2018-11-08
add missing zero to tombstone magic number

krasi-georgiev:
closes #https://github.com/prometheus/tsdb/issues/336

cc @codesome 


codesome:
This won't change anything functionality wise, as were are writing and reading 32 bits. But for readability, LGTM :+1:.

PR #449, opened 2018-11-12
refactor NewSegmentsRangeReader to take multi WAL ranges

krasi-georgiev:
In case of an error when checkpointing the WAL the error doesn't show
the exact WAL index that is corrupter. this is because it uses
MultiReader to read multiply WAL files.
This refactoring allows the NewSegmentsRangeReader to take more than a
single WAL range and it reads all of the ranges by iterating each one.

this changes the logs from
```
create checkpoint: read segments: corruption after 4841144384 bytes:...
```
to
```create checkpoint: read segments: corruption in segment data/wal/00017351 at 123142208: ...```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,5 @@
+## Next Release API changes.
+
+ - `LastCheckpoint` used to return just the segment name and now it returns the full path.
+ - `NewSegmentsRangeReader` can now read over miltiple wal ranges by using the new `SegmentRange` struct.
+ - `CorruptionErr` now also exposes the Segment `Dir` which is added when displayin any errors. 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## Next Release API changes.
+## master / unreleased
 
  - `LastCheckpoint` used to return just the segment name and now it returns the full path.
  - `NewSegmentsRangeReader` can now read over miltiple wal ranges by using the new `SegmentRange` struct.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## master / unreleased
 
- - `LastCheckpoint` used to return just the segment name and now it returns the full path.
+ - `LastCheckpoint` used to return just the segment name and now it returns the full relative path.
  - `NewSegmentsRangeReader` can now read over miltiple wal ranges by using the new `SegmentRange` struct.
- - `CorruptionErr` now also exposes the Segment `Dir` which is added when displayin any errors. 
+ - `CorruptionErr` now also exposes the Segment `Dir` which is added when displaying any errors.
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"math"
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: checkpoint.go [hunk 1]
```diff
@@ -118,27 +117,15 @@ func Checkpoint(w *wal.WAL, from, to int, keep func(id uint64) bool, mint int64)
 			// Ignore WAL files below the checkpoint. They shouldn't exist to begin with.
 			from = last
 
-			r, err := wal.NewSegmentsReader(filepath.Join(w.Dir(), dir))
-			if err != nil {
-				return nil, errors.Wrap(err, "open last checkpoint")
-			}
-			defer r.Close()
-			closers = append(closers, r)
```
krasi-georgiev: @gouthamve 	do you remember why using just `defer r.Close()` is not enough?

krasi-georgiev:
originally reported by @SuperQ

checkpoint.go
krasi-georgiev: ping @gouthamve

krasi-georgiev: yeah I read this one few times , but you are calling `closeAll(closers...)` right at the end before the return so it is essentially the same as defer.

gouthamve: Ah, I see :) When I first added this, it had deletion of the segments in the function itself :P Now they have been moved out of the function.

See: https://github.com/prometheus/tsdb/pull/392/files#diff-682f9cc560724652272cca157347157bR275

krasi-georgiev: ah nice, so this PR is correct?

CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+## Next Release API changes.
```
SuperQ: We've been using the header `## master / unreleased` in the node_exporter for changelog entries.

Then we add a tag/date as part of doing the next release.

krasi-georgiev: good idea, thanks, updated!

CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+## master / unreleased
+
+ - `LastCheckpoint` used to return just the segment name and now it returns the full path.
+ - `NewSegmentsRangeReader` can now read over miltiple wal ranges by using the new `SegmentRange` struct.
+ - `CorruptionErr` now also exposes the Segment `Dir` which is added when displayin any errors.
```
gouthamve:
```suggestion
 - `CorruptionErr` now also exposes the Segment `Dir` which is added when displaying any errors. 
```

CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+## master / unreleased
+
+ - `LastCheckpoint` used to return just the segment name and now it returns the full path.
```
gouthamve: Full path might mean absolute path, which may not be true. It returns the full relative path instead of just saying `full path`?

Minor though, no strong feelings.

CHANGELOG.md
krasi-georgiev: Aaargh, when will vscode add a spell check for the comments!

CHANGELOG.md
krasi-georgiev: good point! updated

PR #451, opened 2018-11-13
return an error when the  last wal segment record is torn.

krasi-georgiev:
this ensures that a repair will be run when the last record in a segment is torn.

to reproduce:
 - kill Prometheus just after a first record is written (before a last record is written).
 - Start Prometheus. It show no signs of the issue and continues to write records after the torn record. This triggers a dead loop at checkpointing.

most likely the cause for:
https://github.com/prometheus/prometheus/issues/4695
~https://github.com/prometheus/prometheus/issues/4842~

cc @fabxc , @SuperQ 	

note to myself :smiling_imp: 
the wal records are grouped and written in pages, each  page is [32kb](https://github.com/prometheus/tsdb/blob/171fc4ab5da16793145d0e1df4403a073b1d2f78/wal/wal.go#L40) and when the last record doesn't fit in the current page it is split in parts(first,last). So the last record of one page would have the first part of a record and the first record in the next page would have the last part of a record.
So in case of a crash just after writing the first part the next WAL read didn't have a  check to trigger a repair in such cases and the PR fixes this.

### DIFF
--- a/wal/wal.go
+++ b/wal/wal.go
@@ -760,15 +760,12 @@ func (r *Reader) next() (err error) {
 		r.curRecTyp = recType(hdr[0])
 
 		// Gobble up zero bytes.
-<<<<<<< HEAD
-		if typ == recPageTerm {
-			// recPageTerm is a single byte that indicates that the rest of the page is padded.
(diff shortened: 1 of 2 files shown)


krasi-georgiev:
ping @gouthamve

PR #453, opened 2018-11-15
repair wal when the record cannot be decoded

krasi-georgiev:
fixes https://github.com/prometheus/prometheus/issues/4705

Currently repair is run only when the error happens in the reader.

A corruption can occur after the record is read and when it is decoded.
This change wraps the error at decoding as a CorruptionErr as this is error
that triggers the repair.

### DIFF
--- a/head_test.go
+++ b/head_test.go
@@ -861,3 +861,81 @@ func TestHead_LogRollback(t *testing.T) {
 	testutil.Assert(t, ok, "expected series record but got %+v", recs[0])
 	testutil.Equals(t, []RefSeries{{Ref: 1, Labels: labels.FromStrings("a", "b")}}, series)
 }
+
+func TestWalRepair(t *testing.T) {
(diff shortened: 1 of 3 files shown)


krasi-georgiev:
ping @gouthamve

head_test.go
```diff
@@ -861,3 +861,81 @@ func TestHead_LogRollback(t *testing.T) {
...
+					continue
+				}
+				testutil.Ok(t, w.Log(test.rec))
+			}
+
+			h, err := NewHead(nil, nil, w, 1)
+			testutil.Ok(t, err)
+			testutil.Ok(t, h.Init())
+
+			sr, err := wal.NewSegmentsReader(dir)
```
gouthamve: `defer sr.Close()` else this test'll break in Windows :P

I'm not sure if I should feel good or bad about catching all these Windows cases xD

PR #454, opened 2018-11-15
Use Go modules

simonpasquier:


brancz:
lgtm :+1:

krasi-georgiev:
wow the common makefile is starting to grow quite a bit. 

Will these extra conditional checks be removed once the go modules are officially supported by default?

simonpasquier:
> Will these extra conditional checks be removed once the go modules are officially supported by default?

Sure but because of projects such as [client_golang](https://github.com/prometheus/client_golang/) that need to support older Go versions, it will take some time to get there.

simonpasquier:
The failing tests are due to Travis, nothing to do with this patch.

> Is there any other project apart from client_golang that required this? If not than maybe just have a custom Makefile for that project only. No need to pollute everything else because of a single project.

IMO it defeats the purpose of having a common Makefile. Part of the changes in this patch aren't related to Go modules (see https://github.com/prometheus/prometheus/pull/4843).

krasi-georgiev:
that is strange 
I re-run the test and it is failing.

simonpasquier:
It is a different failure now. Checking.

simonpasquier:
Since https://github.com/prometheus/tsdb/pull/412 `github.com/prometheus/client_model` is a direct dependency (while it was indirect before): this is why Travis was failing.

krasi-georgiev:
yep that makes sense. 
Thanks.

Still a bit concerned of how quickly the Makefile.common grew in size and logic and I hope it doesn't go out of control for covering too many edge cases.

lgtm

PR #455, opened 2018-11-15
LabelNames() for Querier

codesome:
https://github.com/prometheus/prometheus/pull/4835 requires the `LabelNames()` function to be in the `Querier`. This PR removes it from the `DB` and adds it to `Querier`.

### DIFF
--- a/querier.go
+++ b/querier.go
@@ -37,6 +37,9 @@ type Querier interface {
 	// under the constraint of another label.
 	LabelValuesFor(string, labels.Label) ([]string, error)
 
+	// LabelNames returns all the unique label names present in the block in sorted order.
+	LabelNames() ([]string, error)
+
 	// Close releases the resources of the Querier.
 	Close() error
 }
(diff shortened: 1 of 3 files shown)


gouthamve:
@krasi-georgiev @bwplotka This breaks interface again, but it just fits in better in the `Querier`. See `Querier.LabelValues`.

krasi-georgiev:
sure 
we should have tested the implementation in Prometheus before merging the first version.
 @gouthamve  I will let you review this one since you know how it will be used in Prometheus.

I didn't add a new tag when adding this so can you please add a new 0.2.0 annotated tag(with some details) about this api change.

krasi-georgiev:
yep , and you should be able to annotate the tag to include some details about the API changes.

querier.go
```diff
@@ -60,6 +63,28 @@ func (q *querier) LabelValues(n string) ([]string, error) {
 	return q.lvals(q.blocks, n)
 }
 
+// LabelNames returns all the unique label names present querier blocks.
+func (q *querier) LabelNames() ([]string, error) {
+	labelNamesMap := make(map[string]struct{})
+	for _, b := range q.blocks {
+		names, err := b.LabelNames()
+		if err != nil {
+			return nil, errors.Wrap(err, "LabelNames() from IndexReader")
```
bwplotka: Well technically it is from Querier interface. The fact that it Is  from index reader is just assuming implementation. Worth to maybe change this to Querier?

querier.go
```diff
@@ -37,6 +37,9 @@ type Querier interface {
 	// under the constraint of another label.
 	LabelValuesFor(string, labels.Label) ([]string, error)
 
+	// LabelNames returns all the unique label names present in the block in sorted order.
```
bwplotka: Not yours, but can you fix new lines in this interface? Couple lines before this.

querier.go
codesome: That is a typo (I moved the previous implementation and forgot to change the message). Thanks!

codesome:
@gouthamve is this good to go?

PR #457, opened 2018-11-19
remove opaque metrics

krasi-georgiev:
tsdb can save timestamps in any unit - sec,mils etc.
These metrics should be exposed by the library caller so they include
the unit suffix based on the unit used by the caller.

The api exposed enough details to allow this top be implemented by the
caller.

I will open a PR in Prometheus to add these there.

### DIFF
--- a/db.go
+++ b/db.go
@@ -130,7 +130,6 @@ type dbMetrics struct {
 	compactionsSkipped   prometheus.Counter
 	cutoffs              prometheus.Counter
 	cutoffsFailed        prometheus.Counter
-	startTime            prometheus.GaugeFunc
 	tombCleanTimer       prometheus.Histogram
 }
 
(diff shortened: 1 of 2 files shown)


brian-brazil:
These are still useful for general tsdb debugging, I'm not sure we should be removing them.

krasi-georgiev:
if they can be be implemented by the caller you don't loose the debugging

brian-brazil:
You do if they're not implemented. This is a perfectly fine metric for this library to expose, as it's a key internal bit of state.

krasi-georgiev:
I am opening a PR to add these to Prometheus

krasi-georgiev:
what is the point these will be exactly the same metrics?

gouthamve:
As a user of the library, I would like to see these metrics being exposed without me having to write functions to get the start, end time for example.

We can maybe rename them to something else to avoid confusion with the prometheus metric.

simonpasquier:
> We can maybe rename them to something else to avoid confusion with the prometheus metric.

Definitely yes to this.

krasi-georgiev:
PR to add these to prometheus https://github.com/prometheus/prometheus/pull/4888

@gouthamve 	any suggestions about renaming these to be relevant for what you need it in tsdb?

krasi-georgiev:
seems that there is no better name than the current one so leaving as is with just some more text in the help.

ping @gouthamve , @brian-brazil

brian-brazil:
:+1:

PR #459, opened 2018-11-21
Make newBReader return a struct, not a pointer.

tomwilkie:
This shows up as a hot spot in profiles of queries involving lots of seeks, as each seek creates a new iterator.

### DIFF
--- a/chunkenc/bstream.go
+++ b/chunkenc/bstream.go
@@ -49,8 +49,8 @@ type bstream struct {
 	count  uint8  // how many bits are valid in current byte
 }
 
-func newBReader(b []byte) *bstream {
-	return &bstream{stream: b, count: 8}
+func newBReader(b []byte) bstream {
+	return bstream{stream: b, count: 8}
 }
 
 func newBWriter(size int) *bstream {
--- a/chunkenc/xor.go
+++ b/chunkenc/xor.go
@@ -221,7 +221,7 @@ func (a *xorAppender) writeVDelta(v float64) {
 }
 
 type xorIterator struct {
-	br       *bstream
+	br       bstream
 	numTotal uint16
 	numRead  uint16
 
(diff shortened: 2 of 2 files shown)


gouthamve:
Curious, how is this better?

tomwilkie:
Thats right; I think the bstream is getting allocated on the heap, when it doesn't need to be.

Before:
```
     2.50s      4.11s (flat, cum) 23.85% of Total
         .          .    105:func (c *XORChunk) iterator() *xorIterator {
         .          .    106:	// Should iterators guarantee to act on a copy of the data so it doesn't lock append?
         .          .    107:	// When using striped locks to guard access to chunks, probably yes.
         .          .    108:	// Could only copy data if the chunk is not completed yet.
         .          .    109:	return &xorIterator{
     2.50s      3.74s    110:		br:       newBReader(c.b.bytes()[2:]),
         .      370ms    111:		numTotal: binary.BigEndian.Uint16(c.b.bytes()),
         .          .    112:	}
         .          .    113:}
         .          .    114:
         .          .    115:// Iterator implements the Chunk interface.
         .          .    116:func (c *XORChunk) Iterator() Iterator {
```

After:
```
ROUTINE ======================== github.com/cortexproject/cortex/vendor/github.com/prometheus/tsdb/chunkenc.(*XORChunk).iterator in /Users/twilkie/Documents/src/github.com/cortexproject/cortex/vendor/github.com/prometheus/tsdb/chunkenc/xor.go
     2.37s      2.89s (flat, cum) 17.03% of Total
         .          .    105:func (c *XORChunk) iterator() *xorIterator {
         .          .    106:	// Should iterators guarantee to act on a copy of the data so it doesn't lock append?
         .          .    107:	// When using striped locks to guard access to chunks, probably yes.
         .          .    108:	// Could only copy data if the chunk is not completed yet.
         .          .    109:	return &xorIterator{
     2.37s      2.39s    110:		br:       newBReader(c.b.bytes()[2:]),
         .      500ms    111:		numTotal: binary.BigEndian.Uint16(c.b.bytes()),
         .          .    112:	}
         .          .    113:}
         .          .    114:
         .          .    115:// Iterator implements the Chunk interface.
         .          .    116:func (c *XORChunk) Iterator() Iterator {
```

PR #461, opened 2018-11-30
no overlapping on compaction when an existing block is not within default boundaries.

krasi-georgiev:
closes https://github.com/prometheus/prometheus/issues/4643

### DIFF
--- a/block_test.go
+++ b/block_test.go
@@ -77,7 +77,8 @@ func createEmptyBlock(t *testing.T, dir string, meta *BlockMeta) *Block {
 	return b
 }
 
-// createPopulatedBlock creates a block with nSeries series, and nSamples samples.
+// createPopulatedBlock creates a block with nSeries series, filled with
+// samples of the given mint,maxt time range.
 func createPopulatedBlock(tb testing.TB, dir string, nSeries int, mint, maxt int64) *Block {
 	head, err := NewHead(nil, nil, nil, 2*60*60*1000)
 	testutil.Ok(tb, err)
--- a/block_test.go
+++ b/block_test.go
@@ -78,7 +78,7 @@ func createEmptyBlock(t *testing.T, dir string, meta *BlockMeta) *Block {
 }
 
 // createPopulatedBlock creates a block with nSeries series, and nSamples samples.
-func createPopulatedBlock(tb testing.TB, dir string, nSeries, nSamples int) *Block {
+func createPopulatedBlock(tb testing.TB, dir string, nSeries int, mint, maxt int64) *Block {
 	head, err := NewHead(nil, nil, nil, 2*60*60*1000)
 	testutil.Ok(tb, err)
 	defer head.Close()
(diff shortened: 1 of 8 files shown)


krasi-georgiev:
@gouthamve 	, @brian-brazil 	current master is broken by https://github.com/prometheus/tsdb/commit/d50b9a5619b2efb7559ea1a7187df9fa5c209121

The logic for loading the WAL depends on knowing if there are already blocks so `db.reload()` needs to be called before `head.Init()`

The test in this PR ensures this behaviour and will pass after reverting https://github.com/prometheus/tsdb/commit/d50b9a5619b2efb7559ea1a7187df9fa5c209121

head.go
```diff
@@ -640,13 +640,20 @@ func (h *Head) Appender() Appender {
 func (h *Head) appender() *headAppender {
 	return &headAppender{
 		head:         h,
-		minValidTime: h.MaxTime() - h.chunkRange/2,
+		minValidTime: max(h.MinTime(), h.MaxTime()-h.chunkRange/2),
```
krasi-georgiev: I am not entirely sure about this one so let me know if you think it might brake anything else.

block_test.go
```diff
@@ -78,7 +78,7 @@ func createEmptyBlock(t *testing.T, dir string, meta *BlockMeta) *Block {
 }
 
 // createPopulatedBlock creates a block with nSeries series, and nSamples samples.
```
brian-brazil: Needs updating

db.go
```diff
@@ -395,7 +395,8 @@ func (db *DB) compact() (err error) {
 		if db.head.MaxTime()-db.head.MinTime() <= db.opts.BlockRanges[0]/2*3 {
 			break
 		}
-		mint, maxt := rangeForTimestamp(db.head.MinTime(), db.opts.BlockRanges[0])
+		mint := db.head.MinTime()
+		_, maxt := rangeForTimestamp(mint, db.opts.BlockRanges[0])
```
brian-brazil: We now never use the first return value, so this function can be simplified.

head.go
brian-brazil: > adding samples with timestamps lower than the first appended sample.

That's a valid use case.

krasi-georgiev:
@gouthamve, @brian-brazil  I have refactored the logic to what I think makes sense.

`TestHead_ReadWAL` fails because after loading the wal `{Ref: 11, Labels: labels.FromStrings("a", "2")},` is garbage collected as it has no samples. Should it remain there or the test is wrong?

https://github.com/prometheus/tsdb/blob/01e8296ee1cdcf4c90de6c3ea7b2e2012513781f/head_test.go#L95-L132

brian-brazil:
That shouldn't be GCed, as the GC code isn't called in that test.

krasi-georgiev:
@brian-brazil 	as I mentioned I refactored the logic , I am asking if the behaviour is correct and whether series should remain in the head if they don't have any sample after loading the WAL.

krasi-georgiev:
all tests passing and also updated the changelog

krasi-georgiev:
@gouthamve 	one final look would be good.

head.go
```diff
@@ -59,7 +59,8 @@ type Head struct {
 	appendPool sync.Pool
 	bytesPool  sync.Pool
 
-	minTime, maxTime int64
+	minTime, maxTime int64 // Current min and max of the samples included in the head.
+	minValidTime     int64 // Mint allowed to be added to te head. It shouldn't be lower than the last persisted block.
```
gouthamve: `te head` --> `the head`

gouthamve: `It shouldn't be lower than the last persisted block` --> `It shouldn't be lower than the max time of the last persisted block`

gouthamve: Also, I think we should be updating `h.minValidTime` in `h.Truncate` it doesn't make sense to accept samples that fall in the truncated period.

krasi-georgiev: yep logical!, thanks, updated!

PR #464, opened 2018-12-04
Release v0.3.0

krasi-georgiev:
@gouthamve 	what do you think about this changelog format and wording? 

I am trying to write it so that searching through the changelog file with keywords would also give context about the different changes. This way if a behaviour is removed at some point can also deprecate/refactor the corresponding API

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,8 @@
 ## master / unreleased
 
- - `LastCheckpoint()` used to return just the segment name and now it returns the full relative path.
- - `NewSegmentsRangeReader()` can now read over miltiple wal ranges by using the new `SegmentRange{}` struct.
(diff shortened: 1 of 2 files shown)


brian-brazil:
I think we should stick with the project's standard format. If someone needs this level of detail, they can always look at the commit history.

krasi-georgiev:
for a library it might be a bit different, consumers might want a quick overview of what has changed and what changes they might need in their code to update from one version to another etc. 
@gouthamve 	thoughts? since you are using tsdb as a library, would something like this be useful?

brian-brazil:
We follow the usual format in our other libraries, and they're more user facing than this one.

krasi-georgiev:
simplified.

brian-brazil:
:+1:

Issue #227, opened 2017-12-16
Compaction blocks Prometheus shutdown


SuperQ:
When Prometheus starts a SIGTERM shutdown any in-flight compaction operations will block the shutdown.

krasi-georgiev:
@SuperQ  Looking at the code this shouldn't happen. I am trying to replicate it.
Do you remember how did you run it and maybe a config file?

SuperQ:
This was for Prometheus 2.0, so it's possibly been fixed since then.

krasi-georgiev:
do you want to try again or shall we close this?

SuperQ:
I still need to upgrade our production servers from 2.0.0 to 2.2.1 to see if I can still reproduce the problem.

krasi-georgiev:
Thanks , let me know what happens , preferably with steps to reproduce.

krasi-georgiev:
@SuperQ had a chance to upgrade and test again?

SuperQ:
We'll have a large weekly compaction coming up in a couple of days, I'll do some tests then.

SuperQ:
It looks like this may still be an issue. 

```
level=info ts=2018-09-05T15:02:43.420947875Z caller=compact.go:352 component=tsdb msg="compact blocks" count=3 mint=1536127200000 maxt=1536148800000 ulid=01CPN41N7YWK55R50RR5X0DZV8 sources="[01CPMFBDMAFNTZFYPB5NPQQAQT 01CPMP74WA7PS9SDAG29Q6PQS5 01CPMX2W54JFS9BHR1A5T6MP2T]"
level=warn ts=2018-09-05T15:03:02.010643551Z caller=main.go:377 msg="Received SIGTERM, exiting gracefully..."
level=info ts=2018-09-05T15:03:02.010720913Z caller=main.go:402 msg="Stopping scrape discovery manager..."
level=info ts=2018-09-05T15:03:02.010782991Z caller=main.go:416 msg="Stopping notify discovery manager..."
level=info ts=2018-09-05T15:03:02.010804003Z caller=main.go:438 msg="Stopping scrape manager..."
level=info ts=2018-09-05T15:03:02.010860281Z caller=main.go:398 msg="Scrape discovery manager stopped"
level=info ts=2018-09-05T15:03:02.011451354Z caller=main.go:412 msg="Notify discovery manager stopped"
[...]

PR #465, opened 2018-12-05
use context to cancel compactions

krasi-georgiev:
closes https://github.com/prometheus/tsdb/issues/227


krasi-georgiev:
checking the context inside the series loop seems to have a very small impact on the overall compaction time  and if I am reading the numbers right for a block with 20000 series it will increase the time by 0.08sec 

```
func BenchmarkCompactions(t *testing.B) {
	tmpdir, err := ioutil.TempDir("", "test")
	testutil.Ok(t, err)
	defer os.RemoveAll(tmpdir)

	for i := 0; i < t.N; i++ {
		createPopulatedBlock(t, tmpdir, 20000, 0, 1)
	}
}

go test -benchmem -run=^$ github.com/prometheus/tsdb -bench ^BenchmarkCompactions$ -v  -count=1 -benchtime=1m > no-context.txt

go test -benchmem -run=^$ github.com/prometheus/tsdb -bench ^BenchmarkCompactions$ -v  -count=1 -benchtime=1m > with-context.txt

benchcmp no-context.txt with-context.txt
benchmark                  old ns/op     new ns/op     delta
BenchmarkCompactions-8     705443374     784730260     +11.24%

benchmark                  old allocs     new allocs     delta
BenchmarkCompactions-8     832343         832353         +0.00%

benchmark                  old bytes     new bytes     delta
BenchmarkCompactions-8     271110513     271112374     +0.00%
```

krasi-georgiev:
I don't like that I have to add this new metric `populatingBlocks prometheus.Counter` just for the test, but couldn't figure another way to detect that a compaction has  actually started.

@gouthamve any other ideas to avoid this new metric or make it more useful for the actual library usage?

compact_test.go
```diff
@@ -747,34 +748,70 @@ func TestDisableAutoCompactions(t *testing.T) {
...
+	createTestDb := func() (*DB, func()) {
+		tmpdir, err := ioutil.TempDir("", "testCancelCompaction")
+		testutil.Ok(t, err)
 
-	// Create some blocks to fall within the compaction range.
-	createBlock(t, tmpdir, 3000, 0, 1000)
-	createBlock(t, tmpdir, 3000, 1000, 2000)
-	createBlock(t, tmpdir, 1, 2000, 2001) // The most recent block is ignored so can be e small one.
+		// Create some blocks to fall within the compaction range.
+		createBlock(t, tmpdir, 4000, 0, 1000)
```
bwplotka: Just curious why change in num of series `3000` -> `4000` ?

krasi-georgiev:
the logic is to first check how long it takes the full compaction cycle and than test again when cancelled. This way it should be resistant to CI hiccups, don't you think?
I don't really  like mocking things. If I can't make it pass 10 times in a row or it becomes flaky at some point I will think again. 

> Also where context is used in this scenario (as title suggest). It's all about some channels, there is no context used.. or am I blind?

https://github.com/prometheus/tsdb/pull/465/files#diff-43297cfd6441b52c54c50d646a564339R144

compact_test.go
krasi-georgiev: gives bigger difference between normal run vs cancelled so it is less likely to false fail.

bwplotka:
Thx for context reference. Looks like a new thing, was looking in older code (: 

> the logic is to first check how long it takes the full compaction cycle and than test again when cancelled which should take less than half the time.  This way it should be resistant to CI hiccups,

No it is not resistant. Anything time related is not resistant.

1. You do full compaction cycle. CI hiccups happens. It took 2m
2. Let's say you have a bug that ignores context cancel. So you do second iteration with interrupt. Because of bug that ignores cancellation, compactor is doing full compaction. There was no CI hiccups so it was faster. It took 5s.

Our test here effectively succeded even though bug was in code. So it gives us 0 confidence. 

So..  Either we do mock even if you don't like this or this test can be deleted. But again, it's only my opinion after some expierience (: 

> If I can't make it pass 10 times in a row or it becomes flaky at some point I will think again.

I can tell you that it will be flaky looking on this code, so I save your time. (:

krasi-georgiev:
30 re-triggers and not a single failure. Seems good to go. 
If it does happen to cause some issues can always revisit or remove.

compact.go
```diff
@@ -471,13 +481,19 @@ func (w *instrumentedChunkWriter) WriteChunks(chunks ...chunks.Meta) error {
...
 // write creates a new block that is the union of the provided blocks into dir.
 // It cleans up all files of the old blocks after completing successfully.
 func (c *LeveledCompactor) write(dest string, meta *BlockMeta, blocks ...BlockReader) (err error) {
 	dir := filepath.Join(dest, meta.ULID.String())
 	tmp := dir + ".tmp"
-
+	var writers []io.Closer
 	defer func(t time.Time) {
+		for _, w := range writers {
+			w.Close()
```
bwplotka: Please do not ignore errors, especially on this path. 

We need something more complex like we have in Thanos:  https://github.com/improbable-eng/thanos/blob/master/pkg/runutil/runutil.go#L112

(Not only log line, but actually adding some multierror to `err` return variable, even in defer.

This is important because write might be not successful if index close fails etc...

compact.go
```diff
@@ -520,20 +536,33 @@ func (c *LeveledCompactor) write(dest string, meta *BlockMeta, blocks ...BlockRe
...
+	writers = append(writers, indexw)
 
 	if err := c.populateBlock(blocks, meta, indexw, chunkw); err != nil {
 		return errors.Wrap(err, "write compaction")
 	}
+
+	// Compaction was canceled so remove tmp folders and return early.
+	select {
+	case <-c.ctx.Done():
+		for _, w := range writers {
```
bwplotka: Hm.. this is something we need to improve. There is literally no need to close those twice. It is just enough to `return ErrCompactionCanceled` here and have nice defer that will always close & remove tmp.

compact.go
```diff
@@ -520,20 +536,33 @@ func (c *LeveledCompactor) write(dest string, meta *BlockMeta, blocks ...BlockRe
...
 	// you cannot delete these unless they are closed and the defer is to
 	// make sure they are closed if the function exits due to an error above.
-	if err = chunkw.Close(); err != nil {
-		return errors.Wrap(err, "close chunk writer")
-	}
-	if err = indexw.Close(); err != nil {
-		return errors.Wrap(err, "close index writer")
+	for _, w := range writers {
+		if err := w.Close(); err != nil {
+			return err
```
bwplotka: We need `multierr` here ideally - same for the closing in `defer`

bwplotka: And then we can always empty `writers` slice to not close twice

compact.go
bwplotka: And I know it is no regression, but if we are fixing this let's fix it right. Not saying we need helper for this like Thanos have, but helper would actually makes the code bit simpler (:

compact.go
```diff
@@ -593,9 +622,20 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
...
+		closeAll(closers...)
+		c.metrics.populatingBlocks.Add(-1)
+	}()
+
+	c.metrics.populatingBlocks.Inc()
 
 	for i, b := range blocks {
+		select {
+		case <-c.ctx.Done():
+			return nil
```
bwplotka: `ErrCompactionCanceled` for consistency?

compact.go
```diff
@@ -652,6 +692,12 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
 	}
 
 	for set.Next() {
+		select {
+		case <-c.ctx.Done():
+			return nil
```
bwplotka: ditto? The common practice is to return `ctx.Err` sometimes as well. It is easy then to filter out later on if you want

bwplotka: In our case `ErrCompactionCanceled` would do I guess.

compact_test.go
```diff
@@ -741,3 +744,74 @@ func TestDisableAutoCompactions(t *testing.T) {
 	}
 	testutil.Assert(t, len(db.Blocks()) > 0, "No block was persisted after the set timeout.")
 }
+
+// TestCancelCompactions ensures that when the db is closed
+// any running compaction is cancelled to unblock closing the db.
+func TestCancelCompactions(t *testing.T) {
```
bwplotka: Not a fan of this test as I mentioned before, but we can fix it in later PR (sooner than we want - there is high chance it will flake soon I am sure) (:

db.go
```diff
@@ -271,10 +275,13 @@ func Open(dir string, l log.Logger, r prometheus.Registerer, opts *Options) (db
 		db.lockf = lockf
 	}
 
-	db.compactor, err = NewLeveledCompactor(r, l, opts.BlockRanges, db.chunkPool)
+	ctx, cnl := context.WithCancel(context.Background())
```
bwplotka: we do `cancel` everywhere, so let's stick to that maybe?

db.go
```diff
@@ -126,6 +127,9 @@ type DB struct {
 	// changing the autoCompact var.
 	autoCompactMtx sync.Mutex
 	autoCompact    bool
+
+	// Cancel a running compaction when a shutdown is initiated.
+	compactCnl func()
```
bwplotka: What about moving to something all use in Golang:

```suggestion
	compactCancel context.CancelFunc
```

krasi-georgiev: updated.

db.go
krasi-georgiev: updated

compact.go
krasi-georgiev: yes this was bugging me as well and I think I managed to implement some nice err handling.

compact.go
krasi-georgiev: updated.

compact.go
krasi-georgiev: very good suggestion , thanks!

compact.go
krasi-georgiev: I refactored to use `ctx.Err` , much better, thanks!

krasi-georgiev:
> > 30 re-triggers and not a single failure. Seems good to go.
> > This does not SGTM. This tests will flake no matter what, and you can test it thousands of times on your fast laptop, but it will flake on slow CI anytime soon. (:

Will buy you a beer every-time it flakes :) if we reach 10 beers It is a good enough motivation for me that I remove the test or find another way :smile: 

I actually run it 30 times over a period of 2 days on Travis itself, but yeah if it ends up causing problems will revisit.

compact_test.go
bwplotka: hahaha. DEAL!

compact.go
```diff
@@ -476,9 +486,14 @@ func (w *instrumentedChunkWriter) WriteChunks(chunks ...chunks.Meta) error {
...
 	defer func(t time.Time) {
-		if err != nil {
+		merr.Add(err)
+		for _, w := range writers {
+			merr.Add(w.Close())
+		}
+		if merr.Err() != nil {
 			c.metrics.failed.Inc()
 			// TODO(gouthamve): Handle error how?
 			if err := os.RemoveAll(tmp); err != nil {
```
bwplotka: hm.. we don't want to remove tmp anyway?

compact.go
```diff
@@ -520,20 +535,28 @@ func (c *LeveledCompactor) write(dest string, meta *BlockMeta, blocks ...BlockRe
...
 	// you cannot delete these unless they are closed and the defer is to
 	// make sure they are closed if the function exits due to an error above.
-	if err = chunkw.Close(); err != nil {
-		return errors.Wrap(err, "close chunk writer")
+	for _, w := range writers {
+		merr.Add(w.Close())
 	}
-	if err = indexw.Close(); err != nil {
-		return errors.Wrap(err, "close index writer")
+	writers = writers[:0] // Avoid closing the writers twice in the defer.
```
bwplotka: :+1:

compact.go
```diff
@@ -476,9 +486,14 @@ func (w *instrumentedChunkWriter) WriteChunks(chunks ...chunks.Meta) error {
 func (c *LeveledCompactor) write(dest string, meta *BlockMeta, blocks ...BlockReader) (err error) {
 	dir := filepath.Join(dest, meta.ULID.String())
 	tmp := dir + ".tmp"
-
+	var writers []io.Closer
+	var merr MultiError
 	defer func(t time.Time) {
-		if err != nil {
+		merr.Add(err)
```
bwplotka: We need different MultiError here (: as your sometimes return `merr.Err` which would we added here then.

compact.go
krasi-georgiev: only when there is an error otherwise if everything was ok we don't have a tmp folder as this is renamed in `renameFile(tmp, dir);`

krasi-georgiev: I guess we can still run it as RemoveAll does nothing if the folder doesn't exist

krasi-georgiev: refactored

compact.go
```diff
@@ -476,14 +486,20 @@ func (w *instrumentedChunkWriter) WriteChunks(chunks ...chunks.Meta) error {
...
+		for _, w := range writers {
+			merr.Add(w.Close())
+		}
+
+		// RemoveAll returns no error when tmp doesn't exist so it is safe to always run it.
+		if err := os.RemoveAll(tmp); err != nil {
+			level.Error(c.logger).Log("msg", "removed tmp folder after failed compaction", "err", err.Error())
+		}
+		if merr.Err() != nil {
```
bwplotka: In this case we need to do `err = merr.Err()`

compact.go
```diff
@@ -593,9 +616,20 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
 		allSymbols = make(map[string]struct{}, 1<<16)
 		closers    = []io.Closer{}
 	)
-	defer func() { closeAll(closers...) }()
+	defer func() {
+		closeAll(closers...)
+		c.metrics.populatingBlocks.Add(-1)
```
codesome: Using `Set` instead of `Add` sounds better. Why play with floats like this?

compact.go
```diff
@@ -593,9 +616,20 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
...
+	}()
+
+	c.metrics.populatingBlocks.Inc()
```
codesome: As the description says set to 1, I would prefer explicitly using the `Set` function.

compact.go
```diff
@@ -476,14 +486,21 @@ func (w *instrumentedChunkWriter) WriteChunks(chunks ...chunks.Meta) error {
 func (c *LeveledCompactor) write(dest string, meta *BlockMeta, blocks ...BlockReader) (err error) {
 	dir := filepath.Join(dest, meta.ULID.String())
 	tmp := dir + ".tmp"
-
+	var writers []io.Closer
```
codesome: `writers` -> `closers` (to keep the name consistent across functions).

compact_test.go
```diff
@@ -741,3 +744,74 @@ func TestDisableAutoCompactions(t *testing.T) {
...
+	// Measure the compaction time without interupting it.
+	var timeCompactionUninterrupted time.Duration
+	{
+		db, err := Open(tmpdir, log.NewNopLogger(), nil, &Options{BlockRanges: []int64{1, 2000}})
+		testutil.Ok(t, err)
+		testutil.Equals(t, 3, len(db.Blocks()), "initial block count mismatch")
+		testutil.Equals(t, 0.0, prom_testutil.ToFloat64(db.compactor.(*LeveledCompactor).metrics.ran), "initial compaction counter mismatch")
+		db.compactc <- struct{}{} // Trigger a compaction.
+		var start time.Time
+		for {
```
codesome: We can reduce nesting and simplify this to the following for better readability.
```
for prom_testutil.ToFloat64(db.compactor.(*LeveledCompactor).metrics.populatingBlocks) <= 0 {
    time.Sleep(3 * time.Millisecond)
}
start = time.Now()
```

Also the below 2 `for` loops which are in the same fashion.

compact_test.go
```diff
@@ -741,3 +744,74 @@ func TestDisableAutoCompactions(t *testing.T) {
...
+	{
+		db, err := Open(tmpdirCopy, log.NewNopLogger(), nil, &Options{BlockRanges: []int64{1, 2000}})
+		testutil.Ok(t, err)
+		testutil.Equals(t, 3, len(db.Blocks()), "initial block count mismatch")
+		testutil.Equals(t, 0.0, prom_testutil.ToFloat64(db.compactor.(*LeveledCompactor).metrics.ran), "initial compaction counter mismatch")
+		db.compactc <- struct{}{} // Trigger a compaction.
+		dbClosed := make(chan struct{})
+		for {
+			if prom_testutil.ToFloat64(db.compactor.(*LeveledCompactor).metrics.populatingBlocks) > 0 {
+				time.Sleep(3 * time.Millisecond)
```
codesome: Why are we sleeping here?

compact_test.go
```diff
@@ -741,3 +744,74 @@ func TestDisableAutoCompactions(t *testing.T) {
...
+			if prom_testutil.ToFloat64(db.compactor.(*LeveledCompactor).metrics.populatingBlocks) > 0 {
+				start = time.Now()
+				break
+			}
+			time.Sleep(3 * time.Millisecond)
+		}
+
+		for {
+			if prom_testutil.ToFloat64(db.compactor.(*LeveledCompactor).metrics.ran) == 1 {
+				timeCompactionUninterrupted = time.Since(start)
```
codesome: What was the value that you observed for `timeCompactionUninterrupted` during tests?

compact_test.go
```diff
@@ -741,3 +744,74 @@ func TestDisableAutoCompactions(t *testing.T) {
...
+				}()
+				break
+			}
+		}
+
+		start := time.Now()
+		<-dbClosed
+		actT := time.Since(start)
+		expT := time.Duration(timeCompactionUninterrupted / 2) // Closing the db in the middle of compaction should less than half the time.
+		testutil.Assert(t, actT < expT, "closing the db took more than expected. exp: <%v, act: %v", expT, actT)
```
codesome: For both `actT` and `expT`, I am not sure if a single run of compaction would be fine. How about running compaction multiple times with and without closing DB, and compare their averages instead?

codesome:
(https://github.com/prometheus/tsdb/pull/465#discussion_r254735655 is not a blocker)

compact_test.go
krasi-georgiev: updated, thanks

compact_test.go
krasi-georgiev: just to make it use less CPU

compact_test.go
krasi-georgiev: it is different on different machines and platforms, but in general is is over 300% difference so plenty of difference compared to the cancellation.

compact_test.go
krasi-georgiev: Yeap, not a bad idea.
For now I prefer to keep it as is to keep the tests as short as possible and if it ends up causing problems will revisit.

krasi-georgiev:
> Did you intend to put this sleep outside the `if` statement and inside the `for`?

yep , updated!

> (PS: this loop can be simplified like the above 2 loops which I commented on before)

don't see how can we use the same logic here as it will start a new gorotine on every loop

codesome:
> don't see how can we use the same logic here as it will start a new gorotine on every loop

```go
for prom_testutil.ToFloat64(db.compactor.(*LeveledCompactor).metrics.populatingBlocks) <= 0 { 
    time.Sleep(3 * time.Millisecond)
}
go func() {
    testutil.Ok(t, db.Close())
    close(dbClosed)
}()
```

krasi-georgiev:
yep, not my day today. Thanks!
updated.

compact.go
```diff
@@ -476,14 +486,21 @@ func (w *instrumentedChunkWriter) WriteChunks(chunks ...chunks.Meta) error {
 func (c *LeveledCompactor) write(dest string, meta *BlockMeta, blocks ...BlockReader) (err error) {
 	dir := filepath.Join(dest, meta.ULID.String())
 	tmp := dir + ".tmp"
-
+	var closers []io.Closer
 	defer func(t time.Time) {
+		var merr MultiError
+		merr.Add(err)
+		err = merr.Err()
```
codesome: This should come after the `for` loop below. Or rather, we can reuse `closeAll` here.

compact.go
```diff
@@ -593,9 +616,20 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
 		allSymbols = make(map[string]struct{}, 1<<16)
 		closers    = []io.Closer{}
 	)
-	defer func() { closeAll(closers...) }()
+	defer func() {
+		closeAll(closers...)
```
codesome: I think we should not ignore the error here. Either log or return by binding it to the return variables.

compact.go
```diff
@@ -520,27 +537,33 @@ func (c *LeveledCompactor) write(dest string, meta *BlockMeta, blocks ...BlockRe
...
+	}
+
 	// We are explicitly closing them here to check for error even
 	// though these are covered under defer. This is because in Windows,
 	// you cannot delete these unless they are closed and the defer is to
 	// make sure they are closed if the function exits due to an error above.
-	if err = chunkw.Close(); err != nil {
-		return errors.Wrap(err, "close chunk writer")
+	var merr MultiError
+	for _, w := range closers {
```
codesome: We can reuse `closeAll` here.

compact.go
krasi-georgiev: updated, thanks!

compact.go
krasi-georgiev: updated, thanks!

compact.go
krasi-georgiev: updated, thanks!

PR #468, opened 2018-12-10
remove unused WALFlushInterval option and NopWAL struct

krasi-georgiev:
The `WALFlushInterval` is not used anywhere in the code base.
The `WAL` is not an interface anymore to save some lookup time so can't use `NopWAL` in the tests. Instead can just pass `nil` as the code checks for that and it is essentially a `noop`.


brian-brazil:
:+1:

PR #469, opened 2018-12-11
fix windows tests

krasi-georgiev:


krasi-georgiev:
still green even after 10 re-triggers.

wal_test.go
```diff
@@ -14,427 +14,17 @@
...
-
-	r.Read(func(s []RefSeries) {
-		readSeries = append(readSeries, s...)
-	}, nil, nil)
-
-	testutil.Equals(t, expected, readSeries)
-}
-
-// Symmetrical test of reading and writing to the WAL via its main interface.
-func TestSegmentWAL_Log_Restore(t *testing.T) {
```
gouthamve: Why is this test deleted? Even though we don't log through the this WAL anymore, we'll have to support reading from it till 3.x atleast....

head.go
```diff
@@ -480,7 +480,7 @@ func (h *Head) Init(minValidTime int64) error {
 		return nil
 	}
 	level.Warn(h.logger).Log("msg", "encountered WAL error, attempting repair", "err", err)
-
+	sr.Close() // Close the reader so that the repair can remove the corrupted file under Windows.
```
simonpasquier: _My 2 cents from someone who's not so familiar with the tsdb code_

From an external view, having `defer sr.Close()` at L476 and `sr.Close()` at L483 feels a bit strange though I get that calling `Close()` isn't a problem per se. Would this work instead?

```go
	sr, err := wal.NewSegmentsRangeReader(wal.SegmentRange{Dir: h.wal.Dir(), First: startFrom, Last: -1})
	if err != nil {
		return errors.Wrap(err, "open WAL segments")
	}

	err = h.loadWAL(wal.NewReader(sr))
        sr.Close() // Close the reader so that the repair can remove the corrupted file under Windows.
	if err == nil {
		return nil
        }
```

And shouldn't the segment reader opened at L457 be closed ASAP instead of `defer`?

krasi-georgiev: yep that would work as well, Thanks!

wal_test.go
krasi-georgiev: re-added, one of these was failing that is why I removed them, but lucky enough it wasn't this one.

wal_test.go
```diff
@@ -265,176 +145,6 @@ func TestSegmentWAL_Log_Restore(t *testing.T) {
 	}
 }
 
-func TestWALRestoreCorrupted_invalidSegment(t *testing.T) {
```
gouthamve: Hate to say this but these two tests (this and one after) also need to be supported :/ Other than that, LGTM!

krasi-georgiev:
@gouthamve 	I have reverted the removed tests and disabled these for windows `// +build !windows`. 
I think there is an actual bug in the old WAL, but this is as far as my motivation to fix this goes.

Issue #471, opened 2018-12-11
prometheus_tsdb_wal_corruptions_total is missing in new wal implementation


dkalashnik:
After updating to prometheus 2.5 I miss prometheus_tsdb_wal_corruptions_total metric.
Looks like it is not available in the new wal: 
https://github.com/prometheus/tsdb/blob/master/wal/wal.go#L200-L219

simonpasquier:
Good catch Dima!

PR #473, opened 2018-12-12
re-add the missing prometheus_tsdb_wal_corruptions_total

krasi-georgiev:
closes https://github.com/prometheus/tsdb/issues/471

after implementing the new WAL this metric was missing so adding it
again.
Also added it in a test to make sure it works as expected.


krasi-georgiev:
@codesome 	mind having a quick look?

codesome:
Bit occupied in KubeCon and AFK. Will take a look in couple of days.

head.go
```diff
@@ -480,10 +486,10 @@ func (h *Head) Init(minValidTime int64) error {
 		return nil
 	}
 	level.Warn(h.logger).Log("msg", "encountered WAL error, attempting repair", "err", err)
+	h.metrics.walCorruptionsTotal.Inc()
```
codesome: I believe detect WAL corruptions when we call `loadWAL`. So do we also need to increment it for this: https://github.com/prometheus/tsdb/blob/9e51d56e08958f22f55daf26795ee477def7797e/head.go#L471-L473

And also maybe a small test for that?

krasi-georgiev: When this happens Prometheus will exist, so what is the point of an increment there?

codesome: If there is no way to recover this `Inc()` info after we return, then there is no need of adding it here.

simonpasquier: Maybe add the metrics directly to `wal.Repair()` so we can know when there is a corruption and whether or not it has been repaired?

krasi-georgiev:
@dkalashnik how are you using the `prometheus_tsdb_wal_corruptions_total` metric? 
asking so that I know if the current implementation would work for your use case.

dkalashnik:
@krasi-georgiev We are using it as a part of the generic prometheus dashboard in grafana, so no specific case.

krasi-georgiev:
@dkalashnik  how is your alerting defined based on this metric?

dkalashnik:
@krasi-georgiev We don't have alerts based on that metric, just a panel in dashboard

PR #111, opened 2017-07-21
Add prefix label matcher.

dilyevsky:
Implement labels.PrefixMatcher and use interface conversion in querier
to optimize label tuples search.

[unit-tests]: Fix bug and populate label index for mock index.

Signed-off-by: Dmitry Ilyevsky <ilyevsky@gmail.com>

PR #474, opened 2018-12-12
remove unused `PrefixMatcher`

krasi-georgiev:
Brian mentioned that is probably not used by anyone so just a cleanup 

@gouthamve that was one of your firs PR reviews, remember the times :smiley_cat:  !
https://github.com/prometheus/tsdb/pull/111

### DIFF
--- a/labels/selector.go
+++ b/labels/selector.go
@@ -15,7 +15,6 @@ package labels
 
 import (
 	"regexp"
-	"strings"
 )
 
 // Selector holds constraints for matching against a label set.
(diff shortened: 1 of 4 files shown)


labels/selector.go
```diff
@@ -104,17 +103,3 @@ func Not(m Matcher) Matcher {
 type PrefixMatcher struct {
```
codesome: Should this also be removed?

brian-brazil:
:+1:

PR #479, opened 2018-12-14
fix race for minValidTime

krasi-georgiev:
it happens when truncating the WAL and another goroutine creates a new
`Appender()`

can be replicated with
```
go test -benchmem -run=^$ github.com/prometheus/prometheus/promql -bench . -v --race
```


brian-brazil:
:+1:

PR #482, opened 2018-12-19
fix TestWALSegmentSizeOption for  windows.

krasi-georgiev:

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -707,7 +707,9 @@ func TestWALSegmentSizeOption(t *testing.T) {
 		testutil.Ok(t, app.Commit())
 	}
 
-	files, err := ioutil.ReadDir(filepath.Join(db.Dir(), "wal"))
+	dbDir := db.Dir()
+	db.Close()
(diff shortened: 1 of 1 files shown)


db_test.go
```diff
@@ -707,7 +707,9 @@ func TestWALSegmentSizeOption(t *testing.T) {
 		testutil.Ok(t, app.Commit())
 	}
 
-	files, err := ioutil.ReadDir(filepath.Join(db.Dir(), "wal"))
+	dbDir := db.Dir()
+	db.Close()
```
brian-brazil: This will close on every iteration, so I'd expect the 2nd iteration to fail.

krasi-georgiev: sorry I didn't understand. There is no loop here.

brian-brazil: Ah, indentation was showing oddly on the for above.

krasi-georgiev: yeah, I was just going to comment on that as well. It is all good the tests are passing

PR #489, opened 2018-12-29
refactor createBlock to be more generic

krasi-georgiev:
It allows generating a block with unlimited series count with or without
high cardinality and churn.


krasi-georgiev:
cc @codesome

PR #491, opened 2019-01-02
Add CircleCI releases

SuperQ:
Add CircleCI + Promu to release cmd/tsdb.

Closes: https://github.com/prometheus/tsdb/issues/484

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,79 @@
+---
+version: 2.1
+
+executors:
+  # Whenever the Go version is updated here, .travis.yml and .promu.yml
+  # should also be updated.
+  golang:
+    docker:
+    - image: circleci/golang:1.11
+
+jobs:
+  test:
+    executor: golang
+
+    steps:
+    - checkout
+    - run: make promu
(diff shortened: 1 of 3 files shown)


krasi-georgiev:
since we use travis for the tests could we use to for the build as well?

SuperQ:
Possibly, but we're using circleci for builds everywhere else. Picking a single CI is already a topic of discussion.

krasi-georgiev:
yeah I know and I think the final opinion of @simonpasquier was to slowly migrate to travis. I think the build here is pretty small so maybe try it travis and if it ends up being a pain will use circle again but only for the builds. no need to run the same tests with travis and circle.

SuperQ:
Ok, I guess we need to learn how to use multiple jobs/releases, remote docker, etc in Travis.

krasi-georgiev:
seems that the circle build is triggered for all PRs and it fails as it doesn't find any config. Maybe enabled somewhere globally?

SuperQ:
It's a GitHub integration, I will remove it.

PR #492, opened 2019-01-02
Exposed helper methods for reading index bytes.

bwplotka:
**Changes**:
* ReadSymbols, TOC and ReadOffsetTable are not public functions (used by Thanos).
* decbufXXX are now functions.
* More verbose errors.
* Removed unused crc32 field.
* Some var name changes to make it more verbose:
  * symbols -> allocatedSymbols
  * symbolsSlice -> symbolsV1
  * symbols -> symbolsV2
  *
* Pre-calculate symbolsTableSize.
* Initialized symbols for Symbols() method with valid length.
* Added test for Symbol method.
* Made Decoder LookupSymbol method public. Kept Decode public as it is useful as helper from index package.

### DIFF
--- a/index/encoding_helpers.go
+++ b/index/encoding_helpers.go
@@ -18,6 +18,8 @@ import (
 	"hash"
 	"hash/crc32"
 	"unsafe"
+
+	"github.com/pkg/errors"
 )
 
 // enbuf is a helper type to populate a byte slice with various types.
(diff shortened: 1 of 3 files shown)


bwplotka:
Circle CI does not look right... I guees it is some migration period to CircleCI and it is being set up, right?

brian-brazil:
That was an internal method that should never have existed (it was a layering violation), nor are there any public methods that take offsets. I don't think this makes sense as part of the API.

> Made Decoder LookupSymbol method public, otherwise there is no point in having Decoder struct public.

Decoder should probably be private.

bwplotka:
Thanks for taking a look @brian-brazil ! (:  I can see totally your concerns with limiting API, but we have IMO valid use case: Thanos as consumer of TSDB library wants to be able use helpers for decoding index bytes:
* `index.Decoder` 
* `index.Reader.symbolsSlice` (or `symbols` depends on index version) 

This is to be able to implement index cache, so storing part of index (version, symbol, values, posting index) in different place than rest (actual postings and series). Which allows us to use S3 API to get_range bytes we need from index. 

The truth is that TSDB format (including index format) is an API itself, so why not exposing useful helper methods for accessing it?

I agree that on `IndexReader` interface you don't want to expose offsets (layering violation), that's why I did not add this method to interface. However, exposing **helpers** for decoding index bytes on lower level for the **reuse** purpose is healthy both for Thanos and TSDB (we basically maintain same piece of code!). (: 

If you don't agree with reusability pattern, do you have any suggestion how in Thanos we should overcome lack of such helpers?

bwplotka:
Sound Good, will give it a try.

bwplotka:
PTAL @brian-brazil . Updated description with summary of changes.

index/encoding_helpers.go
```diff
@@ -86,6 +88,60 @@ type decbuf struct {
 	e error
 }
 
+// newDecbufAt returns a new decoding buffer. It expects the first 4 bytes
+// after offset to hold the big endian encoded content length, followed by the contents and the expected
+// checksum.
+func newDecbufAt(bs ByteSlice, off int) decbuf {
+	if bs.Len() < off+4 {
```
bwplotka: straight copy of `r.newDecbufAt` just with `ByteSlice` passed by argument.

index/encoding_helpers.go
```diff
@@ -86,6 +88,60 @@ type decbuf struct {
...
+	if exp := binary.BigEndian.Uint32(b[len(b)-4:]); dec.crc32() != exp {
+		return decbuf{e: errInvalidChecksum}
+	}
+	return dec
+}
+
+// decbufUvarintAt returns a new decoding buffer. It expects the first bytes
+// after offset to hold the uvarint-encoded buffers length, followed by the contents and the expected
+// checksum.
+func newDecbufUvarintAt(bs ByteSlice, off int) decbuf {
```
bwplotka: same here

index/index.go
```diff
@@ -857,27 +813,20 @@ func (r *Reader) lookupSymbol(o uint32) (string, error) {
...
 		res[s] = struct{}{}
 	}
-	for _, s := range r.symbolSlice {
+	for _, s := range r.symbolsV2 {
 		res[s] = struct{}{}
 	}
 	return res, nil
 }
 
 // SymbolTableSize returns the symbol table that is used to resolve symbol references.
```
brian-brazil: This comment is out of date

index/index.go
```diff
@@ -129,13 +134,42 @@ type Writer struct {
...
 }
 
-type indexTOC struct {
-	symbols           uint64
-	series            uint64
-	labelIndices      uint64
-	labelIndicesTable uint64
-	postings          uint64
-	postingsTable     uint64
+// TOC represents index Table Of Content that states were each section of index starts.
```
brian-brazil: where

bwplotka:
Comments addressed.

bwplotka:
kind PTAL @brian-brazil

brian-brazil:
Thanks!

PR #494, opened 2019-01-10
Missing the len of crc32 when calculating maxLen in WriteChunks

naivewong:


gouthamve:
@naivewong The change looks good to me, thanks! You need to sign off on the commit before this can be merged though. See: https://github.com/prometheus/prometheus/wiki/DCO-signing

naivewong:
@gouthamve Thanks! And is now ok?

PR #495, opened 2019-01-11
Live reader another idea

krasi-georgiev:
refactor of https://github.com/prometheus/tsdb/pull/481

cc @cstyan here is a quick idea I had for the live reader.

still has few bugs and no optimisations with any buffers but will do these after the weekend.

### DIFF
--- a/wal/wal.go
+++ b/wal/wal.go
@@ -832,28 +832,13 @@ func (r *Reader) next() (err error) {
 		}
 		r.rec = append(r.rec, buf[:length]...)
 
-		switch r.curRecTyp {
-		case recFull:
-			if i != 0 {
-				return errors.New("unexpected full record")
-			}
-			return nil
-		case recFirst:
-			if i != 0 {
-				return errors.New("unexpected first record")
-			}
-		case recMiddle:
-			if i == 0 {
-				return errors.New("unexpected middle record")
-			}
-		case recLast:
-			if i == 0 {
-				return errors.New("unexpected last record")
-			}
-			return nil
-		default:
-			return errors.Errorf("unexpected record type %d", r.curRecTyp)
+		if err := validateRecord(r.curRecTyp, i); err != nil {
+			return err
 		}
+		if r.curRecTyp == recLast || r.curRecTyp == recFull {
+			return nil
+		}
+
 		// Only increment i for non-zero records since we use it
 		// to determine valid content record sequences.
 		i++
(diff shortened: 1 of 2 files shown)


krasi-georgiev:
> Curious what totalSize in NewLiveReader(r, totalSize) should be. Is it going to be segmentSize(like max segment size most of the time? 
--

yes `totalSize` is keep the api to a minimal and avoid having the `TotalRead` api.

> If yes, how do we handle this case from the format docs:
> > If a single record exceeds the default segment size, a segment with a larger size will be created.

Hm. Are you sure if this is still valid? 


> Also, we need to add some cancellation logic to this where Next() doesn't block forever. I am not sure this will be simpler than the other approach once cancellation is added, specially when using this in Prometheus, things are bound to get more complex.

my idea is that the cancellation logic should be at the consumer side with a select block.

gouthamve:
> Hm. Are you sure if this is still valid?

Yes, might not hit the case with Prometheus but if we're going to make it generic, then this is a possibility.

krasi-georgiev:
this use case can be covered with something like
```
// In case of an EOF error retry when:
   totalRead is less than the wal file size 
   or 
   the record length is greater than the wal file size
if length <= r.totalSize ||  r.totalRead == r.totalSize {
```

gouthamve:
Yes, but I don't think we will know the size of the segment when we create the `LiveReader`. The size will keep changing as new data is being appended to the segment if I'm not wrong.

krasi-georgiev:
now we have the `WALSegmentSize`
https://github.com/prometheus/tsdb/blob/bff5aa4d21113b52b7894d5a72e658a07aa2a77c/db.go#L47

cstyan:
The issue I still have with using `WALSegmentSize` is that, although it's not normal operations, frequent restarts or crashes can result in segments that are not full to the max size. For corruption reasons among others. For example, right now I have this on one of my instances:
```
drwxr-xr-x    3 root     root        4.0K Jan 15 03:18 .
drwxr-xr-x   27 root     root        4.0K Jan 15 03:01 ..
-rw-r--r--    1 root     root      128.0M Jan 15 00:00 00000265
-rw-r--r--    1 root     root      128.0M Jan 15 00:28 00000266
-rw-r--r--    1 root     root      128.0M Jan 15 00:56 00000267
-rw-r--r--    1 root     root      128.0M Jan 15 01:23 00000268
-rw-r--r--    1 root     root      128.0M Jan 15 01:51 00000269
-rw-r--r--    1 root     root      128.0M Jan 15 02:19 00000270
-rw-r--r--    1 root     root       14.7M Jan 15 02:52 00000271
-rw-r--r--    1 root     root       28.8M Jan 15 02:58 00000272
-rw-r--r--    1 root     root       33.9M Jan 15 03:06 00000273
-rw-r--r--    1 root     root       48.3M Jan 15 03:18 00000274
-rw-r--r--    1 root     root      112.8M Jan 15 03:42 00000275
```

krasi-georgiev:
yeah I don't mind using `os.Stat` for this, but maybe try to have this internally without the need to expose `TotalRead()`

PR #496, opened 2019-01-12
change variable name metrics to labels

yeya24:
In cmd/tsdb/main.go 
Some variables' name might be kind of confusing. So I change the **metrics** to **labels**
Signed-off-by: yeya24 <ben.ye@daocloud.io>


yeya24:
> That is an array of labels, so `metrics` sounds fine. Also `metrics` name is used in many different places for this.

Thanks for your reply!

krasi-georgiev:
yep I think labels is fine. Thanks

PR #498, opened 2019-01-15
Missing the length of the encoding byte when calling b.Range

naivewong:


chunks/chunks.go
```diff
@@ -367,7 +367,7 @@ func (s *Reader) Chunk(ref uint64) (chunkenc.Chunk, error) {
 	if n <= 0 {
 		return nil, errors.Errorf("reading chunk length failed with %d", n)
 	}
-	r = b.Range(off+n, off+n+int(l))
+	r = b.Range(off+n, off+n+1+int(l))
```
krasi-georgiev: I am not that familiar with this parts of the code, but isn't this accounted for in the following line : `r[1:1+l]` in `return s.pool.Get(chunkenc.Encoding(r[0]), r[1:1+l])`

bwplotka: Also, if that fixes some bug, can see the unit tests that fails without this and is fixed with this change? (:

naivewong: @krasi-georgiev As you can see in line `r = b.Range(off+n, off+n+int(l))`, **r** is a sub-slice of slice **b** with len **l**. But in the next line: `return s.pool.Get(chunkenc.Encoding(r[0]), r[1:1+l])`, you are 
 accessing the **r[1 : l + 1]**  and it has already exceeded the len **l** it should be. And the codes still work is because **r** is like a reference to **b** and **b** has a size larger than **l**, so **r** can access the range **[1 : l + 1]** without panic. And the **1** i want to add in `r = b.Range(off+n, off+1+n+int(l))` is the length of encoding byte as specified in **[tsdb/docs/format/chunks.md](https://github.com/prometheus/tsdb/blob/master/docs/format/chunks.md)**

And for a simple example, the following codes can run and print 10 without error:
`package main`
`import "fmt"`
`func main() {`
	`a := make([]int, 2)`
	`a[1] = 10`
	`b := a[:1]`
	`c := b[1:2]`
	`fmt.Println(c[0])`
`}`

krasi-georgiev:
@naivewong thanks for the explanation. The change is valid!

I opened a new PR in https://github.com/prometheus/tsdb/pull/508 to rename the vars to make this easier to follow.
@naivewong  since you are familiar with this code now I would appreciate a second look.

chunks/chunks.go
krasi-georgiev: @bwplotka it doesn't fix a bug, but if the golang implementation decided to change how are slices handled this might start panicking.

bwplotka: Thanks for explanation :+1:

Issue #500, opened 2019-01-18
Flaky test with the new live reader


gouthamve:
```
>> running all tests
GO111MODULE= go test -race  ./...
ok  	github.com/prometheus/tsdb	105.957s
ok  	github.com/prometheus/tsdb/chunkenc	1.037s
ok  	github.com/prometheus/tsdb/chunks	1.012s
?   	github.com/prometheus/tsdb/cmd/tsdb	[no test files]
ok  	github.com/prometheus/tsdb/fileutil	1.013s
ok  	github.com/prometheus/tsdb/index	77.752s
ok  	github.com/prometheus/tsdb/labels	1.010s
ok  	github.com/prometheus/tsdb/test	1.010s [no tests to run]
?   	github.com/prometheus/tsdb/testutil	[no test files]
ok  	github.com/prometheus/tsdb/tsdbutil	1.029s
wal_test.go:345: unexpected error: EOF
--- FAIL: TestWAL_FuzzWriteRead_Live (16.81s)
	wal_test.go:268: created dir:  /tmp/wal_fuzz_live882544920
FAIL
FAIL	github.com/prometheus/tsdb/wal	90.877s
make: *** [common-test] Error 1
The command "if [[ "$TRAVIS_OS_NAME" == "windows" ]]; then make test; else make; fi" exited with 2.
Done. Your build exited with 1.
````

@cstyan

krasi-georgiev:
this was under linux

cstyan:
@krasi-georgiev I'm assuming this isn't an issue anymore? I haven't see any failures for this recently.

I think we can close this, going to close my pr #502

krasi-georgiev:
I haven't seen any failures. Not sure what changed. 

Will reopen of it happens again.

PR #502, opened 2019-01-18
LiveReader tests should only fail for reader errors that are not EOF

cstyan:
this should fix flakes in the live reader tests

Fixes: #500 

### DIFF
--- a/wal/wal_test.go
+++ b/wal/wal_test.go
@@ -17,6 +17,7 @@ package wal
 import (
 	"bytes"
 	"fmt"
+	// "io"
 	"io/ioutil"
 	"math/rand"
 	"os"
@@ -25,6 +26,15 @@ import (
 
 	"github.com/prometheus/tsdb/testutil"
 )
+
+// We only want to fail the fuzz test if the reader
+// returns an error other than EOF, which we expect to get.
+// func checkErrorEOF(err error) error {
+// 	if err != io.EOF {
+// 		return err
+// 	}
+// 	return nil
+// }
 
 func TestWAL_Repair(t *testing.T) {
 	for name, test := range map[string]struct {


cstyan:
@krasi-georgiev @gouthamve I reran the tests a few time. This should at least stop the test from failing with the same `unexpected error: EOF`.

wal/wal_test.go
```diff
@@ -259,9 +268,7 @@ func TestWAL_FuzzWriteRead_Live(t *testing.T) {
 			lock.RUnlock()
 			index++
 		}
-		if r.Err() != io.EOF {
-			testutil.Ok(t, r.Err())
```
codesome: Did this fail before because after the `if` check above `r.Err()` would become `EOF` before it reached here?

cstyan: This particular `r.Err()` check was fine, I just made that `checkErrorEOF` function for each of the places we're checking the error for `LiveReader`. The issue here is that EOF is an error we expect to receive occasionally with the live reader, especially with the added slowness of running the tests in Travis.

cstyan:
bump @gouthamve @krasi-georgiev

bwplotka:
Sorry, I might miss the context of LiveReader, but I don't get why sometimes we have io.EOF sometimes not.

> . The issue here is that EOF is an error we expect to receive occasionally with the live reader, especially with the added slowness of running the tests in Travis.

Do we know @cstyan why is that the case? Does that slowness and seing unexpected EOF doesn't mean that basically test was stuck in the middle and failes to finish (e.g timeout?) If that is the case, just ignoring EOF is basically ignoring this test at all.

I think we should inspect LiveReader where this  unexpected io.EOF cames from. If this is expected thing in this test, we might want to make LiveReader return io.EOF as normal thing at the end and expect exactly io.EOF.

bwplotka:
OK, if it is an expected error, then let's assert `io.EOF` and fail if we have anything else (e.g no error even)

cstyan:
No error is not a failure case though.

bwplotka:
Why? We are in control of input and whole test flow, so why it is not deterministic? That's the key issue here.

krasi-georgiev:
would you mind giving some priority to this as it is causing flaky tests quite often now.

Issue #480, opened 2018-12-14
Memory issues on short block sizes


radeklesniewski:
##
**Bug Report**

We are testing Prometheus with Thanos to gather metrics from multiple Openshift clusters. On each we have significant amount of ephemeral pods, which are being scraped by Prometheus. We encounter strange memory consumption pattern during test that resembles our real time metrics almost 1:1. We are able to process millions of metrics when changes for particular label occurs every ~2 hours, while for the changes which are happening more often we see constant increase in memory consumption. We tried to lower retention time to 15 minutes, and block size down to 5 minutes, but this didn't help as a lot as we still were seeing the increase in memory consumption, even after the retention period. Further investigation has shown that short block sizes together with short retention caused degradation of performance of writing the blocks to disk. After 30 minutes of test the 5 minutes block was being flushed longer than 10 minutes, +/- 3m for head GC. 

Additionally, we are observing unexpected behaviour of retention policy execution. As far as we know after debugging Prometheus, there are two conditions for block to be deleted:
1. Block is beyond retention,
2. Block is not opened while dropping old blocks from memory occurs.

Example of unexpected behaviour: 
	Config: retention=15m, min-block-duration=5m, max-block-duration=5m
	Behaviour: old blocks are deleted not after ~20 minutes (retention + newest block duration), but after ~60 minutes (Prometheus instance is not queried while the test is performed, so probably there is no reason for block to be loaded in memory). Test is executed up to ~2 hours and during that time we are observing constant increase in memory consumption, even though after 1st hour of test number of blocks and data series is constant.

After locking label to change every 2 hours, we could easily use longer blocks and retention periods as memory consumption was stable. Memory consumption peaks at 16GB with average at 11GB for 3M metrics per minute, while for labels which were changing every minute for 1.5M after ~30m it crashed at ~100GB.

In both cases proportions between number of different metrics to number of blocks was constant, which raises a question, why is it behaving like that and how can we tweak the Prometheus to be able to perform better with high cardinality labels (e.g. cron job id / pod name). We are able to access and aggregate over such metrics in Thanos so Prometheus would need to be able to scrape and preserve data only for a time needed to flush block to disk.
### 
**What did you expect to see?**
We expected to see proportional relation between block sizes, number of different metrics (cardinality / data points), retention and memory consumption.
### 
**What did you see instead? Under which circumstances?**
Significant memory consumption increase for metrics with high cardinality / variable label
### 
**Environment**
Openshift v.3.9 with live data
[...]

PR #505, opened 2019-01-21
Additional logging in compact.go - logged time needed for writing blocks

radeklesniewski:
This PR includes additional logging about duration of func (c *LeveledCompactor) write(...) execution. PR related to https://github.com/prometheus/tsdb/issues/480.

### DIFF
--- a/compact.go
+++ b/compact.go
@@ -414,6 +414,8 @@ func (c *LeveledCompactor) Compact(dest string, dirs []string, open []*Block) (u
 }
 
 func (c *LeveledCompactor) Write(dest string, b BlockReader, mint, maxt int64, parent *BlockMeta) (ulid.ULID, error) {
+	start := time.Now()
+
 	entropy := rand.New(rand.NewSource(time.Now().UnixNano()))
 	uid := ulid.MustNew(ulid.Now(), entropy)
 
@@ -440,7 +442,7 @@ func (c *LeveledCompactor) Write(dest string, b BlockReader, mint, maxt int64, p
 		return ulid.ULID{}, nil
 	}
 
-	level.Info(c.logger).Log("msg", "write block", "mint", meta.MinTime, "maxt", meta.MaxTime, "ulid", meta.ULID)
+	level.Info(c.logger).Log("msg", "write block", "mint", meta.MinTime, "maxt", meta.MaxTime, "ulid", meta.ULID, "duration", time.Since(start))
 	return uid, nil
 }
 
--- a/compact.go
+++ b/compact.go
@@ -442,7 +442,13 @@ func (c *LeveledCompactor) Write(dest string, b BlockReader, mint, maxt int64, p
 		return ulid.ULID{}, nil
 	}
 
-	level.Info(c.logger).Log("msg", "write block", "mint", meta.MinTime, "maxt", meta.MaxTime, "ulid", meta.ULID, "duration", time.Since(start))
+	level.Info(c.logger).Log(
+		"msg", "write block",
+		"mint", meta.MinTime,
+		"maxt", meta.MaxTime,
+		"ulid", meta.ULID,
+		"duration", time.Since(start),
+	)
 	return uid, nil
 }
 


krasi-georgiev:
`Compact()` already have such logs in

https://github.com/prometheus/tsdb/blob/cc277e398b479bdb601e17e439bad11901824aa3/compact.go#L391-L399

why did you need these extra logs there?

radeklesniewski:
I didn't see that in logs, I'm not sure if it is logged when blocks have equal max & min duration and block compaction is not triggered.

krasi-georgiev:
this one shows during compaction and in your case compaction is disabled. 
In that you just need to add the time elapsed to:
https://github.com/prometheus/tsdb/blob/cc277e398b479bdb601e17e439bad11901824aa3/compact.go#L443


```
level.Info(c.logger).Log(
		"msg", "write block",
		"mint", meta.MinTime,
		"maxt", meta.MaxTime,
		"ulid", meta.ULID,
		"duration", time.Since(start),
	)
```

krasi-georgiev:
@radeklesniewski

compact.go
```diff
@@ -466,6 +466,10 @@ func (w *instrumentedChunkWriter) WriteChunks(chunks ...chunks.Meta) error {
 // write creates a new block that is the union of the provided blocks into dir.
 // It cleans up all files of the old blocks after completing successfully.
 func (c *LeveledCompactor) write(dest string, meta *BlockMeta, blocks ...BlockReader) (err error) {
+	level.Info(c.logger).Log("msg", "writing blocks to disk")
```
bwplotka: It might into `debug` more?

compact.go
```diff
@@ -570,6 +574,10 @@ func (c *LeveledCompactor) write(dest string, meta *BlockMeta, blocks ...BlockRe
 		return errors.Wrap(err, "rename block dir")
 	}
 
+	elapsed := time.Since(start)
+	level.Info(c.logger).Log("msg", "writing blocks to disk finished")
```
bwplotka: Let's put it them into single line. Easier to read and relate. 
```suggestion
	level.Debug(c.logger).Log("msg", "writing blocks to disk finished", "elapsed", time.Since(start))
```

compact.go
bwplotka: Same here:
```suggestion
	level.Debug(c.logger).Log("msg", "writing blocks to disk", "blocks", len(blocks))
```

krasi-georgiev:
@bwplotka 	I think you missed my comment. We already have this log. The only thing that is missing is the duration.

bwplotka:
Yea, my bad - I focused on code only, not on discussion here ): 

Fully agree with @krasi-georgiev then - but @radeklesniewski make sure to not duplicate log lines just for duration as in the proposed code.

radeklesniewski:
@bwplotka @krasi-georgiev  thanks for the feedback. Can we agree on adding duration to below line? https://github.com/prometheus/tsdb/blob/cc277e398b479bdb601e17e439bad11901824aa3/compact.go#L443 
If yes, I will update the PR :)

krasi-georgiev:
yes I already suggested this in my comment:
https://github.com/prometheus/tsdb/pull/505#issuecomment-456111128

compact.go
```diff
@@ -440,7 +442,7 @@ func (c *LeveledCompactor) Write(dest string, b BlockReader, mint, maxt int64, p
 		return ulid.ULID{}, nil
 	}
 
-	level.Info(c.logger).Log("msg", "write block", "mint", meta.MinTime, "maxt", meta.MaxTime, "ulid", meta.ULID)
+	level.Info(c.logger).Log("msg", "write block", "mint", meta.MinTime, "maxt", meta.MaxTime, "ulid", meta.ULID, "duration", time.Since(start))
```
codesome: Nit. To keep the line width sane, can you break each log entity into a single line like this?
https://github.com/prometheus/tsdb/blob/dac2b97dfdf354a874f186f365fa75cd647e755e/compact.go#L391-L399

krasi-georgiev: +1 - like in my example in the comment https://github.com/prometheus/tsdb/pull/505#issuecomment-456111128

radeklesniewski: Thx, formatting updated

PR #508, opened 2019-01-28
rename chunk reader vars to make it easier to follow.

krasi-georgiev:

### DIFF
--- a/chunks/chunks.go
+++ b/chunks/chunks.go
@@ -355,28 +355,28 @@ func (s *Reader) Size() int64 {
 // Chunk returns a chunk from a given reference.
 func (s *Reader) Chunk(ref uint64) (chunkenc.Chunk, error) {
 	var (
-		segmentSeq    = int(ref >> 32)
-		segmentOffset = int((ref << 32) >> 32)
+		sgmSeq    = int(ref >> 32)
+		sgmOffset = int((ref << 32) >> 32)
 	)
-	if segmentSeq >= len(s.bs) {
-		return nil, errors.Errorf("reference sequence %d out of range", segmentSeq)
+	if sgmSeq >= len(s.bs) {
+		return nil, errors.Errorf("reference sequence %d out of range", sgmSeq)
 	}
-	chunksBytes := s.bs[segmentSeq]
+	chkS := s.bs[sgmSeq]
 
-	if segmentOffset >= chunksBytes.Len() {
-		return nil, errors.Errorf("offset %d beyond data size %d", segmentOffset, chunksBytes.Len())
+	if sgmOffset >= chkS.Len() {
+		return nil, errors.Errorf("offset %d beyond data size %d", sgmOffset, chkS.Len())
 	}
 	// With the minimum chunk length this should never cause us reading
 	// over the end of the slice.
-	chunkBytes := chunksBytes.Range(segmentOffset, segmentOffset+binary.MaxVarintLen32)
+	chk := chkS.Range(sgmOffset, sgmOffset+binary.MaxVarintLen32)
 
-	chunkLength, byteCount := binary.Uvarint(chunkBytes)
-	if byteCount <= 0 {
-		return nil, errors.Errorf("reading chunk length failed with %d", byteCount)
+	chkLen, n := binary.Uvarint(chk)
+	if n <= 0 {
+		return nil, errors.Errorf("reading chunk length failed with %d", n)
 	}
-	chunkBytes = chunksBytes.Range(segmentOffset+byteCount, segmentOffset+byteCount+1+int(chunkLength))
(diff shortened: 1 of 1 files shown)


naivewong:
@krasi-georgiev Thanks. The changes are clear!

krasi-georgiev:
> Personally I find the former more readable and well understandable.
short var names is not difficult to read, but without looking at the chunk writer I wouldn't know what r,b,seq,off represent. Maybe the names can be improved, but I still think it should be more descriptive than just single letters.

>  Also not sure what do you mean by segment?
The reader reads a range a chunk files(segments) and segmentSeq is the reference in which segment is this chunk located.

krasi-georgiev:
>The segment word was never used before in context of TSDB, so I feel like it does more confusion then help.

Segments is also used when describing the chunk format in: https://github.com/prometheus/tsdb/blob/master/docs/format/chunks.md

> So I think instead of changing on function with something that some users would like, some not, can we open bigger issue for wider discussion to agree on style for those small, quite critical functions where you can infer context a lot? I think it makes more sense as consistency is addtional factor to those changes.

I don't see this as whether it looks good or not, but rather how quickly can you understand what is happening inside the func. Don't mind more input of course, but ideally from people that would be involved in the long term tsdb maintenance.

bwplotka:
Cool, so this is my input as person that plan to be involved in the long term tsdb maintenance as we ruse this code in production (: 

Anyway, I guess not a blocker, if the majority will decide that this is fine, but I think this particular change might be not necessary. Comment that this related to `segments` would just fine. 

Thanks @krasi-georgiev for referencing `segments` mentions, I was wrong, even though it's not very often used

chunks/chunks.go
```diff
@@ -352,30 +352,31 @@ func (s *Reader) Size() int64 {
...
 
-	if off >= b.Len() {
-		return nil, errors.Errorf("offset %d beyond data size %d", off, b.Len())
+	if segmentOffset >= chunksBytes.Len() {
+		return nil, errors.Errorf("offset %d beyond data size %d", segmentOffset, chunksBytes.Len())
 	}
 	// With the minimum chunk length this should never cause us reading
 	// over the end of the slice.
-	r := b.Range(off, off+binary.MaxVarintLen32)
+	chunkBytes := chunksBytes.Range(segmentOffset, segmentOffset+binary.MaxVarintLen32)
```
codesome: For a moment I was confused about this, different between `chunkBytes` and `chunksBytes` is very subtle. Can one of them be changed?

krasi-georgiev:
@codesome , @bwplotka have another look

beorn7:
I think https://github.com/golang/go/wiki/CodeReviewComments#variable-names is beautifully worded. In no way it implies that _all_ variables have to be one to three letters long. It's more about method receivers and variables that gain an “obvious” meaning from the way they are initialized. I find non-descriptive member variables in structs quite problematic. In this example, there is `b.bs`, which, if it were more descriptive, would have given the `b` in line 363 an obvious meaning, so that it might even have kept its one letter name. `int((ref << 32) >> 32)` is quite obscure. If it were wrapped into a function with a meaningful name, the target variable `off` could again keep its short name because the function call would make its meaning obvious, and `off` is only used in a narrow scope.

Just my 2¢. The tsdb code is often quite dense to read because of exaggerating (or misunderstanding) the Go-idiomatic preference for short variable name.

bwplotka:
Full agree with @beorn7 except one thing:

> int((ref << 32) >> 32) is quite obscure

It is unreadable, but it has nothing to do with variable name (:

krasi-georgiev:
Thanks all for the feedback. In any way this is a step in the right direction so merging as is. Any more improvements are welcome in a separate PR.
Thanks again!

beorn7:
_Disclaimer: This is just meant as some playful and perhaps a bit enlightening musings, not as nitpicking or complaining about this code in particular._

>> int((ref << 32) >> 32) is quite obscure
>
> It is unreadable, but it has nothing to do with variable name (: 

What I meant that this expression is probably one where you would be tempted to add a comment about what it is doing and why. Assigning it to a descriptive variable name might already give a hint.

The original `off = int((ref << 32) >> 32)` was already giving the hint that we are extracting an offset from `ref`. `o = int((ref << 32) >> 32)` would be clearly more obscure. `sgmOffset = int((ref << 32) >> 32)` is better because we get the hint that it is about a segment. Personally, I would say, now that we have left the territory of handy short names anyway, we could as well write `segmentOffset = int((ref << 32) >> 32)`. The latter is IMHO clearly better than `off = int((ref << 32) >> 32) // Extract segment offset.`

The other approach is to have a helper function `func segmentOffset(ref uint64) int { return int((ref << 32) >> 32)}`, which would then allow a short variable again: `off = segmentOffset(ref)`.

PR #233, opened 2017-12-26
index: reduce empty postings trees

fabxc:
If a query has multiple matchers that are intersected, we can potentially reduce the postings tree to the `EmptyPostings()` beforehand. This way we don't have to hit or preload any of the underlying postings lists at all.

This is mostly relevant for `Intersect()` but I added it for the other ones too for consistency.

PR #509, opened 2019-01-28
index: Fast path for EmptyPostings cases in Merge, Intersect and Without

bwplotka:
Updates old https://github.com/prometheus/tsdb/pull/233 onto fresh TSDB code.

closes https://github.com/prometheus/tsdb/pull/233

### DIFF
--- a/index/postings.go
+++ b/index/postings.go
@@ -92,7 +92,7 @@ func (p *MemPostings) Get(name, value string) Postings {
 	if lp == nil {
 		return EmptyPostings()
 	}
-	return newListPostings(lp)
+	return newListPostings(lp...)
 }
 
 // All returns a postings list over all documents ever added.
@@ -202,7 +202,7 @@ func (p *MemPostings) Iter(f func(labels.Label, Postings) error) error {
 
 	for n, e := range p.m {
 		for v, p := range e {
-			if err := f(labels.Label{Name: n, Value: v}, newListPostings(p)); err != nil {
+			if err := f(labels.Label{Name: n, Value: v}, newListPostings(p...)); err != nil {
 				return err
 			}
 		}
@@ -283,6 +283,8 @@ func (e errPostings) Err() error       { return e.err }
 var emptyPostings = errPostings{}
 
 // EmptyPostings returns a postings list that's always empty.
+// NOTE: Returning EmptyPostings sentinel when index.Postings struct has no postings is recommended.
+// It triggers optimized flow in other functions like Intersect, Without etc.
 func EmptyPostings() Postings {
 	return emptyPostings
 }
@@ -296,13 +298,20 @@ func ErrPostings(err error) Postings {
 // input postings.
 func Intersect(its ...Postings) Postings {
 	if len(its) == 0 {
-		return emptyPostings
+		return EmptyPostings()
 	}
 	if len(its) == 1 {
 		return its[0]
 	}
+
 	l := len(its) / 2
-	return newIntersectPostings(Intersect(its[:l]...), Intersect(its[l:]...))
+	a := Intersect(its[:l]...)
+	b := Intersect(its[l:]...)
+
+	if a == EmptyPostings() || b == EmptyPostings() {
+		return EmptyPostings()
+	}
+	return newIntersectPostings(a, b)
 }
 
 type intersectPostings struct {
@@ -366,7 +375,12 @@ func Merge(its ...Postings) Postings {
 	if len(its) == 1 {
 		return its[0]
 	}
-	return newMergedPostings(its)
+
+	p, ok := newMergedPostings(its)
+	if !ok {
+		return EmptyPostings()
+	}
+	return p
 }
 
 type postingsHeap []Postings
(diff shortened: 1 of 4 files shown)


bwplotka:
ping @gouthamve

krasi-georgiev:
@bwplotka 	I did a PR against your branch with few adjustments.
https://github.com/bwplotka/tsdb/pull/2

krasi-georgiev:
LGTM, Thanks!

krasi-georgiev:
@codesome 	want to have a quick look before I merge it?

krasi-georgiev:
ping @bwplotka

bwplotka:
Updated @krasi-georgiev and @codesome 

Thanks for ping & reviews. Ping me directly on IRC or slack if you want my review/PR update I am flooded with GH notifications ):

I changed a lot now, because of conflicts - I rewrote test cases to make more sense, let me know if that is OK with you guys.

index/postings.go
```diff
@@ -395,18 +409,24 @@ type mergedPostings struct {
...
 			if it.Err() != nil {
-				return &mergedPostings{err: it.Err()}
+				return &mergedPostings{err: it.Err()}, true
 			}
 		}
 	}
-	return &mergedPostings{h: ph}
+
+	if len(ph) == 0 {
+		return nil, false
```
krasi-georgiev: shouldn't this be before anything else?

index/postings_test.go
```diff
@@ -61,39 +61,125 @@ func TestMemPostings_ensureOrder(t *testing.T) {
...
+				toPostings(),
+			},
+			res: toPostings(),
 		},
 	}
 
 	for _, c := range cases {
-		a := newListPostings(c.a)
-		b := newListPostings(c.b)
+		t.Run("", func(t *testing.T) {
```
krasi-georgiev: maybe add some indication to know when a test fails which case failed.
```
map[string]struct {
.
.

for t, c := range cases {
		t.Run(t, func(t *testing.T) {
```

index/postings_test.go
```diff
@@ -61,39 +61,125 @@ func TestMemPostings_ensureOrder(t *testing.T) {
...
+			i := Intersect(c.in...)
+
+			if c.res == EmptyPostings() {
+				testutil.Equals(t, EmptyPostings(), i)
+				return
+			}
+
+			if i == EmptyPostings() {
+				t.Fatal("intersect unexpected result: EmptyPostings sentinel")
+			}
```
krasi-georgiev: why do we need?
```
if c.res == EmptyPostings() {
	testutil.Equals(t, EmptyPostings(), i)
	return
}

if i == EmptyPostings() {
	t.Fatal("intersect unexpected result: EmptyPostings sentinel")
}
```

the test passes even without these extra checks.

index/postings_test.go
```diff
@@ -541,3 +698,81 @@ func TestIntersectWithMerge(t *testing.T) {
...
+			base: toPostings(1, 2, 3),
+			drop: toPostings(3, 4, 5),
+
+			res: toPostings(1, 2),
+		},
+	}
+
+	for _, c := range cases {
+		t.Run("", func(t *testing.T) {
+			if c.res == nil {
```
krasi-georgiev: don't think we need this. A panic is ok in a test in my opinion

index/postings.go
bwplotka: hm? We know this only after `it.Next` bit.

index/postings_test.go
bwplotka: not needed. Passing `""` cause go test to put a number, so exactly what you propose.

index/postings_test.go
bwplotka: Exactly to expect `EmptyPostings` explicitly!

index/postings_test.go
bwplotka: If you think so, can remove

index/postings_test.go
krasi-georgiev: yes that is right but when you have many cases it is kind of difficult to count which one failed exactly.

index/postings.go
krasi-georgiev: yep , I missed that bit.

index/postings_test.go
```diff
@@ -197,35 +283,106 @@ func TestMultiMerge(t *testing.T) {
...
+				EmptyPostings(),
+				EmptyPostings(),
+				EmptyPostings(),
+				EmptyPostings(),
+			},
+			res: EmptyPostings(),
+		},
+		{
+			in: []Postings{
+				toPostings(1, 2, 3, 4, 5),
```
krasi-georgiev: what is blocking to use `newListPostings` everywhere?

index/postings_test.go
krasi-georgiev: yep, now I see it.

index/postings_test.go
bwplotka: boilerplate

index/postings.go
```diff
@@ -382,13 +388,21 @@ func Merge(its ...Postings) Postings {
...
+
 	sort.Slice(pl, func(i, j int) bool { return pl[i] < pl[j] })
 	return newListPostings(pl)
 }
 
 // Without returns a new postings list that contains all elements from the full list that
-// are not in the drop list
+// are not in the drop list.
 func Without(full, drop Postings) Postings {
+	if full == EmptyPostings() {
```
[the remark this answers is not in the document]
krasi-georgiev: nope, drop are the ones to be removed

index/postings_test.go
bwplotka: Disagree, let's not obfuscate the tests more -> everyone can count

krasi-georgiev: that is why I added the title in:
https://github.com/prometheus/tsdb/pull/559/files

bwplotka: Yea, I agree there is a value for this, but I am afraid I won't have time to do this, this week ): 

I would suggesting merging this PR with just indexing (numbers), then we can work on making it more verbose later on. What do you think?

index/postings_test.go
krasi-georgiev: how about if you change 
`newListPostings([]uint64) -> newListPostings(list ...uint64)`
than you can call it as `newListPostings(1, 2, 3, 4, 5)`
just tried it and it removes that boilerplate without the need for an extra func.


not a blocker though can leave it as if you prefer.

index/postings_test.go
krasi-georgiev: sure I don't mind can always add it later if it gets annoying to count :smile:

index/postings_test.go
bwplotka: It will increase this PR size, but did it, I like this idea :+1:

krasi-georgiev:
Thanks Bartek!

bwplotka:
Thanks Krasi (:

PR #512, opened 2019-01-29
Delete compact block on reload error

krasi-georgiev:
To avoid a dead loop a failed reload immediately after a compaction should delete the
resulting block to avoid creating blocks with the same time range.

fixes https://github.com/prometheus/prometheus/issues/5135


brian-brazil:
Wouldn't this leave us with a race if the Prometheus happened to die before the deletion happened?

codesome:
Continuing what @brian-brazil said, how about having a boolean field in the meta (`loaded`?) - `false` would mean "it was compacted and not loaded even once".

Would this make it simple to handle the race?

brian-brazil:
That'd mean that we'd be changing blocks after they're written, which I don't think we should start doing.

krasi-georgiev:
I don't see a race here.
Prometheus will die and will load the block as normal at the next start up.

krasi-georgiev:
most changes that look unrelated are for the windows tests to pass.

codesome:
> I don't see a race here.

On a second look now, feels the same.

krasi-georgiev:
@brian-brazil Do you see a race somewhere else? 
re my comment at  https://github.com/prometheus/tsdb/pull/512#issuecomment-458878260

compact_test.go
```diff
@@ -741,3 +745,78 @@ func TestDisableAutoCompactions(t *testing.T) {
 	}
 	testutil.Assert(t, len(db.Blocks()) > 0, "No block was persisted after the set timeout.")
 }
+
+// TestDeleteCompactionBlockAfterFailedReload ensures that a failed reload imidiately after a compaction
```
codesome:
```suggestion
// TestDeleteCompactionBlockAfterFailedReload ensures that a failed reload immediately after a compaction
```

db_test.go
```diff
@@ -429,11 +448,11 @@ func TestDB_Snapshot(t *testing.T) {
 	// reopen DB from snapshot
 	db, err = Open(snap, nil, nil, nil)
 	testutil.Ok(t, err)
-	defer db.Close()
+	defer func() { testutil.Ok(t, db.Close()) }()
 
 	querier, err := db.Querier(mint, mint+1000)
 	testutil.Ok(t, err)
-	defer querier.Close()
+	defer func() { querier.Close() }()
```
codesome: Did you mean to have `testutil.Ok(t, querier.Close())`?

communityexplorer:
how can I know which file should I delete?

codesome:
@communityexplorer Can you give us some context?

compact_test.go
```diff
@@ -741,3 +745,78 @@ func TestDisableAutoCompactions(t *testing.T) {
...
+			testutil.Equals(t, expBlocks, len(actBlocks)-1) // -1 to exclude the corrupted block.
+			testutil.Ok(t, os.RemoveAll(lastBlockIndex))    // Corrupt the block by removing the index file.
+
+			testutil.Equals(t, 0.0, prom_testutil.ToFloat64(db.metrics.reloadsFailed), "initial 'failed db reload' count metrics mismatch")
+			testutil.Equals(t, 0.0, prom_testutil.ToFloat64(db.compactor.(*LeveledCompactor).metrics.ran), "initial `compactions` count metric mismatch")
+
+			// Do the compaction and check the metrics.
+			// Compaction should succeed, but the reload should fail and
+			// the new block created from the compaction should be deleted.
+			db.EnableCompactions()
```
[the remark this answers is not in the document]
krasi-georgiev: thanks 
updated!

compact_test.go
```diff
@@ -741,3 +745,78 @@ func TestDisableAutoCompactions(t *testing.T) {
...
+			})
+			defer func() {
+				testutil.Ok(t, db.Close())
+				delete()
+			}()
+			db.DisableCompactions()
+
+			expBlocks := bootStrap(db)
+
+			// Create a block that will trigger the reloard to fail.
```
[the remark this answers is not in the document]
krasi-georgiev: thanks 
updated!

compact_test.go
krasi-georgiev: thanks 
updated!

db_test.go
krasi-georgiev: thanks 
updated!

krasi-georgiev:
> We need to do it in `db.CleanTombstones()` too. It creates new blocks and calls `db.reload()`.

good thought. I just double checked and this is already handled.

https://github.com/prometheus/tsdb/blob/0b72f9af4c7a7953f89c993e69bbf81b12af49c3/db.go#L964-L971

block.go
```diff
@@ -269,10 +270,18 @@ type Block struct {
...
 // to instantiate chunk structs.
-func OpenBlock(logger log.Logger, dir string, pool chunkenc.Pool) (*Block, error) {
+func OpenBlock(logger log.Logger, dir string, pool chunkenc.Pool) (pb *Block, err error) {
 	if logger == nil {
 		logger = log.NewNopLogger()
 	}
+	var closers []io.Closer
+	defer func() {
+		if err != nil {
+			for _, c := range closers {
```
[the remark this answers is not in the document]
krasi-georgiev: thanks. Updated!

codesome:
Apparently, it was my PR https://github.com/prometheus/tsdb/pull/341 :)

communityexplorer:
> @communityexplorer Can you give us some context?
as we know, it recreated many repeated blocks, how can we know  which blocks should we delete, is it necessary to restart prometheus after that, how can we avoid this promblem?

codesome:
@communityexplorer In this PR the blocks to be deleted are known (the one being created in compaction), and it is done automatically.

communityexplorer:
@codesome @krasi-georgiev 
when will this solved version release?  my PR is #5120, this problem has been frequent recently in our production env, can you give release as soon as possible?

communityexplorer:
another question:
(memory size) = vm.max_map_count * (memory chunk)
what is the memory chunk size for prometheus?
thanks so much

simonpasquier:
@communityexplorer this fix should be available in Prometheus v2.8.

PR #513, opened 2019-01-31
optimize createBlock - generate the labels map only once.

krasi-georgiev:
This reduces the CPU time in half.

cc @codesome 

### DIFF
--- a/block_test.go
+++ b/block_test.go
@@ -102,11 +102,12 @@ func genSeries(totalSeries, labelCount int, mint, maxt int64) []Series {
 	}
 	series := make([]Series, totalSeries)
 
+	lbls := make(map[string]string, labelCount)
+	for len(lbls) < labelCount {
+		lbls[randString()] = randString()
+	}
 	for i := 0; i < totalSeries; i++ {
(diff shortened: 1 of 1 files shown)


brian-brazil:
This code isn't doing the same thing, each created series would now have the same labels.

codesome:
Agree with @brian-brazil

krasi-georgiev:
yes that is right but with maps the order is randomised so it is essentially different series.

Anyway I added some more code to add one unique label so now it is clear that it creates unique series.

block_test.go
```diff
@@ -102,15 +102,19 @@ func genSeries(totalSeries, labelCount int, mint, maxt int64) []Series {
 	}
 	series := make([]Series, totalSeries)
 
+	lbls := make(map[string]string, labelCount)
+	var lastKey string
+	for len(lbls) < labelCount {
+		lastKey := randString()
```
codesome:
```suggestion
		lastKey = randString()
```

krasi-georgiev:
it is not in an immediate  needed so will close and revisit if needed.

PR #514, opened 2019-02-04
compact: Fixed populateBlock to adjust out ouf boundary chunks automatically.

bwplotka:
Fixes https://github.com/prometheus/prometheus/issues/5105

This is seems to be in the case when you do snapshot and it will invoke `write.Compact` for head
block with certain MaxTime being time.Now in the moment of invoke. During compaction
head block's chunks can be appended anytime resulting in chunks being out max Time.

Changes:
* For block write: 
    * If chunk is partially outside max time, remove leftovers
    * If chunk is completely outside skip it.
    * If chunk is before min time - error out, we don't have use case for this.
* Improvement comments for compactor write API. We aim to guarantee block chunks being
within requested min and max time.
* Added test cases
* Avoided returning via argument if not needed (meta stats)
* Handled error from reader closers in populate Block. For now they are logged as error level.
(We should not block write if we cannot close reader I guess, but we have to be aware at least)

Just to double check: Blocks has (minTime, maxTime> chunks right? Where was documentation for this?
I remember we wanted to note down this somewhere (:

### DIFF
--- a/compact.go
+++ b/compact.go
@@ -54,7 +54,7 @@ type Compactor interface {
 	// Results returned when compactions are in progress are undefined.
 	Plan(dir string) ([]string, error)
 
-	// Write persists a Block into a directory.
+	// Write persists a Block into a directory by including data within the requested mint:maxt range.
 	// No Block is written when resulting Block has 0 samples, and returns empty ulid.ULID{}.
 	Write(dest string, b BlockReader, mint, maxt int64, parent *BlockMeta) (ulid.ULID, error)
 
@@ -471,8 +471,6 @@ func (w *instrumentedChunkWriter) WriteChunks(chunks ...chunks.Meta) error {
 	return w.ChunkWriter.WriteChunks(chunks...)
 }
 
-// write creates a new block that is the union of the provided blocks into dir.
-// It cleans up all files of the old blocks after completing successfully.
 func (c *LeveledCompactor) write(dest string, meta *BlockMeta, blocks ...BlockReader) (err error) {
 	dir := filepath.Join(dest, meta.ULID.String())
 	tmp := dir + ".tmp"
@@ -522,9 +520,12 @@ func (c *LeveledCompactor) write(dest string, meta *BlockMeta, blocks ...BlockRe
 	}
 	defer indexw.Close()
 
-	if err := c.populateBlock(blocks, meta, indexw, chunkw); err != nil {
+	stats, err := c.populateBlock(blocks, meta.MinTime, meta.MaxTime, indexw, chunkw)
+	if err != nil {
 		return errors.Wrap(err, "write compaction")
 	}
+	meta.Stats = stats
+
 	// We are explicitly closing them here to check for error even
 	// though these are covered under defer. This is because in Windows,
 	// you cannot delete these unless they are closed and the defer is to
@@ -582,41 +583,46 @@ func (c *LeveledCompactor) write(dest string, meta *BlockMeta, blocks ...BlockRe
 }
 
 // populateBlock fills the index and chunk writers with new data gathered as the union
-// of the provided blocks. It returns meta information for the new block.
-func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta, indexw IndexWriter, chunkw ChunkWriter) error {
+// of the provided blocks that overlaps with requested min and max time.
+// It returns new block's statistics.
+func (c *LeveledCompactor) populateBlock(blocks []BlockReader, mint int64, maxt int64, indexw IndexWriter, chunkw ChunkWriter) (stats BlockStats, err error) {
 	if len(blocks) == 0 {
-		return errors.New("cannot populate block from no readers")
+		return BlockStats{}, errors.New("cannot populate block from no readers")
 	}
 
 	var (
 		set        ChunkSeriesSet
 		allSymbols = make(map[string]struct{}, 1<<16)
-		closers    = []io.Closer{}
+		closers    []io.Closer
 	)
-	defer func() { closeAll(closers...) }()
+	defer func() {
+		if err := closeAll(closers...); err != nil {
+			level.Error(c.logger).Log("msg", "one or more readers failed to close", "err", err.Error())
+		}
+	}()
 
 	for i, b := range blocks {
 		indexr, err := b.Index()
 		if err != nil {
-			return errors.Wrapf(err, "open index reader for block %s", b)
+			return BlockStats{}, errors.Wrapf(err, "open index reader for block %s", b)
 		}
 		closers = append(closers, indexr)
 
 		chunkr, err := b.Chunks()
 		if err != nil {
-			return errors.Wrapf(err, "open chunk reader for block %s", b)
+			return BlockStats{}, errors.Wrapf(err, "open chunk reader for block %s", b)
 		}
 		closers = append(closers, chunkr)
 
 		tombsr, err := b.Tombstones()
 		if err != nil {
-			return errors.Wrapf(err, "open tombstone reader for block %s", b)
+			return BlockStats{}, errors.Wrapf(err, "open tombstone reader for block %s", b)
 		}
 		closers = append(closers, tombsr)
 
 		symbols, err := indexr.Symbols()
 		if err != nil {
-			return errors.Wrap(err, "read symbols")
+			return BlockStats{}, errors.Wrap(err, "read symbols")
 		}
 		for s := range symbols {
 			allSymbols[s] = struct{}{}
@@ -624,7 +630,7 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta, 
 
 		all, err := indexr.Postings(index.AllPostingsKey())
 		if err != nil {
-			return err
+			return BlockStats{}, err
 		}
 		all = indexr.SortedPostings(all)
 
@@ -636,7 +642,7 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta, 
 		}
 		set, err = newCompactionMerger(set, s)
 		if err != nil {
-			return err
+			return BlockStats{}, err
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


krasi-georgiev:
my review is in https://github.com/bwplotka/tsdb/pull/3

compact.go
```diff
@@ -54,7 +54,7 @@ type Compactor interface {
 	// Results returned when compactions are in progress are undefined.
 	Plan(dir string) ([]string, error)
 
-	// Write persists a Block into a directory.
+	// Write persists index and chunks from a Block for samples overlaps with mint:maxt range into a directory.
```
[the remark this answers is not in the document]
bwplotka: Because saying "new block is the union of given blocks" is not true. It is the union of new blocks that intersect with request min and max time, right? And that's why I changed this comment and others for `populate` as well.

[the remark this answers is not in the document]
krasi-georgiev: yep that is right. 
How about:
`// Write persists a Block into a directory by including data within the requested mint:maxt range.`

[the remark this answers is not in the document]
krasi-georgiev: ok , thanks
Done.

bwplotka:
Addressed comments @krasi-georgiev  let me know if I missed anything from your PR to my PR ;p

compact.go
```diff
@@ -471,7 +471,7 @@ func (w *instrumentedChunkWriter) WriteChunks(chunks ...chunks.Meta) error {
 	return w.ChunkWriter.WriteChunks(chunks...)
 }
 
-// write creates a new block that is the union of the provided blocks into dir.
+// write creates a new block into a directory by including data within the requested mint:maxt range.
```
krasi-georgiev: not even sure why we need this comment it is a complete copy of the comment for `Write(...`

compact.go
```diff
@@ -660,49 +668,66 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
...
+				sort.Slice(dranges, func(i, j int) bool {
+					return dranges[i].Mint < dranges[j].Maxt
+				})
+			}
+
+			if len(dranges) == 0 || !chk.OverlapsClosedInterval(dranges[0].Mint, dranges[len(dranges)-1].Maxt) {
+				continue
+			}
+
+			// Re-encode the chunk to not have deleted values or to match requested max time.
```
krasi-georgiev:
```suggestion
			// Rewrite chunk to remove deleted values.
```

compact.go
```diff
@@ -660,49 +668,66 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
...
+				continue
+			}
 
-				it := &deletedIterator{it: chk.Chunk.Iterator(), intervals: dranges}
-				for it.Next() {
-					ts, v := it.At()
-					app.Append(ts, v)
-				}
+			if chk.MaxTime > maxt {
+				// Rewrite chunk to remove samples that are outside requested time range.
```
krasi-georgiev:
```suggestion
				// Remove chunk values outside the requested time range.
```

compact.go
```diff
@@ -660,49 +668,66 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
...
+
+			// Re-encode the chunk to not have deleted values or to match requested max time.
+			newChunk := chunkenc.NewXORChunk()
+			app, err := newChunk.Appender()
+			if err != nil {
+				return BlockStats{}, err
+			}
 
-				chks[i].Chunk = newChunk
+			fmt.Println(dranges)
```
krasi-georgiev:
```suggestion
```

compact_test.go
```diff
@@ -650,16 +723,8 @@ func TestCompaction_populateBlock(t *testing.T) {
 			c, err := NewLeveledCompactor(nil, nil, []int64{0}, nil)
 			testutil.Ok(t, err)
 
-			meta := &BlockMeta{
-				MinTime: tc.compactMinTime,
-				MaxTime: tc.compactMaxTime,
-			}
-			if meta.MaxTime == 0 {
-				meta.MaxTime = math.MaxInt64
-			}
```
krasi-georgiev: I think this will fail without 
```
if tc.compactMaxTime == 0 {
				tc.compactMaxTime = math.MaxInt64
			}
```

compact.go
bwplotka: agree

compact.go
bwplotka: HM.. not only deleted values. Also those which are out of max time right?

compact_test.go
bwplotka: I added explicit ranged to all use cases

compact.go
krasi-georgiev: that is right, but in regards to this block it removes samples added to the `dranges` so semantically it removes data from the dranges or in other words remove deletes values(dranges) from the chunk

krasi-georgiev: maybe `// Rewrite chunk to remove drange values.`

bwplotka: :+1:

compact.go
```diff
@@ -659,50 +665,68 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
 			continue
 		}
 
-		for i, chk := range chks {
-			if chk.MinTime < meta.MinTime || chk.MaxTime > meta.MaxTime {
-				return errors.Errorf("found chunk with minTime: %d maxTime: %d outside of compacted minTime: %d maxTime: %d",
-					chk.MinTime, chk.MaxTime, meta.MinTime, meta.MaxTime)
+		for i := 0; i < len(chks); i++ {
```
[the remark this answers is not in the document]
bwplotka: Yup, so as you can see we remove elements from the `chks` while iterating.

It is safe to do, however `i` (current iterator) needs to be decremented (as the current element was removed). This is not possible in `for i := range array` since `i` is always incremented. Move to this allows for flexibility.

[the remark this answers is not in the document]
krasi-georgiev: yep, noted. I remember hitting this scenario few times.

beorn7:
I guess once this is in, we update dependencies in the release-2.7 branch of prometheus/prometheus and release 2.7.2? (I'm keen on rolling out 2.7 at SC...)

bwplotka:
As @gouthamve said, we have still race with using head chunks.Bytes() and head appending samples to open chunk.

I think being explicit about open chunks (by specifying maxTime to `math.MaxInt64`) would fix this as that would trigger `Iterating` over it. Let me check if that is doable in this PR.

bwplotka:
Wonder if this will help: https://github.com/bwplotka/tsdb/pull/4

brian-brazil:
What's the status of this?

bwplotka:
Sorry for lag. Last time i spoke offline about this with @gouthamve there are lot's of race conditions to fix and we did not have exact decision to follow on. Let's use some time next week to get back to this. It needs some work from our side.

brian-brazil:
How is this going?

bwplotka:
Syncing with @gouthamve on this. AFAIK as @gouthamve said this is not fixing all cases, and we are both busy on something else. 

I am closing this for now, but we should prioritze this if the issue will become more problematic. Should be gather some feedback on issue maybe?

brian-brazil:
I've reports of this causing sporadic backup failures from another user.

brian-brazil:
> I think being explicit about open chunks (by specifying maxTime to math.MaxInt64) would fix this as that would trigger Iterating over it.

Where do you think it'd trigger iterating? I don't see that code anywhere, but adding it in populateBlock by the deletion logic with https://github.com/bwplotka/tsdb/pull/4/files should do the trick.

It also needs isolation to be correct, but that's another story.

bwplotka:
Ok got back into this @brian-brazil Looked through code and has some questions:

* Why we have this check even: https://github.com/prometheus/tsdb/blob/8d86e921cf00c882f36577ddfc452911270fed32/compact.go#L761
It's clear that in case of snapshot on head that has open chunks it can grow out of specified maxTime. IMO the easiest thing would to in case of chunk.MaxTime > meta.MaxTime either ignore this chunk (quite silly as we might lose a lot of samples e.g for 15s interval up to 30 minutes) or rather rewrite it to `[chunk.minTime, meta.MaxTime]` and put into block. Does it make sense? We don't need special MathMax markers for maxTime to have this like in https://github.com/bwplotka/tsdb/pull/4/files, we could just check if they are out of meta.MaxTime. Also the MathMax markers would be not enough as there can be race between Snapshot and chunk `cut` so we can have closed chunk but outside of desired meta.MaxTime (unless there is some locking there).

* Maybe silly question - but why we don't hit this during cutting the head every 2h? We choose arbitrary MaxTime based on min compaction level so it can happen that the MaxTime is *during* some chunk, right? What am I missing here (:

bwplotka:
@brian-brazil some attempt proposed here: https://github.com/prometheus/tsdb/pull/639

brian-brazil:
> We don't need special MathMax markers for maxTime to have this like in https://github.com/bwplotka/tsdb/pull/4/files, we could just check if they are out of meta.MaxTime

That'd remove the sanity check for non-head chunks though.

> Maybe silly question - but why we don't hit this during cutting the head every 2h?

I'm not sure.

bwplotka:
Let's dive into this more then, maybe @krasi-georgiev or @codesome knows?

bwplotka:
Just FYI: 
> Maybe silly question - but why we don't hit this during cutting the head every 2h?

It's because we try to sync chunks range with block range: https://github.com/prometheus/tsdb/blob/69740485c196961bbc6433271e0f6357ff251b93/head.go#L1600

zhanglijingisme:
Hi @bwplotka @brian-brazil 
I'm using Prometheus v2.9.1 and hit this today, when I'm trying to create snapshot for data (3.8G size).
`{"status":"error","errorType":"internal","error":"create snapshot: snapshot head block: write compaction: found chunk with minTime: 1564020019830 maxTime: 1564023139830 outside of compacted minTime: 1564012800000 maxTime: 1564023138442"}`

Seems this still exists in v2.9.1 (released in Apr.17th) - Is there any fix on latest version 2.11.1 (released in Jul. 10th)?

bwplotka:
@zhanglijingisme Yes. It is part of [0.9.0](https://github.com/prometheus/tsdb/blob/master/CHANGELOG.md#090) which is brough by Prometheus [2.11.0](https://github.com/prometheus/prometheus/blob/master/CHANGELOG.md#2110--2019-07-09) (:

PR #515, opened 2019-02-05
Update index.md(format of series entry)

naivewong:
These several changes refer to #511, [AddSeries(index.go)](https://github.com/prometheus/tsdb/blob/master/index/index.go#L297), and [writeOffsetTable(index.go)](https://github.com/prometheus/tsdb/blob/master/index/index.go#L449).


krasi-georgiev:
looking at the code the file looks outdated.

first it ads the labels set count
`w.buf2.putUvarint(len(lset))`
than the label name/value index from the symbols table
`w.buf2.putUvarint32(index)`
than it adds the chunks count
`w.buf2.putUvarint(len(chunks))`
than it adds all chunks mint,delta, and ref
```
w.buf2.putVarint64(c.MinTime)
w.buf2.putUvarint64(uint64(c.MaxTime - c.MinTime))
w.buf2.putUvarint64(c.Ref)
```
than it adds to another buffer the bytes length so far
`w.buf1.putUvarint(w.buf2.len())`
than it adds the crc
`w.buf2.putHash(w.crc32)`

or in other words the chunks count and lset count is not shown in the file.

also it would look cleaner if it is like:

#labels
<uvarint32>

#chunks
<uvarint64>

and all the other sub-blocks under these don't need to include any types like: ref(l_i.name) ~~<uvarint32>~~



https://github.com/prometheus/tsdb/blob/1bcda9d23f9951f1333cd2a3ba2a97b6b6ec69c4/index/index.go#L320-L364

naivewong:
Sorry, I don't quite understand. 
First, 

> or in other words the chunks count and lset count is not shown in the file.

but I can see the lset count and chunks count already appear in [index.md](https://github.com/prometheus/tsdb/blob/master/docs/format/index.md) as follows, and the order is exactly the same with the codes(index.go).
```
┌─────────────────────────────────────────────────────────────────────────┐
│ len <uvarint>                                                           │
├─────────────────────────────────────────────────────────────────────────┤
│ ┌──────────────────┬──────────────────────────────────────────────────┐ │
│ │                  │ ┌──────────────────────────────────────────┐     │ │
│ │                  │ │ ref(l_i.name) <uvarint>                  │     │ │
│ │     #labels      │ ├──────────────────────────────────────────┤ ... │ │
│ │    <uvarint>     │ │ ref(l_i.value) <uvarint>                 │     │ │
│ │                  │ └──────────────────────────────────────────┘     │ │
│ ├──────────────────┼──────────────────────────────────────────────────┤ │
│ │                  │ ┌──────────────────────────────────────────┐     │ │
│ │                  │ │ c_0.mint <varint>                        │     │ │
│ │                  │ ├──────────────────────────────────────────┤     │ │
│ │                  │ │ c_0.maxt - c_0.mint <uvarint>            │     │ │
│ │                  │ ├──────────────────────────────────────────┤     │ │
│ │                  │ │ ref(c_0.data) <uvarint>                  │     │ │
│ │      #chunks     │ └──────────────────────────────────────────┘     │ │
│ │     <uvarint>    │ ┌──────────────────────────────────────────┐     │ │
│ │                  │ │ c_i.mint - c_i-1.maxt <uvarint>          │     │ │
│ │                  │ ├──────────────────────────────────────────┤     │ │
│ │                  │ │ c_i.maxt - c_i.mint <uvarint>            │     │ │
│ │                  │ ├──────────────────────────────────────────┤ ... │ │
│ │                  │ │ ref(c_i.data) - ref(c_i-1.data) <varint> │     │ │
│ │                  │ └──────────────────────────────────────────┘     │ │
│ └──────────────────┴──────────────────────────────────────────────────┘ │
├─────────────────────────────────────────────────────────────────────────┤
│ CRC32 <4b>                                                              │
└─────────────────────────────────────────────────────────────────────────┘
```
Second, 

> also it would look cleaner if it is like:
>```
>#labels
><uvarint32>
>
>#chunks
><uvarint64>
>``` 

I agree with you, since `w.buf2.putUvarint(len(lset))` and `w.buf2.putUvarint(len(chunks))`, and in `func (e *encbuf) putUvarint(x int)      { e.putUvarint64(uint64(x)) }` in [encoding_helpers.go](https://github.com/prometheus/tsdb/blob/master/index/encoding_helpers.go#L43), putUvarint actually puts a uvarint64.
Then it maybe look like
```
#labels
<uvarint64>

#chunks
<uvarint64>
```

Third, 

> and all the other sub-blocks under these don't need to include any types like: ref(l_i.name)

I am not quite sure what you mean, but in my understanding, the ref() means it is an index(or offset) in symbols table, offset in chunk file, etc. 
Just like
```
// here we have an index for the symbol file if v2, otherwise it's an offset
index, ok := w.symbols[l.Name]
if !ok {
	return errors.Errorf("symbol entry for %q does not exist", l.Name)
}
w.buf2.putUvarint32(index) 
```
you put an a label name's symbols table index, and it's represented as **ref(l_i.name)** in index.md

krasi-georgiev:
aaah now I get it

`#labels <uvarint64>` and `#chunks <uvarint64>` actually represent the count so to be more clear for the next lost souls lookingit should be something like

`#labels count <uvarint64>` and `#chunks count <uvarint64>`

```
┌─────────────────────────────────────────────────────────────────────────┐
│ len <uvarint>                                                           │
├─────────────────────────────────────────────────────────────────────────┤
│ ┌─────────────────────────────────────────────────────────────────────┐ │
| |        #labels count <uvarint64>                                    │ │
| |─────────────────────────────────────────────────────────────────────┘ │
│ │             ┌────────────────────────────────────────────────┐        │
│ │             │ ref(l_i.name) <uvarint>                        │        │
│ │             ├────────────────────────────────────────────────┤ ...    │
│ │             │ ref(l_i.value) <uvarint>                       │        │
│ │             └────────────────────────────────────────────────┘        │

```

krasi-georgiev:
much better, thanks

also I think the dots (...) should be under, not on the side

docs/format/index.md
```diff
@@ -85,34 +85,40 @@ After the labels, the number of indexed chunks is encoded, followed by a sequenc
...
-│ │                  │ └──────────────────────────────────────────┘     │ │
-│ └──────────────────┴──────────────────────────────────────────────────┘ │
-├─────────────────────────────────────────────────────────────────────────┤
-│ CRC32 <4b>                                                              │
-└─────────────────────────────────────────────────────────────────────────┘
+┌──────────────────────────────────────────────────────────────────────────┐
+│ len <uvarint>                                                            │
+├──────────────────────────────────────────────────────────────────────────┤
+│ ┌──────────────────────────────────────────────────────────────────────┐ │
+│ │                     #labels <uvarint64>                              │ │
```
krasi-georgiev: #labels count <uvarint64>

docs/format/index.md
```diff
@@ -85,34 +85,40 @@ After the labels, the number of indexed chunks is encoded, followed by a sequenc
...
+│ ├──────────────────────────────────────────────────────────────────────┤ │
+│ │              ┌────────────────────────────────────────────┐          │ │
+│ │              │ ref(l_i.name) <uvarint32>                  │          │ │
+│ │              ├────────────────────────────────────────────┤          │ │
+│ │              │ ref(l_i.value) <uvarint32>                 │          │ │
+│ │              └────────────────────────────────────────────┘          │ │
+│ │                             ...                                      │ │
+│ ├──────────────────────────────────────────────────────────────────────┤ │
+│ │                     #chunks <uvarint64>                              │ │
```
krasi-georgiev: #chunks count <uvarint64>

bwplotka:
Yea it might confusing @krasi-georgiev 

Essentially this `#` is meant to be "number of" symbol. So `#chunks count` does not make much sense. (: 

I am fine with either:
```
#chunks
```
or 
```
chunks count
```

krasi-georgiev:
I didn't know sorry. Where does this come from?

bwplotka:
Probably this: https://en.wikipedia.org/wiki/Number_sign 

=D

krasi-georgiev:
I still much prefer to have this more explicit as it might confuse other people as well so maybe just `chunks count` , `labels count`. Without the hash

krasi-georgiev:
Thanks!

PR #517, opened 2019-02-06
Nolock Option for TSDB CLI Tool

glightfoot:
Adds an option to not lock the TSDB when opening. Might need some additional warnings, or maybe just made default, but this seems to work fine on my internal testing against a heavily used 2.7.1 prometheus server.

fixes https://github.com/prometheus/tsdb/pull/517


cmd/tsdb/main.go
```diff
@@ -55,6 +57,13 @@ func main() {
 		analyzeLimit         = analyzeCmd.Flag("limit", "how many items to show in each list").Default("20").Int()
 	)
 
+	options := &tsdb.Options{
+		WALSegmentSize:    wal.DefaultSegmentSize,
+		RetentionDuration: 15 * 24 * 60 * 60 * 1000, // 15 days in milliseconds
+		BlockRanges:       tsdb.ExponentialBlockRanges(int64(2*time.Hour)/1e6, 3, 5),
+		NoLockfile:        false,
```
krasi-georgiev: since this is set as a global option I expect this to be set here
```suggestion
		NoLockfile:        *noLock,
```

brian-brazil:
Rather than the user having to know of this magic flag, could we verify where it's safe and do it automatically?

glightfoot:
@brian-brazil I like that more, but don't know enough about the workings to know if it's safe 100% of the time or not. From my browsing, it looks like it's only needed for `ls` and `analyze` - if those are guaranteed to be safe running against a running prometheus, I can just hard code it and remove the flag

brian-brazil:
The question is if there's any potentially mutating actions, such as migrating the WAL or compactions.

krasi-georgiev:
The only problem I see when running on a live DB could be incorrect results if some compaction happens during the list/analyse command or under windows the compaction might fail to delete the block when it is open for reading.

Either way I think the suggestion to make it a default and non configurable is fine and can a short warning like:
`listCmd              = cli.Command("ls", "list db blocks (running on a live DB can have unpredicted results).`
analyzeCmd           = cli.Command("analyze", "analyze churn, label pair cardinality (running on a live DB can have unpredicted results).")

brian-brazil:
That seems dangerous, a read only Open method that didn't load the WAL would be safer.

krasi-georgiev:
Aaaah , now I understand what you mean I completely forgot that opening the DB also loads the WAL file.

Once we have the [noop wal PR](https://github.com/prometheus/tsdb/pull/519) merged this would be possible.

glightfoot:
Ok, I'll wait until #519 is merged and then modify to use that instead

brian-brazil:
There's a bit more to it than that, compaction and repair also needs disabling.

krasi-georgiev:
compaction can be disabled with `db.DisableCompactions()`.

@brian-brazil are you referring to WAL repair? This is also disabled when you pass a nil WAL.

brian-brazil:
No, the index repair. DisableCompactions would also be racy.

glightfoot:
What about adding a ReadOnly field to the db.Options struct? Or would you rather have individual options for disabling specific features (NopWal, NoCompact, NoRepair, etc) and add an OpenReadOnly method to the db? I'm happy to do the work if we can identify all the areas that need changing

brian-brazil:
A ReadOnly option sounds good to me.

krasi-georgiev:
ok lets try with the ReadOnly option when opening a DB to see how it would look like in code.

glightfoot:
@krasi-georgiev - Roughly like that. Still needs #519 merged before testing and finishing up the work on it

krasi-georgiev:
yep it doesn't look too bad. Would also need a test to ensure that opening a db with ReadOnly does indeed disable all mutating actions.

For example 
- open the db with a corrupt index and ensure it is not repaired (testdata/repair_index_version)
- Ensure that the compaction is disabled - 
```
testutil.Equals(t, 0.0, prom_testutil.ToFloat64(db.metrics.compactionsTriggered,"...")
testutil.Equals(t, 0.0, prom_testutil.ToFloat64(db.metrics.compactionsSkipped,"...")

db.compactc <- struct{}{}
testutil.Equals(t, 1.0, prom_testutil.ToFloat64(db.metrics.compactionsTriggered,"...")
testutil.Equals(t, 1.0, prom_testutil.ToFloat64(db.metrics.compactionsSkipped, "...")
```

- Ensure that no wal is created.

krasi-georgiev:
this other [PR](https://github.com/prometheus/tsdb/pull/532) reminded me that the size/time retention is another thing that need to be disabled

krasi-georgiev:
ping @glightfoot

krasi-georgiev:
@glightfoot 	I hope you don't mind, but since I don't see any activity here will continue the work in https://github.com/prometheus/tsdb/pull/541

PR #519, opened 2019-02-07
Added ability to create db with NopWal

### LINKED ISSUE #452: Ability to create db with NopWAL
We create a test db in prometheus for unit-testing and general tests. Ideally we never write anything to disk there but turns out we do because we cannot disable the WAL.

vn-ki:
Closes #452 

Signed-off-by: Vishnunarayan K I <appukuttancr@gmail.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## master / unreleased
  - [REMOVED] `FromData` is considered unused so was removed.
+ - [FEATURE] Added option WALSegmentSize -1 to disable the WAL.
 
 ## 0.6.1
   - [BUGFIX] Update `last` after appending a non-overlapping chunk in `chunks.MergeOverlappingChunks`. [#539](https://github.com/prometheus/tsdb/pull/539)
@@ -14,7 +15,7 @@
    - Added `MergeOverlappingChunks` function in `chunks/chunks.go` to merge multiple time-overlapping Chunk Metas.
    - Added `MinTime` and `MaxTime` method for `BlockReader`.
  - [FEATURE] New `dump` command to tsdb tool to dump all samples.
- - [FEATURE] New `encoding` package for common binary encoding/decoding helpers. 
+ - [FEATURE] New `encoding` package for common binary encoding/decoding helpers.
     - Added to remove some code duplication.
  - [ENHANCEMENT] When closing the db any running compaction will be cancelled so it doesn't block.
    - `NewLeveledCompactor` takes a context.
@@ -40,7 +41,7 @@
  - [CHANGE] `LastCheckpoint()` used to return just the segment name and now it returns the full relative path.
  - [CHANGE] `NewSegmentsRangeReader()` can now read over miltiple wal ranges by using the new `SegmentRange{}` struct.
  - [CHANGE] `CorruptionErr{}` now also exposes the Segment `Dir` which is added when displaying any errors.
- - [CHANGE] `Head.Init()` is changed to `Head.Init(minValidTime int64)`  
+ - [CHANGE] `Head.Init()` is changed to `Head.Init(minValidTime int64)`
  - [CHANGE] `SymbolTable()` renamed to `SymbolTableSize()` to make the name consistent with the  `Block{ symbolTableSize uint64 }` field.
  - [CHANGE] `wal.Reader{}` now exposes `Segment()` for the current segment being read  and `Offset()` for the current offset.
  - [FEATURE] tsdbutil analyze subcomand to find churn, high cardinality, etc.
--- a/db.go
+++ b/db.go
@@ -54,7 +54,10 @@ var DefaultOptions = &Options{
 
 // Options of the DB storage.
 type Options struct {
-	// Segments (wal files) max size
+	// Segments (wal files) max size.
+	// WALSegmentSize = 0, segment size is default size.
+	// WALSegmentSize > 0, segment size is WALSegmentSize.
+	// WALSegmentSize < 0, wal is disabled.
 	WALSegmentSize int
 
 	// Duration of persisted data to keep.
@@ -288,14 +291,20 @@ func Open(dir string, l log.Logger, r prometheus.Registerer, opts *Options) (db 
 	}
 	db.compactCancel = cancel
 
+	var wlog *wal.WAL
 	segmentSize := wal.DefaultSegmentSize
-	if opts.WALSegmentSize > 0 {
-		segmentSize = opts.WALSegmentSize
+	// Wal is enabled.
+	if opts.WALSegmentSize >= 0 {
+		// Wal is set to a custom size.
+		if opts.WALSegmentSize > 0 {
+			segmentSize = opts.WALSegmentSize
+		}
+		wlog, err = wal.NewSize(l, r, filepath.Join(dir, "wal"), segmentSize)
+		if err != nil {
+			return nil, err
+		}
 	}
-	wlog, err := wal.NewSize(l, r, filepath.Join(dir, "wal"), segmentSize)
-	if err != nil {
-		return nil, err
-	}
+
 	db.head, err = NewHead(r, l, wlog, opts.BlockRanges[0])
 	if err != nil {
 		return nil, err
--- a/db_test.go
+++ b/db_test.go
@@ -742,29 +742,53 @@ func TestWALFlushedOnDBClose(t *testing.T) {
 	testutil.Equals(t, []string{"labelvalue"}, values)
 }
 
-func TestWALSegmentSizeOption(t *testing.T) {
-	options := *DefaultOptions
-	options.WALSegmentSize = 2 * 32 * 1024
-	db, delete := openTestDB(t, &options)
-	defer delete()
-	app := db.Appender()
-	for i := int64(0); i < 155; i++ {
-		_, err := app.Add(labels.Labels{labels.Label{Name: "wal", Value: "size"}}, i, rand.Float64())
-		testutil.Ok(t, err)
-		testutil.Ok(t, app.Commit())
+func TestWALSegmentSizeOptions(t *testing.T) {
+	tests := map[int]func(dbdir string, segmentSize int){
+		// Default Wal Size.
+		0: func(dbDir string, segmentSize int) {
+			files, err := ioutil.ReadDir(filepath.Join(dbDir, "wal"))
+			testutil.Ok(t, err)
+			for _, f := range files[:len(files)-1] {
+				testutil.Equals(t, int64(DefaultOptions.WALSegmentSize), f.Size(), "WAL file size doesn't match WALSegmentSize option, filename: %v", f.Name())
+			}
+			lastFile := files[len(files)-1]
+			testutil.Assert(t, int64(DefaultOptions.WALSegmentSize) > lastFile.Size(), "last WAL file size is not smaller than the WALSegmentSize option, filename: %v", lastFile.Name())
+		},
+		// Custom Wal Size.
+		2 * 32 * 1024: func(dbDir string, segmentSize int) {
+			files, err := ioutil.ReadDir(filepath.Join(dbDir, "wal"))
+			testutil.Assert(t, len(files) > 1, "current WALSegmentSize should result in more than a single WAL file.")
+			testutil.Ok(t, err)
+			for _, f := range files[:len(files)-1] {
+				testutil.Equals(t, int64(segmentSize), f.Size(), "WAL file size doesn't match WALSegmentSize option, filename: %v", f.Name())
+			}
+			lastFile := files[len(files)-1]
+			testutil.Assert(t, int64(segmentSize) > lastFile.Size(), "last WAL file size is not smaller than the WALSegmentSize option, filename: %v", lastFile.Name())
+		},
+		// Wal disabled.
+		-1: func(dbDir string, segmentSize int) {
+			if _, err := os.Stat(filepath.Join(dbDir, "wal")); !os.IsNotExist(err) {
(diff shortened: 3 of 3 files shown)


krasi-georgiev:
maybe can use the segmentSize ? when set to 0 it would be essentially No Wal.

vn-ki:
> maybe can use the segmentSize ? when set to 0 it would be essentially No Wal.

@krasi-georgiev When I looked the function `wal.NewSize`, I saw that it created directories. The current implementation avoids that. I'm not sure what would happen if `segmentSize` is set to 0. What do you think? Should I change it to use `segmentSize`?

vn-ki:
@krasi-georgiev Updated. But the test errors still exists.

krasi-georgiev:
leave the test for the end, but the logic is wrong.

we should have 3 cases - default wal size , no wal, custom wal size.

vn-ki:
> we should have 3 cases - default wal size , no wal, custom wal size.

I'm sorry. I don't understand. I have to add `NopWal` back in? And chack that as well as `WalSegmentSize`? 

What is the third case?

krasi-georgiev:
I think something like this will work 
db.opts.WALSegmentSize == 0 : use the default size
db.opts.WALSegmentSize == -1 : wal is nil or disabled
db.opts.WALSegmentSize > 0 : wal is set to custom size.


about the test dig deeper and if you still can't figure it out by the end of next week ping me and I will try to help.

vn-ki:
I have updated it. Please check if this is what you mean.

> about the test dig deeper and if you still can't figure it out by the end of next week ping me and I will try to help.

Okay. Thank you.

I took a look and I think it is because some tests check for the wal files that is created. Two of the failed tests create tmp dirs for this check.

Should I remove these checks (if they're tested in `wal_test.go`) or make a different `db.Option` with wal for these tests?

krasi-georgiev:
> I have updated it. Please check if this is what you mean.

yes and also update the comment in the options about the 3 possible cases.
https://github.com/prometheus/tsdb/pull/519/files#diff-7a117cbfd5485ed35efcbb2995c37aa0R55

>I took a look and I think it is because some tests check for the wal files that is created. Two of the failed tests create tmp dirs for this check.
Should I remove these checks (if they're tested in `wal_test.go`) or make a different `db.Option` with wal for these tests?

Without looking at the tests can't tell, but study the code and if you are still unsure next week ping me and I will try to help.

krasi-georgiev:
also don't forget that need to add a test to ensure this behaviour - setting the no WAL option should not create a WAL file.

maybe can add it to 
https://github.com/prometheus/tsdb/blob/9f28ffa6f40a3d79a3eca28c775324eb65d354a4/db_test.go#L749

krasi-georgiev:
ping me when this is ready for another review

vn-ki:
@krasi-georgiev Please take a look.

vn-ki:
@krasi-georgiev The travis seems unrelated.

```
Attempt to get headers for https://sourceforge.net/projects/ezwinports/files/make-4.2.1-without-guile-w32-bin.zip/download failed.
  The remote file either doesn't exist, is unauthorized, or is forbidden for url 'https://sourceforge.net/projects/ezwinports/files/make-4.2.1-without-guile-w32-bin.zip/download'. Exception calling "GetResponse" with "0" argument(s): "Unable to connect to the remote server"
Downloading make 64 bit
  from 'https://sourceforge.net/projects/ezwinports/files/make-4.2.1-without-guile-w32-bin.zip/download'
ERROR: The remote file either doesn't exist, is unauthorized, or is forbidden for url 'https://sourceforge.net/projects/ezwinports/files/make-4.2.1-without-guile-w32-bin.zip/download'. Exception calling "GetResponse" with "0" argument(s): "Unable to connect to the remote server" 
This package is likely not broken for licensed users - see https://chocolatey.org/docs/features-private-cdn.
  make may be able to be automatically uninstalled.
The install of make was NOT successful.
Error while running 'C:\ProgramData\chocolatey\lib\make\tools\chocolateyinstall.ps1'.

```

krasi-georgiev:
I would say better to refactor TestWALSegmentSizeOption to check all 3  WALSegmentSize variants.

The required setup can be wrapped in a bootstrap function which takes the wal options as a part and check that the output matched the expected wal segment size.

have a look at `TestDeleteCompactionBlockAfterFailedReload`

krasi-georgiev:
@vn-ki are you still willing to continue the work on this PR?

vn-ki:
@krasi-georgiev Yes, I'm still interested in working on this. I'm having mid sems now. So got caught up in that. I'll look into this later today.

db.go
```diff
@@ -267,9 +270,14 @@ func Open(dir string, l log.Logger, r prometheus.Registerer, opts *Options) (db
 	if opts.WALSegmentSize > 0 {
 		segmentSize = opts.WALSegmentSize
 	}
-	wlog, err := wal.NewSize(l, r, filepath.Join(dir, "wal"), segmentSize)
-	if err != nil {
-		return nil, err
+	var wlog *wal.WAL
+	if opts.WALSegmentSize == -1 {
+		wlog = nil
```
krasi-georgiev: can drop this as `var wlog *wal.WAL` sets it to nil anyway.

can be just 
```
// check that the wal is not disabled.
if opts.WALSegmentSize != -1 {
	wlog, err = wal.NewSize(l, r, filepath.Join(dir, "wal"), segmentSize)
	if err != nil {
		return nil, err
	}
}
```

krasi-georgiev:
@vn-ki 	when do you think you can continue with this as it is blocking PR https://github.com/prometheus/tsdb/pull/541 ?

vn-ki:
I don't think I can look into this before this weekend. Sorry for blocking
the other PR. Feel free to close this if you want the other PR merged
before the weekend.

krasi-georgiev:
The weekend is fine I think.

vn-ki:
@krasi-georgiev I have updated the PR. Please see if this is what you meant.

db_test.go
```diff
@@ -743,28 +743,44 @@ func TestWALFlushedOnDBClose(t *testing.T) {
...
+			for i, f := range files {
+				if len(files)-1 != i {
+					testutil.Equals(t, int64(options.WALSegmentSize), f.Size(), "WAL file size doesn't match WALSegmentSize option, filename: %v", f.Name())
+					continue
+				}
+				testutil.Assert(t, int64(options.WALSegmentSize) > f.Size(), "last WAL file size is not smaller than the WALSegmentSize option, filename: %v", f.Name())
+			}
+		},
+		-1: func(dbDir string, options Options) {
+			if _, err := os.Stat(filepath.Join(dbDir, "wal")); os.IsNotExist(err) {
```
krasi-georgiev: can't this be something like: 
```
if _, err := os.Stat(filepath.Join(dbDir, "wal")); !os.IsNotExist(err) {
	t.Fatal("wal directory is present when the wal is disabled")
}
```

db_test.go
```diff
@@ -743,28 +743,44 @@ func TestWALFlushedOnDBClose(t *testing.T) {
...
-	options.WALSegmentSize = 2 * 32 * 1024
-	db, delete := openTestDB(t, &options)
-	defer delete()
-	app := db.Appender()
-	for i := int64(0); i < 155; i++ {
-		_, err := app.Add(labels.Labels{labels.Label{Name: "wal", Value: "size"}}, i, rand.Float64())
-		testutil.Ok(t, err)
-		testutil.Ok(t, app.Commit())
+	tests := map[int]func(dbdir string, options Options){
+		2 * 32 * 1024: func(dbDir string, options Options) {
```
krasi-georgiev: some comments for each test case would be a nice reminder.

// Custom wal size.

// Default wal size.

// WAL is disabled.

vn-ki:
@krasi-georgiev Done

db_test.go
```diff
@@ -743,28 +743,56 @@ func TestWALFlushedOnDBClose(t *testing.T) {
...
+		// Test Default Wal Size
+		0: func(dbDir string, options Options) {
+			files, err := ioutil.ReadDir(filepath.Join(dbDir, "wal"))
+			testutil.Ok(t, err)
+			for i, f := range files {
+				if len(files)-1 != i {
+					testutil.Equals(t, int64(DefaultOptions.WALSegmentSize), f.Size(), "WAL file size doesn't match WALSegmentSize option, filename: %v", f.Name())
+					continue
+				}
+				testutil.Assert(t, int64(DefaultOptions.WALSegmentSize) > f.Size(), "last WAL file size is not smaller than the WALSegmentSize option, filename: %v", f.Name())
```
krasi-georgiev: can be simplified a bit for both cases.

```
for _, f := range files[:len(files)-1] {
	testutil.Equals(t, int64(DefaultOptions.WALSegmentSize), f.Size(), "WAL file size doesn't match WALSegmentSize option, filename: %v", f.Name())
}
lastFile := files[len(files)-1]
testutil.Assert(t, int64(DefaultOptions.WALSegmentSize) > lastFile.Size(), "last WAL file size is not smaller than the WALSegmentSize option, filename: %v", lastFile.Name())
```

db.go
```diff
@@ -292,9 +295,13 @@ func Open(dir string, l log.Logger, r prometheus.Registerer, opts *Options) (db
 	if opts.WALSegmentSize > 0 {
 		segmentSize = opts.WALSegmentSize
 	}
-	wlog, err := wal.NewSize(l, r, filepath.Join(dir, "wal"), segmentSize)
-	if err != nil {
-		return nil, err
+	fmt.Printf("%d", segmentSize)
```
krasi-georgiev:
```suggestion
```

db_test.go
```diff
@@ -743,28 +743,56 @@ func TestWALFlushedOnDBClose(t *testing.T) {
...
+			}
+		},
+		// Test Custom Wal Size
```
krasi-georgiev: full stops for all comments and no need to add `Test` for all comments.
```suggestion
		// Custom Wal Size.
```

vn-ki:
@krasi-georgiev Done

db_test.go
```diff
@@ -743,28 +743,52 @@ func TestWALFlushedOnDBClose(t *testing.T) {
...
-	options := *DefaultOptions
-	options.WALSegmentSize = 2 * 32 * 1024
-	db, delete := openTestDB(t, &options)
-	defer delete()
-	app := db.Appender()
-	for i := int64(0); i < 155; i++ {
-		_, err := app.Add(labels.Labels{labels.Label{Name: "wal", Value: "size"}}, i, rand.Float64())
-		testutil.Ok(t, err)
-		testutil.Ok(t, app.Commit())
+	tests := map[int]func(dbdir string, options Options){
```
krasi-georgiev: can just use the segmentSize as an int, no need to pass all options.

```suggestion
	tests := map[int]func(dbdir string, segmentSize int){
```

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## master / unreleased
  - [REMOVED] `FromData` is considered unused so was removed.
+ - [FEATURE] Feature to create db with WAL.
```
krasi-georgiev:
```suggestion
 - [FEATURE] Added option WALSegmentSize -1 to disable the WAL.
```

db_test.go
```diff
@@ -743,28 +743,52 @@ func TestWALFlushedOnDBClose(t *testing.T) {
 }
 
 func TestWALSegmentSizeOption(t *testing.T) {
```
```suggestion
func TestWALSegmentSizeOptions(t *testing.T) {
```

db.go
```diff
@@ -55,6 +55,9 @@ var DefaultOptions = &Options{
 // Options of the DB storage.
 type Options struct {
 	// Segments (wal files) max size
+	// When WALSegmentSize is 0, segment size is default size
```
krasi-georgiev:
```
// Segments (wal files) max size.
// WALSegmentSize = 0, segment size is default size.
// WALSegmentSize > 0, segment size is WALSegmentSize.
// WALSegmentSize = -1, wal is disabled.
```
full stops and removed some repetitive words.

krasi-georgiev:
LGTM

ping @codesome

db.go
```diff
@@ -54,7 +54,10 @@ var DefaultOptions = &Options{
 
 // Options of the DB storage.
 type Options struct {
-	// Segments (wal files) max size
+	// Segments (wal files) max size.
+	// WALSegmentSize = 0, segment size is default size.
```
codesome: Right now it takes the default size for `WALSegmentSize < -1` too. Should we make it disabled for `WALSegmentSize < 0` then?

krasi-georgiev: yep , not a bad idea. I like it.

db.go
```diff
@@ -292,9 +295,13 @@ func Open(dir string, l log.Logger, r prometheus.Registerer, opts *Options) (db
 	if opts.WALSegmentSize > 0 {
 		segmentSize = opts.WALSegmentSize
 	}
-	wlog, err := wal.NewSize(l, r, filepath.Join(dir, "wal"), segmentSize)
-	if err != nil {
-		return nil, err
+	var wlog *wal.WAL
+	if opts.WALSegmentSize >= 0 {
```
krasi-georgiev:
```
	// Wal is enabled.
	if opts.WALSegmentSize >= 0 {
		// Wal is set to a custom size.
		if opts.WALSegmentSize > 0 {
			segmentSize = opts.WALSegmentSize
		}	
		wlog, err = wal.NewSize(l, r, filepath.Join(dir, "wal"), segmentSize)
		if err != nil {
			return nil, err
		}	
	}
```

vn-ki: I'm so sorry. I somehow did not see this. I have updated.

krasi-georgiev:
Thanks

PR #521, opened 2019-02-13
Update encoding_helpers.go (len of be64 should be 8)

naivewong:


krasi-georgiev:
Thanks

brian-brazil:
Should we have a unittest for this?

krasi-georgiev:
I have been very influenced by this talk https://www.youtube.com/watch?v=EZ05e7EMOLM 
In short for easy long term maintenance I push for testing behaviour and not implementation. 
In this case the same can be applied here by reading a corrupted index file and ensuring that the corresponding error is returned and it doesn't panic.

@naivewong how do you feel about adding something like this or adding it somewhere to the existing tests?

naivewong:
@krasi-georgiev I've re-checked that, but it seems that the situation that reading a corrupted index file will return the error in be64() can never be reproduced in the test code. 

The be64() currently is currently called in [NewTOCFromByteSlice()](https://github.com/prometheus/tsdb/blob/master/index/index.go#L150) and the be64() returns error only when `len(d.b) < 8`, and because of `b := bs.Range(bs.Len()-indexTOCLen, bs.Len())` in NewTOCFromByteSlice() which means the size of b is always fixed and it is safe to call be64(), we cannot generate that error by simply cutting the index file.

But instead, if you want, I can add some test codes into index_test.go simply for this function.
Or if you have other suggestions?

krasi-georgiev:
just had another look and it seems the best bit to be tested is the `RecordDecoder` or something higher in the API (maybe the WAL) that uses the `RecordDecoder`

There is already a test for the `RecordDecoder` so I would try there first.
https://github.com/prometheus/tsdb/blob/109252f3aa66b954d237978b5fac7a4e57edf4da/record_test.go#L24

PR #522, opened 2019-02-13
clarify which segments are deleted when we find a corrupted segment

cstyan:
The previous version of this log line can be misinterpreted as "segments before corruption" as in older segments not newer segments.

`level=warn ts=2019-02-13T17:09:48.066786461Z caller=wal.go:299 component=tsdb msg="deleting all segments behind corruption" segment=6142`


krasi-georgiev:
Thanks

PR #523, opened 2019-02-14
Add test for be64() in encoding_helpers.go (#521)

naivewong:
Refer to PR #521, 
1. Update the be64() in encoding_helpers.go
2. Add test in record_test.go to test if be64() can return error correctly


record_test.go
```diff
@@ -71,3 +71,43 @@ func TestRecord_EncodeDecode(t *testing.T) {
...
 	}, decTstones)
 }
+
+// Refer to PR#521, this function is to test the be64() of decbuf in encoding_helpers.go
+// Provided corrupted records, test if the invalid size error can be returned by be64() correctly.
+func TestRecord_CorruputedRecord(t *testing.T) {
+	var enc RecordEncoder
+	var dec RecordDecoder
+
+	// Test corrupted sereis record
```
krasi-georgiev:
```
t.Run("Test corrupted series record", func(t *testing.T) {
		series := []RefSeries{
			{
				Ref:    100,
				Labels: labels.FromStrings("abc", "def", "123", "456"),
			},
		}

		corrupted := enc.Series(series, nil)[:4]
		_, err := dec.Series(corrupted, nil)
		testutil.Equals(t, err, errInvalidSize)
	})
```

the other part can be simplified the same way.

record_test.go
```diff
@@ -71,3 +72,49 @@ func TestRecord_EncodeDecode(t *testing.T) {
...
 	}, decTstones)
 }
+
+// Refer to PR#521, this function is to test the be64() of decbuf in encoding_helpers.go.
+// Provided corrupted records, test if the invalid size error can be returned by be64() correctly.
+func TestRecord_Corruputed(t *testing.T) {
+	var enc RecordEncoder
+	var dec RecordDecoder
+
+	// Test corrupted sereis record.
```
krasi-georgiev: comments are not needed when we have this in the following line.

encoding_helpers.go
```diff
@@ -120,7 +120,7 @@ func (d *decbuf) be64() uint64 {
 	if d.e != nil {
 		return 0
 	}
-	if len(d.b) < 4 {
+	if len(d.b) < 8 {
```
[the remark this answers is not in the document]
krasi-georgiev: did you see this comment?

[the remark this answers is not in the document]
naivewong: Yes, I change [:4] to [:5], so the original one will panic and the changed one will pass

record_test.go
```diff
@@ -71,3 +72,46 @@ func TestRecord_EncodeDecode(t *testing.T) {
 		{ref: 13, intervals: Intervals{{Mint: 5000, Maxt: 1000}}},
 	}, decTstones)
 }
+
+// Refer to PR#521, this function is to test the be64() of decbuf in encoding_helpers.go.
+// Provided corrupted records, test if the invalid size error can be returned by be64() correctly.
```
krasi-georgiev:
```
// TestRecord_Corruputed ensures that corrupted records return the correct error.
// Bugfix check for  pull/521 and pull/523
```

just to be closer to the golang comment standards.

record_test.go
```diff
@@ -71,3 +72,46 @@ func TestRecord_EncodeDecode(t *testing.T) {
...
+
+	t.Run("Test corrupted series record", func(t *testing.T) {
+		series := []RefSeries{
+			{
+				Ref:    100,
+				Labels: labels.FromStrings("abc", "def", "123", "456"),
+			},
+		}
+
+		corrupted := enc.Series(series, nil)[:5]
```
krasi-georgiev: shall we make these [:8] to be closer to what is expcted?

krasi-georgiev:
LGTM

@codesome 	do you want to take a look.

record_test.go
```diff
@@ -71,3 +72,46 @@ func TestRecord_EncodeDecode(t *testing.T) {
 		{ref: 13, intervals: Intervals{{Mint: 5000, Maxt: 1000}}},
 	}, decTstones)
 }
+
+// TestRecord_Corruputed ensures that corrupted records return the correct error.
+// Bugfix check for pull/521 and pull/523.
```
codesome: This doesn't exactly test pull/521 though the code is the same. 

[Not related to this PR] On that point, why not merge `index/encoding_helpers.go` and `encoding_helpers.go`. Would it make sense to move it to `tsdbutil`? This can be done in another PR. /cc @krasi-georgiev

krasi-georgiev: > [Not related to this PR] On that point, why not merge `index/encoding_helpers.go` and `encoding_helpers.go`. Would it make sense to move it to `tsdbutil`? This can be done in another PR. /cc @krasi-georgiev

yes a good idea. @naivewong do you want to try and move both duplicates in tsdbutil? Do it in a separate PR and will merge that one first.

this way the same test will cover both bugs.

PR #524, opened 2019-02-14
LiveReader can get into an infinite loop on corrupt WALs.

tomwilkie:
Add unit tests for tailing a corrupt WAL, to demonstrate.  The test write two well-formed records to the WAL, but where the length of the two together are greater than 1 page.  This is invalid, but caused the live reader to infinite loop.

I fixed this by making the `fill` phase of the live reader return the number of bytes, and then always reporting the error that occurs when the we can't read any more.  Furthermore, I moved all of the handling of `r.err` into the next function, and made other functions return errors, to make the lifecycle of this field clearer.

Finally this triggered an issue with the handling of the zero padding at the end of the pages, so I refactored the `readRecord` function to be given the whole buffer and made it clearer how it was handling that.  This also opened up another way to catch the first issue, as now `readRecord` can detect records which might overflow a page.


wal/wal_test.go
```diff
@@ -311,37 +412,50 @@ func TestWAL_FuzzWriteRead_Live(t *testing.T) {
...
+			if last <= seg.i {
+				continue
 			}
+
+			// read to end of segment.
+			readSegment(r)
+
+			fi, err := os.Stat(SegmentName(dir, seg.i))
+			testutil.Ok(t, err)
+			testutil.Assert(t, r.TotalRead() == fi.Size(), "expected to have read whole segment, but read %d or %d", r.TotalRead(), fi.Size())
```
cstyan: or -> of

wal/wal.go
```diff
@@ -1042,71 +1037,44 @@ func validateRecord(typ recType, i int) error {
...
-		// We are pedantic and check whether the zeros are actually up to a page boundary.
-		// It's not strictly necessary but may catch sketchy state early.
-		k := pageSize - (total % pageSize)
-		if k == pageSize {
-			return nil, 1, nil // Initial 0 byte was last page byte.
+func readRecord(buf []byte, start, end int, header []byte) ([]byte, int, error) {
+	// Special case: for recPageTerm, check that are all zeros to end of page,
+	// consume them but don't return them.
+	if buf[start] == byte(recPageTerm) {
+		if end != pageSize {
```
cstyan: this probably works but is assuming we're always reading uncorrupted pages and so `recPageTerm` would always end at the end of the internal buffer

tomwilkie:
Seems to be a platform-specific bug now, passed on everything but linux and all the failures are of the form "wal_test.go:443: expected to have read whole segment, but read 3833856 or 4161536".  This is testing the assumption that reads won't EOF once there is a new segment.  Looking at `nextSegment`, the code fsyncs and closes the previous segment after starting the new one - so this assumption is invalid.

Thoughts?

wal/wal.go
```diff
@@ -1042,71 +1037,44 @@ func validateRecord(typ recType, i int) error {
...
-	recData := buf[readIndex:readTo]
-	readIndex += int(length)
-	total += int64(length)
 
-	// TODO(callum) what should we do here, throw out the record? We should add a metric at least.
-	if c := crc32.Checksum(recData, castagnoliTable); c != crc {
-		return recData, readIndex, errors.Errorf("unexpected checksum %x, expected %x", c, crc)
+	rec := buf[start+recordHeaderSize : start+recordHeaderSize+length]
+	if c := crc32.Checksum(rec, castagnoliTable); c != crc {
+		return nil, 0, errors.Errorf("unexpected checksum %x, expected %x", c, crc)
```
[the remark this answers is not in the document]
tomwilkie: I thought I had changed `Next()`, so any non-EOF error got reported straight away.  But I had all errors being reported straight away.  Either way this error won't cause a loop IIUC. 

 I'll make sure the EOFs from buildRecord cause another read though.

wal/wal.go
tomwilkie: `recPageTerm` should always end with the end of a page, and anything else is corrupt.  I think thats what this code honours?  If we see `recPageTerm` without a full page, we return EOF.  If we see `recPageTerm` and a full page, we check they are zero.  If they aren't, its corrupt, and we return the error.  Is that right?

wal/wal.go
```diff
@@ -965,41 +963,38 @@ func (r *LiveReader) Record() []byte {
...
-		r.readIndex += n
-		r.total += int64(n)
+		temp, n, err := readRecord(r.buf[:], r.readIndex, r.writeIndex, r.hdr[:])
 		if err != nil {
-			r.err = err
-			return false
+			return false, err
 		}
 
+		r.readIndex += n
```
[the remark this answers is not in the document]
tomwilkie: Which comment?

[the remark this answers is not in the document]
tomwilkie: Added more comments to explain.

tomwilkie:
"expected to have read whole segment" is fixed; we were returning EOF when we didn't have enough data, and should have been trying to read more.  Still concerned about sequencing of syncs && closes, but I think its okay as we're in the same process.

wal/wal.go
```diff
@@ -1042,71 +1047,44 @@ func validateRecord(typ recType, i int) error {
...
-		}
-		// Not enough data to read all of the record data.
-		// Treat this the same as an EOF, it's an error we would expect to see.
+	copy(header, buf[start:start+recordHeaderSize])
+	length := int(binary.BigEndian.Uint16(header[1:]))
+	crc := binary.BigEndian.Uint32(header[3:])
+	if start+recordHeaderSize+length > pageSize {
+		return nil, 0, errors.New("record too big for page")
+	}
+	if start+recordHeaderSize+length > end {
```
cstyan: shouldn't this be a terminal error? the issue we saw yesterday was a segment (supposedly not corrupted according to the head initialization) that had a record whose data (unchunked) would pass the page boundary

wal/wal.go
```diff
@@ -908,47 +909,54 @@ type LiveReader struct {
...
+		if ok, err := r.buildRecord(); ok {
 			return true
-		}
-		if r.err != nil && r.err != io.EOF {
+		} else if err != nil && err != io.EOF {
+			r.err = err
 			return false
 		}
+
+		// If we've consumed the whole page, reset the indexed and continue.
```
cstyan: indexed -> indexes? or indices?

wal/wal.go
cstyan: actually looks like the previous if case is supposed to catch that

tomwilkie: +1 year, terminal error above, EOF otherwise.

wal/wal.go
```diff
@@ -908,47 +909,54 @@ type LiveReader struct {
...
-	r.readIndex = 0
-	r.writeIndex = copied
+	return n, err
 }
 
-// Next returns true if r.rec will contain a full record.
-// False does not indicate that there will never be more data to
-// read for the current io.Reader.
+// Next returns true if Record() will contain a full record.
+// False guarantees there are no more records if the segment is closed,
```
cstyan: this is true except in the case of potentially corrupted segments like the one we've been looking at (6141), but we're still undecided if that segment is the result of a bug

wal/wal.go
```diff
@@ -908,47 +909,54 @@ type LiveReader struct {
...
+// otherwise if Err() == io.EOF you should try again when more data
+// has been written.
 func (r *LiveReader) Next() bool {
 	for {
-		if r.buildRecord() {
+		// If buildRecord returns a non-EOF error, its game up - the WAL is
```
cstyan: should this refer to current segment rather than WAL, since right now we only pass it an io.Reader for a single segment?

wal/wal.go
```diff
@@ -1042,71 +1047,44 @@ func validateRecord(typ recType, i int) error {
...
-		if (readTo - readIndex) > pageSize {
-			return nil, 0, errors.Errorf("invalid record, record size would be larger than max page size: %d", int(length))
-		}
-		// Not enough data to read all of the record data.
-		// Treat this the same as an EOF, it's an error we would expect to see.
+	copy(header, buf[start:start+recordHeaderSize])
+	length := int(binary.BigEndian.Uint16(header[1:]))
+	crc := binary.BigEndian.Uint32(header[3:])
+	if start+recordHeaderSize+length > pageSize {
+		return nil, 0, errors.New("record too big for page")
```
cstyan: this error should be "record would overflow current page"? and we could include info like length and start+recordHeader+length?

`return nil, 0, fmt.Errorf("record would overflow current page: %d > %d", start+recordHeaderSize+length, pageSize)`?

wal/wal.go
```diff
@@ -894,8 +894,9 @@ func NewLiveReader(r io.Reader) *LiveReader {
 	return &LiveReader{rdr: r}
 }
 
-// Reader reads WAL records from an io.Reader. It buffers partial record data for
-// the next read.
+// Reader reads WAL records from an io.Reader. It allows reading of WALs
```
gouthamve: `Reader` --> `LiveReader`

wal/wal.go
gouthamve: +1

wal/wal.go
gouthamve: +1

cstyan:
seems I actually can't push to this branch:
```
diff --git a/wal/wal.go b/wal/wal.go
index 72f22e8..4fab1c3 100644
--- a/wal/wal.go
+++ b/wal/wal.go
@@ -894,7 +894,7 @@ func NewLiveReader(r io.Reader) *LiveReader {
        return &LiveReader{rdr: r}
 }

-// Reader reads WAL records from an io.Reader. It allows reading of WALs
+// LiveReader reads WAL records from an io.Reader. It allows reading of WALs
 // that are still in the process of being written, and returns records
 // as soon as they can read.
 type LiveReader struct {
@@ -934,10 +934,10 @@ func (r *LiveReader) fillBuffer() (int, error) {
 // has been written.
 func (r *LiveReader) Next() bool {
        for {
-               // If buildRecord returns a non-EOF error, its game up - the WAL is
-               // corrupt. If buildRecord returns an EOF, we try and read more. If
-               // that fails to read anything (n=0 && err=EOF), we return EOF and
-               // the user can try again later.
+               // If buildRecord returns a non-EOF error, its game up - the current segment
+               // is corrupt. If buildRecord returns an EOF, we try and read more. If that
+               // fails to read anything (n=0 && err=EOF), we return EOF and the user can
+               // try again later.
                // If we have a full page, buildRecord is guaranteed to return non-EOF
                // or a record; it has checks the records fit in pages.
                if ok, err := r.buildRecord(); ok {
@@ -1073,7 +1073,7 @@ func readRecord(buf []byte, start, end int, header []byte) ([]byte, int, error)
        length := int(binary.BigEndian.Uint16(header[1:]))
        crc := binary.BigEndian.Uint32(header[3:])
        if start+recordHeaderSize+length > pageSize {
-               return nil, 0, errors.New("record too big for page")
+               return nil, 0, fmt.Errorf("record would overflow current page: %d > %d", start+recordHeaderSize+length, pageSize)
        }
        if start+recordHeaderSize+length > end {
                return nil, 0, io.EOF
```

cstyan:
@tomwilkie  @gouthamve So after some digging, and then just going back to something I had in the reader in the past, there's nothing actually wrong with the data in the WAL segments that the `LiveReader` is having issues with. 

I still don't know how Repair is getting us in to a state where somewhere within the segment page boundaries get messed up.

So all I did was add `shiftBuffer` back, change the rec length check to this: 
```
if start+recordHeaderSize+length > pageSize {
	if length > pageSize {
		return nil, 0, fmt.Errorf("record too big for page: %d > %d", start+recordHeaderSize+length, pageSize)
	}
	return nil, 0, io.EOF
}
```
and then we can read these seemingly broken but already repaired segments just fine.

If we changed the code to swallow errors when replaying segments, what would happen right now (I think, because this is what happens with the data I copied off that prometheus instance yesterday) is that a corruption will be detected in 6142 and so everything after it will be deleted, leaving it as the newest segment. The repair leaves it in a state that the current live reader thinks is broken, and then we're stuck trying to read that segment forever. We replay 6141, try to start tailing 6142 and fail at some point, repeat.

tomwilkie:
> here's nothing actually wrong with the data in the WAL segments that the `LiveReader` is having issues with.

Unless we can find out why there are records spanning pages, I still think this is a bug.  Its pretty clear to me the intention of the WAL code is for records to not span pages.

> So all I did was add `shiftBuffer` back

I don't follow - shiftBuffer was pretty much a noop, it only shifted when we had consumed a whole page, and as such, never actually shifted anything.

> If we changed the code to swallow errors when replaying segments ... corruption will be detected in 6142 and so everything after it will be deleted,

I'm not suggesting in the WAL code we should delete anything; just that we should ignore (log) errors when replaying at startup and move on to next segment.

tomwilkie:
I couldn't find a reason why this was happening, and it doesn't happen on my local Prometheus.  So I made LiveReader tolerate records spanning pages.  Not happy about it though...

I did unify the interfaces & tests, which is nice.

wal/live_reader.go
```diff
@@ -0,0 +1,246 @@
...
+	buf        [pageSize]byte
+	readIndex  int   // Index in buf to start at for next read.
+	writeIndex int   // Index in buf to start at for next write.
+	total      int64 // Total bytes processed during reading in calls to Next().
+	index      int   // Used to track partial records, should be 0 at the start of every new record.
+
+	// For testing, we can treat EOF as a non-error.
+	eofNonErr bool
+
+	// We sometime see records span page boundaries.  Should never happen, but it
```
cstyan: I still think this might be because of the bugs in repair. If repair didn't fix a torn page and then we continue writing to that segment (without @gouthamve fix for `donePages`) we could then write part of a record into the end of existing page and the rest into a new page. The done pages issue could have messed up page boundaries up until the end of that segment. 

Can we add a metric and log line to count records we accept when this is set to true?

wal/live_reader.go
```diff
@@ -0,0 +1,246 @@
...
+// Next returns true if Record() will contain a full record.
+// If Next returns false, you should always checked the contents of Error().
+// Return false guarantees there are no more records if the segment is closed
+// and not corrupt, otherwise if Err() == io.EOF you should try again when more
+// data has been written.
+func (r *LiveReader) Next() bool {
+	for {
+		// If buildRecord returns a non-EOF error, its game up - the segment is
+		// corrupt. If buildRecord returns an EOF, we try and read more. If
+		// that fails to read anything (n=0 && err=EOF), we return EOF and
```
cstyan: I don't see where we explicitly read more if we get an `io.EOF` so this comment is a bit confusing. If you're referring to the call to `fillBuffer` later on can we just say that?

wal/live_reader.go
```diff
@@ -0,0 +1,246 @@
...
+// Read a sub-record (see recType) from the buffer. It could potentially
+// be a full record (recFull) if the record fits within the bounds of a single page.
+// Returns a byte slice of the record data read, the number of bytes read, and an error
+// if there's a non-zero byte in a page term record or the record checksum fails.
+// This is a non-method function to make it clear it does not mutate the reader.
+func readRecord(buf []byte, start, end int, header []byte, permissive bool) ([]byte, int, error) {
+	// Special case: for recPageTerm, check that are all zeros to end of page,
+	// consume them but don't return them.
+	if buf[start] == byte(recPageTerm) {
+		if end != pageSize {
```
cstyan: would this not break if we're now handling the records that span page boundaries?

wal/live_reader.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	copy(header, buf[start:start+recordHeaderSize])
+	length := int(binary.BigEndian.Uint16(header[1:]))
+	crc := binary.BigEndian.Uint32(header[3:])
+	if !permissive && start+recordHeaderSize+length > pageSize {
+		return nil, 0, fmt.Errorf("record would overflow current page: %d > %d", start+recordHeaderSize+length, pageSize)
+	}
+	if recordHeaderSize+length > pageSize {
+		return nil, 0, fmt.Errorf("record would overflow current page: %d > %d", start+recordHeaderSize+length, pageSize)
```
cstyan: this error message should probably be `invalid record; length of record is greater than a single page`

wal/reader_test.go
```diff
@@ -0,0 +1,486 @@
...
+	// Try and LiveReader it.
+	m, _, err := w.Segments()
+	testutil.Ok(t, err)
+
+	seg, err := OpenReadSegment(SegmentName(dir, m))
+	testutil.Ok(t, err)
+
+	r := NewLiveReader(seg)
+	testutil.Assert(t, r.Next() == false, "expected no records")
+	testutil.Assert(t, r.Err().Error() == "record would overflow current page: 65542 > 32768", "expected error, got: %v", r.Err())
```
cstyan: this error msg would need to change if you make the change above for `record is larger than page size`

wal/live_reader.go
tomwilkie: It would, yes.  Unfortunately, I don't see a solution - if we allow for unaligned read, how do we know when the page boundary is?

tomwilkie: Hmm, I guess we already track offset so I could use that.

cstyan: yeah that's how we were doing it before
```
// We are pedantic and check whether the zeros are actually 
// to a page boundary.	
// It's not strictly necessary but may catch sketchy state early.
k := pageSize - (r.total % pageSize)
```

PR #526, opened 2019-02-15
Merge encoding_helpers.go to tsdbutil

naivewong:


bwplotka:
Hey, what is the reason behind this? (: 

`util` in Golang packages are antipattern as described here: https://blog.golang.org/package-names#TOC_5 This is because it usually ends up being pile of mixed functions from everywhere and makes reading the code more confusing.

naivewong:
Oh sorry, I forgot to tag this PR. This PR would refer to #523.

And in this case, since the tsdbutil package already existed and the encoding_helpers.go contains two general tools (encbuf, decbuf) which are defined and used seperately in packcage index and tsdb(and maybe possible in the future packages), then I think maybe it's resonable to merge the duplicates to tsdbutil.

krasi-georgiev:
@bwplotka 	good point. Maybe a better option is to create a package called `encoding`.

naivewong:
@krasi-georgiev But for the current changes, is it ok to you? I may do the changes all together later

encoding/encoding_helpers.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+func (d *Decbuf) Err() error  { return d.E }
+func (d *Decbuf) Len() int    { return len(d.B) }
+func (d *Decbuf) Get() []byte { return d.B }
+
+// NewDecbufAt returns a new decoding buffer. It expects the first 4 bytes
+// after offset to hold the big endian encoded content length, followed by the contents and the expected
+// checksum.
+func NewDecbufAt(bs ByteSlice, off int, castagnoliTable *crc32.Table) Decbuf {
```
codesome: Better to move this just after the type declaration like before.

encoding/encoding_helpers.go
```diff
@@ -0,0 +1,244 @@
...
+	if exp := binary.BigEndian.Uint32(b[len(b)-4:]); DecbufCrc32(&dec, castagnoliTable) != exp {
+		return Decbuf{E: ErrInvalidChecksum}
+	}
+	return dec
+}
+
+// NewDecbufUvarintAt returns a new decoding buffer. It expects the first bytes
+// after offset to hold the uvarint-encoded buffers length, followed by the contents and the expected
+// checksum.
+func NewDecbufUvarintAt(bs ByteSlice, off int, castagnoliTable *crc32.Table) Decbuf {
```
codesome: Same, better to move this just after the type declaration like before.

encoding/encoding_helpers.go
```diff
@@ -0,0 +1,244 @@
...
+	dec := Decbuf{B: b[:len(b)-4]}
+
+	if DecbufCrc32(&dec, castagnoliTable) != binary.BigEndian.Uint32(b[len(b)-4:]) {
+		return Decbuf{E: ErrInvalidChecksum}
+	}
+	return dec
+}
+
+// DecbufCrc32 returns a CRC32 checksum over the remaining bytes.
+func DecbufCrc32(d *Decbuf, castagnoliTable *crc32.Table) uint32 {
```
codesome: This was a member function of `Decbuf` before and would make sense to keep it like that. (need to modify function calls with this change).
```suggestion
func (d *Decbuf) DecbufCrc32(castagnoliTable *crc32.Table) uint32 {
```

krasi-georgiev:
yes I would say move all relevant funcs inside a package called `encoding` or some other more relevant name.

naivewong:
@codesome @krasi-georgiev How about the current changes? and the two failed test jobs seem not related to my changes.

krasi-georgiev:
I don't see any failed checks?

the idea is to move all functions in its own package `encoding`

naivewong:
Oh seems travis ci rerun the tests several hours ago. And it seems that I have already moved all the functions to `encoding` package. What else left?

codesome:
@krasi-georgiev I re-ran the tests, they were flaky.

krasi-georgiev:
how about encoding/encoding_helpers.go -> encoding/encoding.go

krasi-georgiev:
actually it is fine as is. LGTM

krasi-georgiev:
Thanks!

PR #527, opened 2019-02-16
Avoid creation of 0 sized segments.

gouthamve:
If the corrupt segment is full, then we set donePages on open,
https://github.com/prometheus/tsdb/blob/c59ed492b284c1f5c09d9f27167b2d27df7d09c3/wal/wal.go#L235-L243

Then when we try to repair, we set the segment to be a new segment but
we don't update the donePages: https://github.com/prometheus/tsdb/blob/c59ed492b284c1f5c09d9f27167b2d27df7d09c3/wal/wal.go#L334

We we try to log to this, because donePages is full, we will never log
anything to this segment and create a new one: https://github.com/prometheus/tsdb/blob/c59ed492b284c1f5c09d9f27167b2d27df7d09c3/wal/wal.go#L486

This does not cause issues because we simply concatenate the segments on
read, there by transparently skipping this `0b` segment.

DONOT MERGE YET

### DIFF
--- a/wal/wal.go
+++ b/wal/wal.go
@@ -332,6 +332,7 @@ func (w *WAL) Repair(origErr error) error {
 		return err
 	}
 	w.segment = s
+	w.donePages = 0
 
 	f, err := os.Open(tmpfn)
 	if err != nil {
--- a/wal/wal.go
+++ b/wal/wal.go
@@ -228,19 +228,23 @@ func NewSize(logger log.Logger, reg prometheus.Registerer, dir string, segmentSi
 	}
 	// Fresh dir, no segments yet.
 	if j == -1 {
-		if w.segment, err = CreateSegment(w.dir, 0); err != nil {
-			return nil, err
-		}
-	} else {
-		if w.segment, err = OpenWriteSegment(logger, w.dir, j); err != nil {
-			return nil, err
-		}
-		// Correctly initialize donePages.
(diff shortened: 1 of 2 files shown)


tomwilkie:
LGTM for correctness, but yeah, this doesn't really cause issue AFAICT.

gouthamve:
@krasi-georgiev @tomwilkie @cstyan @codesome Can you review now?

cstyan:
lgtm, we might also want a repair test where the corruption is in a middle segment though?

tomwilkie:
> lgtm, we might also want a repair test where the corruption is in a middle segment though?

We added one in #528, but interestingly this fix didn't make a difference.  The test changes here do reproduce a problem though, which this fixes.

krasi-georgiev:
LGTM

wal/wal.go
```diff
@@ -332,6 +332,7 @@ func (w *WAL) Repair(origErr error) error {
 		return err
 	}
 	w.segment = s
+	w.donePages = 0
```
krasi-georgiev: how about instead of this add an internal func like w.setSegment to avoid this bug again.

gouthamve:
@krasi-georgiev That is a nice suggestion, did it.

krasi-georgiev:
LGTM, Thanks

PR #528, opened 2019-02-18
Test to corrupt segments mid-WAL, repair and check we can read the correct number of records.

tomwilkie:
The test fails against master, as the underlying segmentReader has already moved on to the next segment.

Also included - the fix.  Make the segmentReader pad short segments with zeros, and only advance to next segment on the call to `Read` _after_ completely consuming this one; previously it would increment `curr` on the call to `Read` which failed, causing the returns `CorruptionErr` to reference the wrong segment.

### DIFF
--- a/testutil/logging.go
+++ b/testutil/logging.go
@@ -10,10 +10,12 @@ type logger struct {
 	t *testing.T
 }
 
+// NewLogger returns a gokit compatible Logger which calls t.Log.
 func NewLogger(t *testing.T) log.Logger {
 	return logger{t: t}
 }
 
+// Log implements log.Logger.
 func (t logger) Log(keyvals ...interface{}) error {
 	t.t.Log(keyvals...)
 	return nil
--- a/testutil/logging.go
+++ b/testutil/logging.go
@@ -1,3 +1,16 @@
+// Copyright 2019 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package testutil
 
 import (
--- a/testutil/logging.go
+++ b/testutil/logging.go
@@ -1,0 +1,20 @@
+package testutil
+
+import (
+	"testing"
+
+	"github.com/go-kit/kit/log"
+)
+
+type logger struct {
+	t *testing.T
+}
+
+func NewLogger(t *testing.T) log.Logger {
+	return logger{t: t}
+}
+
+func (t logger) Log(keyvals ...interface{}) error {
+	t.t.Log(keyvals...)
+	return nil
+}
--- a/wal/wal.go
+++ b/wal/wal.go
@@ -717,7 +717,8 @@ func (r *segmentBufReader) Read(b []byte) (n int, err error) {
 		return n, err
 	}
 
-	// We hit EOF; fake out zero padding at the end of short segments.
+	// We hit EOF; fake out zero padding at the end of short segments, so we
+	// don't increment curr too early and report the wrong segment as corrupt.
 	if r.off%pageSize != 0 {
 		i := 0
 		for ; n+i < len(b) && (r.off+i)%pageSize != 0; i++ {
@@ -729,11 +730,13 @@ func (r *segmentBufReader) Read(b []byte) (n int, err error) {
 		return n + i, nil
 	}
 
-	// Now we actually need to move to next segment.
+	// There is no more deta left in the curr segment and there are no more
+	// segments left.  Return EOF.
 	if r.cur+1 >= len(r.segs) {
 		return n, io.EOF
 	}
 
+	// Move to next segment.
 	r.cur++
 	r.off = 0
 	r.buf.Reset(r.segs[r.cur])
--- a/wal/wal.go
+++ b/wal/wal.go
@@ -681,20 +681,19 @@ func NewSegmentsRangeReader(sr ...SegmentRange) (io.ReadCloser, error) {
 
 // segmentBufReader is a buffered reader that reads in multiples of pages.
 // The main purpose is that we are able to track segment and offset for
-// corruption reporting.
+// corruption reporting.  We have to be careful not to increment curr too
+// early.
 type segmentBufReader struct {
 	buf  *bufio.Reader
 	segs []*Segment
-	cur  int
-	off  int
-	more bool
+	cur  int // Index into segs.
+	off  int // Offset of read data into current segment.
 }
 
 func newSegmentBufReader(segs ...*Segment) *segmentBufReader {
 	return &segmentBufReader{
-		buf:  bufio.NewReaderSize(nil, 16*pageSize),
+		buf:  bufio.NewReaderSize(segs[0], 16*pageSize),
 		segs: segs,
-		cur:  -1,
 	}
 }
 
@@ -707,25 +706,38 @@ func (r *segmentBufReader) Close() (err error) {
 	return err
 }
 
+// Read implements io.Reader, expect that we occasionally return err=nil
+// when n != len(b), which violates the inteface.  Only use this with
+// io.ReadFull.
 func (r *segmentBufReader) Read(b []byte) (n int, err error) {
-	if !r.more {
-		if r.cur+1 >= len(r.segs) {
-			return 0, io.EOF
-		}
-		r.cur++
-		r.off = 0
-		r.more = true
-		r.buf.Reset(r.segs[r.cur])
-	}
 	n, err = r.buf.Read(b)
 	r.off += n
+
+	// If we succeeded, or hit a non-EOF, we can stop.
 	if err != io.EOF {
 		return n, err
 	}
-	// Just return what we read so far, but don't signal EOF.
-	// Only unset more so we don't invalidate the current segment and
-	// offset before the next read.
-	r.more = false
+
+	// We hit EOF; fake out zero padding at the end of short segments.
+	if r.off%pageSize != 0 {
+		i := 0
+		for ; n+i < len(b) && (r.off+i)%pageSize != 0; i++ {
+			b[n+i] = 0
+		}
+
+		// Return early, even if we didn't fill b.
+		r.off += i
+		return n + i, nil
+	}
+
+	// Now we actually need to move to next segment.
+	if r.cur+1 >= len(r.segs) {
+		return n, io.EOF
+	}
+
+	r.cur++
+	r.off = 0
+	r.buf.Reset(r.segs[r.cur])
 	return n, nil
 }
 
--- a/wal/wal.go
+++ b/wal/wal.go
@@ -706,9 +706,7 @@ func (r *segmentBufReader) Close() (err error) {
 	return err
 }
 
-// Read implements io.Reader, expect that we occasionally return err=nil
-// when n != len(b), which violates the inteface.  Only use this with
-// io.ReadFull.
+// Read implements io.Reader.
 func (r *segmentBufReader) Read(b []byte) (n int, err error) {
 	n, err = r.buf.Read(b)
 	r.off += n
--- a/wal/wal.go
+++ b/wal/wal.go
@@ -682,7 +682,8 @@ func NewSegmentsRangeReader(sr ...SegmentRange) (io.ReadCloser, error) {
 // segmentBufReader is a buffered reader that reads in multiples of pages.
 // The main purpose is that we are able to track segment and offset for
 // corruption reporting.  We have to be careful not to increment curr too
-// early.
+// early, as it is used by Reader.Err() to tell Repair which segment is corrupt.
+// As such we pad the end of non-page align segments with zeros.
 type segmentBufReader struct {
 	buf  *bufio.Reader
 	segs []*Segment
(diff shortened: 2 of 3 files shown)


tomwilkie:
I think this is an improvement, but I'm trying to think if there are still corrupt WALs which could fail repair - for instance, a WAL where a short segment causes the `io.ReadFull` to consume into the next segment.  Given the io.ReadFull is bounded by a page, the only case I can think of are segments containing records spanning pages - which shouldn't happen, but does.

To make this more robust (defense in depth), I think we should:
- Change repair to not depend on the the index of the WAL segment from the error, but rebuild the entire WAL (which would be slow)
- Get rid of repair all together, and when we find a corrupt WAL delete the entire thing and write out a new checkpoint (after all, we've just replayed the WAL into the head).
- Change the WAL code to not append to the existing segment, but always start a new segment.

WDYT?

wal/wal_test.go
```diff
@@ -547,6 +549,139 @@ func TestWAL_Repair(t *testing.T) {
 	}
 }
 
+func TestCorruptAndCarryOn(t *testing.T) {
+	dir, err := ioutil.TempDir("", "wal_repair")
+	testutil.Ok(t, err)
+	defer os.RemoveAll(dir)
+
+	logger := log.NewLogfmtLogger(os.Stderr)
```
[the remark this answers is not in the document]
tomwilkie: 👍 put the types in testutil

wal/wal_test.go
```diff
@@ -547,6 +549,139 @@ func TestWAL_Repair(t *testing.T) {
...
+
+	var records int
+
+	for i := 0; i < 10; i++ {
+		t.Log("Interation", i, "records", records)
```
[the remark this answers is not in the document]
tomwilkie: Loop was from earlier iterator, have removed it.

tomwilkie:
Thanks @krasi-georgiev!

wal/wal.go
```diff
@@ -707,25 +706,38 @@ func (r *segmentBufReader) Close() (err error) {
 	return err
 }
 
+// Read implements io.Reader, expect that we occasionally return err=nil
+// when n != len(b), which violates the inteface.  Only use this with
+// io.ReadFull.
```
gouthamve: Can you explain why we do that?

gouthamve: Wait, it infact doesn't violate the interface. https://golang.org/pkg/io/#Reader It's okay to return early.

cstyan:
> a WAL where a short segment causes the `io.ReadFull` to consume into the next segment. Given the io.ReadFull is bounded by a page, the only case I can think of are segments containing records spanning pages - which shouldn't happen, but does.

I don't think this is actually true. I suspect the corruptions we were seeing (both torn pages and failed crc checks) were because we were reading data at incorrect offsets in the file because the previous corrupt segment hadn't been padded with 0's.

testutil/logging.go
```diff
@@ -0,0 +1,20 @@
...
+	"testing"
+
+	"github.com/go-kit/kit/log"
+)
+
+type logger struct {
+	t *testing.T
+}
+
+func NewLogger(t *testing.T) log.Logger {
```
krasi-georgiev: go-lint : exported func requires a comment.

wal/wal.go
```diff
@@ -681,20 +681,19 @@ func NewSegmentsRangeReader(sr ...SegmentRange) (io.ReadCloser, error) {
 
 // segmentBufReader is a buffered reader that reads in multiples of pages.
 // The main purpose is that we are able to track segment and offset for
-// corruption reporting.
+// corruption reporting.  We have to be careful not to increment curr too
+// early.
```
krasi-georgiev: What do you mean by `too early.`?

wal/wal_test.go
```diff
@@ -547,6 +548,121 @@ func TestWAL_Repair(t *testing.T) {
...
+	dir, err := ioutil.TempDir("", "wal_repair")
+	testutil.Ok(t, err)
+	defer os.RemoveAll(dir)
+
+	logger := testutil.NewLogger(t)
+
+	// Produce a WAL with a two segments of 3 pages with 3 records each,
+	// so when we truncate the file we guaranteed to split a record.
+	{
+		w, err := NewSize(logger, nil, dir, pageSize*3)
```
krasi-georgiev: can set this as a var and use it few places
`expWalSize := int64(pageSize * 3)`

wal/wal_test.go
```diff
@@ -547,6 +548,121 @@ func TestWAL_Repair(t *testing.T) {
...
+	// Truncate the first file, splitting a record.
+	{
+		f, err := os.OpenFile(filepath.Join(dir, fmt.Sprintf("%08d", 0)), os.O_RDWR, 0666)
+		testutil.Ok(t, err)
+
+		fi, err := f.Stat()
+		testutil.Ok(t, err)
+		testutil.Equals(t, int64(3*pageSize), fi.Size())
+
+		err = f.Truncate((3 * pageSize) / 2)
```
krasi-georgiev: could be `err = f.Truncate(fi.Size() / 2)`

wal/wal_test.go
```diff
@@ -547,6 +548,121 @@ func TestWAL_Repair(t *testing.T) {
...
 
+func TestCorruptAndCarryOn(t *testing.T) {
+	dir, err := ioutil.TempDir("", "wal_repair")
+	testutil.Ok(t, err)
+	defer os.RemoveAll(dir)
+
+	logger := testutil.NewLogger(t)
+
+	// Produce a WAL with a two segments of 3 pages with 3 records each,
+	// so when we truncate the file we guaranteed to split a record.
```
krasi-georgiev:
```suggestion
	// so when we truncate the file we guaranteed to split the middle record.
```

wal/wal.go
tomwilkie: Yeah fair enough, don't know what I was thinking.

wal/wal_test.go
```diff
@@ -547,6 +548,121 @@ func TestWAL_Repair(t *testing.T) {
...
+	{
+		w, err := NewSize(logger, nil, dir, pageSize*3)
+		testutil.Ok(t, err)
+
+		for i := 0; i < 18; i++ {
+			buf := make([]byte, (pageSize/3)-recordHeaderSize)
```
krasi-georgiev: can set this as a var and use it few places
`expRecordSize := pageSize/3 - recordHeaderSize`

wal/wal_test.go
cstyan: we're guaranteed

tomwilkie:
> I don't think this is actually true. I suspect the corruptions we were seeing (both torn pages and failed crc checks) were because we were reading data at incorrect offsets in the file because the previous corrupt segment hadn't been padded with 0's.

We definitely have a segment with records spanning pages (and not just from concatenating segments, which isn't something the LiveReader does).  But I have no idea how we got into this state, and its kinda unrelated.

wal/wal_test.go
```diff
@@ -547,6 +549,144 @@ func TestWAL_Repair(t *testing.T) {
...
+				testutil.Ok(t, err)
+
+				fi, err := f.Stat()
+				testutil.Ok(t, err)
+
+				t.Log("segment", segment.index, "size", fi.Size())
+				testutil.Equals(t, int64(3*pageSize), fi.Size())
+
+				err = f.Close()
+				testutil.Ok(t, err)
```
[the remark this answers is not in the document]
tomwilkie: I much prefer the regularity in the tests of checking errors separately to function calls.  Makes it easier to follow and read.

wal/wal.go
```diff
@@ -681,20 +681,19 @@ func NewSegmentsRangeReader(sr ...SegmentRange) (io.ReadCloser, error) {
...
-	off  int
-	more bool
+	cur  int // Index into segs.
+	off  int // Offset of read data into current segment.
 }
 
 func newSegmentBufReader(segs ...*Segment) *segmentBufReader {
 	return &segmentBufReader{
-		buf:  bufio.NewReaderSize(nil, 16*pageSize),
+		buf:  bufio.NewReaderSize(segs[0], 16*pageSize),
```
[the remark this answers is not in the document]
tomwilkie: `Segment` does implement io.Reader, through the embedded `os.File`.

wal/wal_test.go
```diff
@@ -547,6 +548,121 @@ func TestWAL_Repair(t *testing.T) {
...
+
+			t.Log("segment", segment.index, "size", fi.Size())
+			testutil.Equals(t, int64(3*pageSize), fi.Size())
+
+			err = f.Close()
+			testutil.Ok(t, err)
+		}
+	}
+
+	// Truncate the first file, splitting a record.
```
cstyan: can we specify that we're truncating at the offset that's in the middle of the 2nd page, which would leave us with 4 valid records (1 full page and 1 rec in the next page)

wal/wal.go
```diff
@@ -707,25 +706,36 @@ func (r *segmentBufReader) Close() (err error) {
...
-		r.cur++
-		r.off = 0
-		r.more = true
-		r.buf.Reset(r.segs[r.cur])
-	}
 	n, err = r.buf.Read(b)
 	r.off += n
+
+	// If we succeeded, or hit a non-EOF, we can stop.
 	if err != io.EOF {
```
krasi-georgiev: Maybe `if err==nil || err != io.EOF {`  ?
I was a it confused at the begining so maybe be a bit more mor e explicit to avoid confusion how come we succeeded here.

wal/wal.go
```diff
@@ -707,25 +706,38 @@ func (r *segmentBufReader) Close() (err error) {
...
 		return n, err
 	}
-	// Just return what we read so far, but don't signal EOF.
-	// Only unset more so we don't invalidate the current segment and
-	// offset before the next read.
-	r.more = false
+
+	// We hit EOF; fake out zero padding at the end of short segments.
```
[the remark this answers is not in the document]
tomwilkie: This is the whole point of the PR.   From the description:

> The test fails against master, as the underlying segmentReader has already moved on to the next segment.

> make the segmentReader pad short segments with zeros, and only advance to next segment on the call to `Read` _after_ completely consuming this one

> previously it would increment `curr` on the call to `Read` which failed, causing the returns `CorruptionErr` to reference the wrong segment.

wal/wal.go
```diff
@@ -707,25 +706,38 @@ func (r *segmentBufReader) Close() (err error) {
...
+		for ; n+i < len(b) && (r.off+i)%pageSize != 0; i++ {
+			b[n+i] = 0
+		}
+
+		// Return early, even if we didn't fill b.
+		r.off += i
+		return n + i, nil
+	}
+
+	// Now we actually need to move to next segment.
```
cstyan: can we move this comment to line 737 and have a comment here "end of the existing segments"

testutil/logging.go
tomwilkie: We don't seem to require go-lint in this repo, there are lots of existing warning in the codebase.  If we do, can I suggest we add it to CI?  Either was I'll add a comment here.

wal/wal.go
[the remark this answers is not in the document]
krasi-georgiev: weird, I looked in that struct as well but can't see any Read method.

wal/wal.go
tomwilkie: > previously it would increment `curr` on the call to `Read` which failed, causing the returns `CorruptionErr` to reference the wrong segment.

I'll add this to the comment.

wal/wal.go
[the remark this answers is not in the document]
krasi-georgiev: weird. it does show in the golang docs, but not when I use the code navigation in vscode.
https://golang.org/pkg/os/#File.Read

wal/wal.go
tomwilkie: @cstyan mentioned this too, will do.

wal/wal.go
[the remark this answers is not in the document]
krasi-georgiev: Yeah sorry I meant to add some more details in the comment.

wal/wal_test.go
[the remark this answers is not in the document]
krasi-georgiev: most places use the short version, but I see your point. Not a blocker.

wal/wal.go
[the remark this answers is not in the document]
tomwilkie: Done!

tomwilkie:
Thanks for the feedback @cstyan @krasi-georgiev !  I think I've covered it all, left me know if I've missed anything.

krasi-georgiev:
> To make this more robust (defense in depth), I think we should:

> * Change repair to not depend on the the index of the WAL segment from the error, but rebuild the entire WAL (which would be slow)

> * Get rid of repair all together, and when we find a corrupt WAL delete the entire thing and write out a new checkpoint (after all, we've just replayed the WAL into the head).

this one sounds as the best solution, but maybe do it in a separate PR if it happens that the current implementation causes other problems.

> * Change the WAL code to not append to the existing segment, but always start a new segment.

tomwilkie:
> this one sounds as the best solution, but maybe do it in a separate PR if it happens that the current implementation causes other problems.

For sure!  I was thinking aloud there.  I'll follow up with a ticket for this.

wal/wal_test.go
```diff
@@ -547,6 +548,125 @@ func TestWAL_Repair(t *testing.T) {
 	}
 }
 
+func TestCorruptAndCarryOn(t *testing.T) {
```
krasi-georgiev: The name is quite clear, but if you don't mind please add a more extended comment the behaviour that this test is verifying. It will probably be better when looking at it in few months time. For example if some code changes cause it to fail.

testutil/logging.go
```diff
@@ -0,0 +1,22 @@
+package testutil
```
codesome: Nits. Copyright notice for the new file.

testutil/logging.go
```diff
@@ -0,0 +1,35 @@
+// Copyright 2013 The Prometheus Authors
```
codesome: Was expecting something like this to happen :)
```suggestion
// Copyright 2019 The Prometheus Authors
```

testutil/logging.go
krasi-georgiev: I wonder why the license test didn't catch this?

wal/wal_test.go
krasi-georgiev: I meant something like

```
TestCorruptAndCarryOn ensures that an error during reading the segment .... and moves to reading the next segment only after .....
```

PR #539, opened 2019-03-04
[bug] didn't update 'last' variable in chunks.MergeOverlappingChunks()

naivewong:
1. Update 'last' variable after appending a chunk.
2. I add test case for it. The blocks of labels ("before", "fix") are overlapping together and they can pass in the original codes. The blocks of labels ("after", "fix") are overlapping only for the last two blocks and cannot pass in the original codes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 ## master / unreleased
+  - [BUGFIX] Update `last` after appending a non-overlapping chunk in `chunks.MergeOverlappingChunks`. [#539](https://github.com/prometheus/tsdb/pull/539)
 
 ## 0.6.0
   - [CHANGE] `AllowOverlappingBlock` is now `AllowOverlappingBlocks`.
-  - [BUGFIX] Update `last` after appending a non-overlapping chunk in `chunks.MergeOverlappingChunks`. [#539](https://github.com/prometheus/tsdb/pull/539)
 
 ## 0.5.0
  - [FEATURE] Time-ovelapping blocks are now allowed. [#370](https://github.com/prometheus/tsdb/pull/370)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.6.0
   - [CHANGE] `AllowOverlappingBlock` is now `AllowOverlappingBlocks`.
+  - [BUGFIX] Update `last` after appending a non-overlapping chunk in `chunks.MergeOverlappingChunks`. [#539](https://github.com/prometheus/tsdb/pull/539)
 
 ## 0.5.0
  - [FEATURE] Time-ovelapping blocks are now allowed. [#370](https://github.com/prometheus/tsdb/pull/370)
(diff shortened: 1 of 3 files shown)


krasi-georgiev:
@naivewong 	failing tests.

naivewong:
@krasi-georgiev Yes I saw that, the failing test is as follows
```
{
	// No special deduplication expected.
	title: "Populate from two blocks containing duplicated chunk.",
	inputSeriesSamples: [][]seriesSamples{
		{
			{
				lset:   map[string]string{"a": "b"},
				chunks: [][]sample{{{t: 1}, {t: 2}}, {{t: 10}, {t: 20}}},
			},
		},
		{
			{
				lset:   map[string]string{"a": "b"},
				chunks: [][]sample{{{t: 10}, {t: 20}}},
			},
		},
	},
	expSeriesSamples: []seriesSamples{
		{
			lset:   map[string]string{"a": "b"},
			chunks: [][]sample{{{t: 1}, {t: 2}}, {{t: 10}, {t: 20}}, {{t: 10}, {t: 20}}},
		},
	},
},
```
Actually this reveals the bug, there are three chunks in this case, the last two chunks are overlapping.
In the original codes, since the 'last' variable isn't update and remained 0, the overlapping of the last two chunks(index 1 and 2) cannot be detected. Then you will see **"No deduplication"** in this case.

But If you change the second input samples to `chunks: [][]sample{{{t: 1}, {t: 2}}},`, then it will become the first two chunks overlapping. And in this case, the original codes can detect and merge the first two chunks(index 0 and 1) because the `last=0`. And the original code will output `chunks: [][]sample{{{t: 1}, {t: 2}}, {{t: 10}, {t: 20}}}` which actually do the **deduplication**.

So I think the deduplication should happen here.

codesome:
@naivewong Thanks for this! Can you update the failing test mentioning that the chunk should be merged/deduped? This should go in 2.8.

compact_test.go
```diff
@@ -653,7 +653,8 @@ func TestCompaction_populateBlock(t *testing.T) {
 			expErr:         errors.New("found chunk with minTime: 10 maxTime: 20 outside of compacted minTime: 0 maxTime: 10"),
 		},
 		{
-			// No special deduplication expected.
+			// Deduplication expected.
+			// Introduced by https://github.com/prometheus/tsdb/pull/539.
```
codesome: Introduced by 370 and 539, as merging chunks happened in 370 and 539 is a bug fix in that.

krasi-georgiev:
should also add an entry in the changelog file.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 0.6.0
   - [CHANGE] `AllowOverlappingBlock` is now `AllowOverlappingBlocks`.
+  - [BUGFIX] Update `last` after appending a non-overlapping chunk in `chunks.MergeOverlappingChunks`. [#539](https://github.com/prometheus/tsdb/pull/539)
```
codesome: Can you move this under `## master / unreleased`? And I will merge it. Thanks for your work!

codesome: I did it myself, sorry for hijacking.

PR #540, opened 2019-03-05
Remove FromData function

zhulongcheng:
The role of `FromData`   function has been replaced by `pool.Get`.

Related PR: #118   
https://github.com/prometheus/tsdb/pull/118/files#diff-5047d434b918cce51937ad6204d8afb0L358

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## master / unreleased
+ - [REMOVED] `FromData` is considered unused so was removed.
 
 ## 0.6.1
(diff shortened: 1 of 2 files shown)


krasi-georgiev:
should also add an entry in the CHANGELOG file.

krasi-georgiev:
Thanks!

sylr:
Thanos uses this function https://github.com/improbable-eng/thanos/blob/master/pkg/compact/downsample/aggr.go#L89 :(

bwplotka:
So this is fine as long as we can use replacement, right?  (: The question, can we? I will look at it soon.

sylr:
this has been reverted by #600 so it's fine.

bwplotka:
xD why?

krasi-georgiev:
because cortex also uses it.

bwplotka:
Cool, thanks for clarification <3

Issue #516, opened 2019-02-06
TSDB CLI tool doesn't work on live Prometheus


glightfoot:
It seems like the tsdb cli tool doesn't work on a live instance of prometheus. [This blog post](https://www.robustperception.io/using-tsdb-analyze-to-investigate-churn-and-cardinality) makes it seem like the intention is to be able to have it be able to work on a live instance, but it currently errors out with 

> tsdb ls /volumes/monitoring-prometheus
> lock DB directory: resource temporarily unavailable

I've added a flag to enable opening the tsdb with no locking, and this seems to work fine in my testing.

glightfoot:
https://github.com/prometheus/tsdb/pull/517/files

PR #541, opened 2019-03-05
ReadOnly option

krasi-georgiev:
just continuing the work from https://github.com/prometheus/tsdb/pull/517

fixes https://github.com/prometheus/tsdb/issues/516
 

glightfoot:
@krasi-georgiev - is there anything I can do to help with this?

krasi-georgiev:
@glightfoot 	no thanks , I will finish this probably tomorrow.

krasi-georgiev:
ready for review
ping @gouthamve , @codesome

krasi-georgiev:
ping @gouthamve , @codesome

cmd/tsdb/main.go
```diff
@@ -33,12 +33,23 @@ import (
...
 	"gopkg.in/alecthomas/kingpin.v2"
 )
 
+func openReadOnlyTSDB(dir string, l log.Logger, r prometheus.Registerer) (db *tsdb.DB, err error) {
+	// Dereference the DefaultOptions so that it makes a copy and
+	// applies changes to the copy and not globally.
+	opts := *tsdb.DefaultOptions
+	opts.NoLockfile = true
+	opts.ReadOnly = true
+	db, err = tsdb.Open(dir, l, r, &opts)
```
bwplotka: return `tsdb.Open(dir, l, r, &opts)` is not enough?

cmd/tsdb/main.go
```diff
@@ -62,9 +73,6 @@ func main() {
 		dumpMaxTime          = dumpCmd.Flag("max-time", "maximum timestamp to dump").Default(strconv.FormatInt(math.MaxInt64, 10)).Int64()
 	)
 
-	safeDBOptions := *tsdb.DefaultOptions
```
bwplotka: I think I would prefer `readOnlyOptions` struct and pass it everywhere.

cmd/tsdb/main.go
```diff
@@ -33,12 +33,23 @@ import (
...
 	"github.com/go-kit/kit/log"
 	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/tsdb"
 	"github.com/prometheus/tsdb/chunks"
 	"github.com/prometheus/tsdb/labels"
 	"gopkg.in/alecthomas/kingpin.v2"
 )
 
+func openReadOnlyTSDB(dir string, l log.Logger, r prometheus.Registerer) (db *tsdb.DB, err error) {
```
bwplotka: Not sure why those logger and registerer opts if you never fill them (:  I think it proves that just `readOnlyOpts` struct would be enough instead of this function

db.go
```diff
@@ -79,6 +80,10 @@ type Options struct {
 	// Overlapping blocks are allowed if AllowOverlappingBlocks is true.
 	// This in-turn enables vertical compaction and vertical query merge.
 	AllowOverlappingBlocks bool
+
+	// ReadOnly disables all mutating actions.
```
bwplotka: "Disables", means it will return error when acessing Appender methods etc?

db.go
```diff
@@ -79,6 +80,10 @@ type Options struct {
...
+	// Useful for opening the same db more than once.
+	ReadOnly bool
```
bwplotka: Have we considered totally different constructor instead returning limited TSDB interface ?

db.go
```diff
@@ -588,8 +598,10 @@ func (db *DB) reload() (err error) {
 		}
 	}
 
-	if err := db.deleteBlocks(deletable); err != nil {
-		return err
+	if !db.opts.ReadOnly {
```
bwplotka: All those `ifs` in this file. Do we really need to invoke `reload` and exactly the same `Open` functions in readOnly TSDB mode?

What readOnly mode should give? I think only tiny shim for Querying, right? Can we just add separate layer that does just that? It would massively reduce extra complexity in critical code here. What do you think? (:

krasi-georgiev:
superseded by https://github.com/prometheus/tsdb/pull/588

PR #542, opened 2019-03-05
Add Head.compactable method

zhulongcheng:
Extract whether head has a compactable range  into a method (`Head.compactable` ).

### DIFF
--- a/db.go
+++ b/db.go
@@ -382,7 +382,7 @@ func (a dbAppender) Commit() error {
 
 	// We could just run this check every few minutes practically. But for benchmarks
 	// and high frequency use cases this is the safer way.
-	if a.db.head.MaxTime()-a.db.head.MinTime() > a.db.head.chunkRange/2*3 {
+	if a.db.head.compactable() {
 		select {
 		case a.db.compactc <- struct{}{}:
 		default:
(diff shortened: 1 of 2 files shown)


head.go
```diff
@@ -1020,6 +1020,13 @@ func (h *Head) MaxTime() int64 {
 	return atomic.LoadInt64(&h.maxTime)
 }
 
+// compactable returns whether the head has a compactable range.
+func (h *Head) compactable() bool {
+	// The head has a compactable range if 1.5 level 0 ranges are between the oldest
+	// and newest timestamp. The 0.5 acts as a buffer of the appendable window.
```
krasi-georgiev: you can combine the func comment with this one to make it only a single comment describing the func

zhulongcheng: ok

head.go
```diff
@@ -1020,6 +1020,13 @@ func (h *Head) MaxTime() int64 {
 	return atomic.LoadInt64(&h.maxTime)
 }
 
+// compactable returns whether the head has a compactable range.
+// The head has a compactable range if 1.5 level 0 ranges are between the oldest
+// and newest timestamp.
```
krasi-georgiev: how about something like 
```
// compactable returns whether the head has a compactable range.
// The head has a compactable range when the head time range is 1.5 times the chunk range.
// The 0.5 acts as a buffer of the appendable window.
```

db.go
```diff
@@ -409,13 +409,11 @@ func (db *DB) compact() (err error) {
 			return nil
 		default:
 		}
-		// The head has a compactable range if 1.5 level 0 ranges are between the oldest
-		// and newest timestamp. The 0.5 acts as a buffer of the appendable window.
-		if db.head.MaxTime()-db.head.MinTime() <= db.opts.BlockRanges[0]/2*3 {
```
krasi-georgiev: wondering why it was `<=` here and not just `<`

@gouthamve 	do you think this will cause issues?

head.go
zhulongcheng: Great 👍 

Thanks.

krasi-georgiev:
LGTM

ping @codesome

zhulongcheng:
ping

PR #544, opened 2019-03-05
Verify chunk format version

zhulongcheng:
Verify chunk format version.

### DIFF
--- a/block_test.go
+++ b/block_test.go
@@ -140,12 +140,8 @@ func TestOpenBlock_ChunkCorrupted(t *testing.T) {
 			test.corrFunc(f)
 			testutil.Ok(t, f.Close())
 
-			b, err := OpenBlock(nil, blockDir, nil)
-			if err != nil {
-				testutil.Equals(t, test.expErr.Error(), err.Error())
-			} else {
-				testutil.Ok(t, b.Close())
-			}
+			_, err = OpenBlock(nil, blockDir, nil)
+			testutil.Equals(t, test.expErr.Error(), err.Error())
 		})
 	}
 }
--- a/block_test.go
+++ b/block_test.go
@@ -80,7 +80,7 @@ func TestCreateBlock(t *testing.T) {
 	testutil.Ok(t, err)
 }
 
-func TestOpenBlock_ChunkCorrupted(t *testing.T) {
+func TestCorruptedChunk(t *testing.T) {
 	for name, test := range map[string]struct {
 		corrFunc func(f *os.File) // Func that applies the corruption.
 		expErr   error
@@ -94,7 +94,8 @@ func TestOpenBlock_ChunkCorrupted(t *testing.T) {
 		},
 		"invalid magic number": {
 			func(f *os.File) {
-				_, err := f.Seek(0, 0)
+				magicChunksOffset := int64(0)
+				_, err := f.Seek(magicChunksOffset, 0)
 				testutil.Ok(t, err)
 
 				// Set invalid magic number.
@@ -108,7 +109,8 @@ func TestOpenBlock_ChunkCorrupted(t *testing.T) {
 		},
 		"invalid chunk format version": {
 			func(f *os.File) {
-				_, err := f.Seek(4, 0)
+				chunksFormatVersionOffset := int64(4)
+				_, err := f.Seek(chunksFormatVersionOffset, 0)
 				testutil.Ok(t, err)
 
 				// Set invalid chunk format version.
--- a/block_test.go
+++ b/block_test.go
@@ -24,6 +24,7 @@ import (
 	"testing"
 
 	"github.com/go-kit/kit/log"
+	"github.com/prometheus/tsdb/chunks"
 	"github.com/prometheus/tsdb/testutil"
 	"github.com/prometheus/tsdb/tsdbutil"
 )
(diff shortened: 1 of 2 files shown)


krasi-georgiev:
it seems that some parts of this code like the `NewReader` func are not used anywhere so wondering whether we should remove it completely.

krasi-georgiev:
just double checked with @codesome  and @bwplotka 	and you can remove `NewReader` it seems to be some left over code.

zhulongcheng:
A new commit to remove `chunks.NewReader`.
If it is not ready to be removed, I will reset the commit.

krasi-georgiev:
would you mind opening a separate PR for the NewReader removal to keep the commit history nicer.

chunks/chunks_test.go
```diff
@@ -26,3 +28,34 @@ func TestReaderWithInvalidBuffer(t *testing.T) {
 	_, err := r.Chunk(0)
 	testutil.NotOk(t, err)
 }
+
+func TestNewReaderWithInvalidChunkHeader(t *testing.T) {
```
krasi-georgiev: Where it makes sense I am vouching for [behaviour tests](https://www.youtube.com/watch?v=EZ05e7EMOLM)(using only public API interfaces) and here this would mean to test that reading corrupted chunks from the disk return the correct errors. So it looks like we should use `OpenBlock`.

you can create a block (`createBlock`) , than manually corrupt a chunk and try to open the block. You can have a look at `TestWAL_Repair` for ideas how to corrupt a chunk.

zhulongcheng: Great suggestions.

Thank you!

zhulongcheng:
> would you mind opening a separate PR for the `NewReader` removal to keep the commit history nicer.

https://github.com/prometheus/tsdb/pull/547

block_test.go
```diff
@@ -60,6 +62,73 @@ func TestSetCompactionFailed(t *testing.T) {
...
+		"invalid header size": {
+			func(f *os.File) {
+				err := f.Truncate(1)
+				testutil.Ok(t, err)
+			},
+			errors.New("invalid chunk header in segment 0: invalid size"),
+		},
+		"invalid magic number": {
+			func(f *os.File) {
+				magicChunksSize := 4
```
krasi-georgiev: I don't see any problem on exposing `chunks.MagicChunksSize` so that we don't need to duplicate it for the test. The same for the other bits below.

block_test.go
```diff
@@ -60,6 +62,73 @@ func TestSetCompactionFailed(t *testing.T) {
...
+				magicChunksOffset := int64(0)
+				_, err := f.Seek(magicChunksOffset, 0)
+				testutil.Ok(t, err)
+
+				// Set invalid magic number 0x00000000
```
krasi-georgiev:
```suggestion
				// Set invalid magic number.
```

block_test.go
```diff
@@ -60,6 +62,73 @@ func TestSetCompactionFailed(t *testing.T) {
...
+			errors.New("invalid magic number 0"),
+		},
+		"invalid chunk format version": {
+			func(f *os.File) {
+				chunksFormatVersionSize := 1
+				chunksFormatVersionOffset := int64(4)
+				_, err := f.Seek(chunksFormatVersionOffset, 0)
+				testutil.Ok(t, err)
+
+				// Set invalid chunk format version 0
```
krasi-georgiev:
```suggestion
				// Set invalid chunk format version.
```

block_test.go
```diff
@@ -60,6 +62,73 @@ func TestSetCompactionFailed(t *testing.T) {
...
+			files, err := sequenceFiles(chunkDir(blockDir))
+			testutil.Ok(t, err)
+			testutil.Assert(t, len(files) > 0, "No chunk created.")
+
+			f, err := os.OpenFile(files[0], os.O_RDWR, 0666)
+			testutil.Ok(t, err)
+
+			// Apply corruption function.
+			test.corrFunc(f)
+
```
krasi-georgiev:
```suggestion
	testutil.Ok(t, f.Close())
```

Without this it will probably fail on Windows when you do the `os.RemoveAll(tmpdir)`.

block_test.go
```diff
@@ -60,6 +62,73 @@ func TestSetCompactionFailed(t *testing.T) {
...
+				testutil.Ok(t, err)
+				testutil.Equals(t, chunksFormatVersionSize, n)
+			},
+			errors.New("invalid chunk format version 0"),
+		},
+	} {
+		t.Run(name, func(t *testing.T) {
+			tmpdir, err := ioutil.TempDir("", "test_open_block_chunk_corrupted")
+			testutil.Ok(t, err)
+			defer os.RemoveAll(tmpdir)
```
krasi-georgiev:
```suggestion
			defer func() {
				testutil.Ok(t, os.RemoveAll(tmpdir))
			}()
```

block_test.go
```diff
@@ -60,6 +62,73 @@ func TestSetCompactionFailed(t *testing.T) {
...
+			testutil.Ok(t, err)
+			testutil.Assert(t, len(files) > 0, "No chunk created.")
+
+			f, err := os.OpenFile(files[0], os.O_RDWR, 0666)
+			testutil.Ok(t, err)
+
+			// Apply corruption function.
+			test.corrFunc(f)
+
+			_, err = OpenBlock(nil, blockDir, nil)
```
krasi-georgiev: to ensure that `os.RemoveAll(tmpdir)` works for windows in case of no error after the corruption function.
```suggestion
			b, err = OpenBlock(nil, blockDir, nil)
			if err == nil {
				testutil.Ok(t, b.Close())
			}
```

zhulongcheng: In this test, `OpenBlock` should always return an error.

Would you explain why no error after the corruption function？

block_test.go
zhulongcheng: OK,  Thanks again.

block_test.go
krasi-georgiev: yes that ine is not important.

krasi-georgiev:
failing windows tests.

probably missing Close() somewhere

zhulongcheng:
> failing windows tests.
> 
> probably missing Close() somewhere

Yes.  `NewDirReader` should close chunk files before return an error.

https://github.com/prometheus/tsdb/pull/544/commits/6282482df598a485fd26a4b9e959ebd274cc8625

chunks/chunks.go
```diff
@@ -32,7 +32,13 @@ import (
 
 const (
 	// MagicChunks is 4 bytes at the head of a series file.
-	MagicChunks = 0x85BD40DD
+	MagicChunks     = 0x85BD40DD
+	MagicChunksSize = 4
```
krasi-georgiev: comment for exported var

block_test.go
```diff
@@ -60,6 +63,72 @@ func TestSetCompactionFailed(t *testing.T) {
...
+				binary.BigEndian.PutUint32(b[:chunks.MagicChunksSize], 0x00000000)
+				n, err := f.Write(b)
+				testutil.Ok(t, err)
+				testutil.Equals(t, chunks.MagicChunksSize, n)
+			},
+			errors.New("invalid magic number 0"),
+		},
+		"invalid chunk format version": {
+			func(f *os.File) {
+				_, err := f.Seek(4, 0)
```
krasi-georgiev: I think it was ok like before with the var name for the seek position, It was easier to understand why seeking at that position, but I also don't mind if you prefer to keep it like this as well.

block_test.go
```diff
@@ -60,6 +63,72 @@ func TestSetCompactionFailed(t *testing.T) {
 	testutil.Ok(t, b.Close())
 }
 
+func TestOpenBlock_ChunkCorrupted(t *testing.T) {
```
krasi-georgiev:
```suggestion
func TestCorruptedChunk(t *testing.T) {
```

chunks/chunks.go
```diff
@@ -425,7 +425,17 @@ func NewDirReader(dir string, pool chunkenc.Pool) (*Reader, error) {
 		cs = append(cs, f)
 		bs = append(bs, realByteSlice(f.Bytes()))
 	}
-	return newReader(bs, cs, pool)
+
+	reader, err := newReader(bs, cs, pool)
+	if err != nil {
+		for i, c := range cs {
```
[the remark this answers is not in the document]
zhulongcheng: Yes.  import cycle if no separate error package.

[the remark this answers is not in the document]
krasi-georgiev: I just merged the other PR so can continue with this one.

zhulongcheng:
Updated. @krasi-georgiev

krasi-georgiev:
Thanks!
@zhulongcheng 	 good job :+1:

PR #546, opened 2019-03-10
Rename metrics from prometheus_tsdb to tsdb_

roidelapluie:
The prefix prometheus_ will only be added when used within Prometheus.

Signed-off-by: Julien Pivotto <roidelapluie@inuits.eu>

### DIFF
--- a/compact.go
+++ b/compact.go
@@ -95,44 +95,44 @@ func newCompactorMetrics(r prometheus.Registerer) *compactorMetrics {
 	m := &compactorMetrics{}
 
 	m.ran = prometheus.NewCounter(prometheus.CounterOpts{
-		Name: "prometheus_tsdb_compactions_total",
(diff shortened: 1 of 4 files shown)


roidelapluie:
see https://github.com/prometheus/prometheus/pull/5328

brian-brazil:
These are prometheus_tsdb metrics, not prometheus tsdb metrics - so this is fine. Just tsdb would be ambiguous as there's many many tsdbs out there.

roidelapluie:
ok. Is is still worth it to refactor to use WrapRegistererWithPrefix?

roidelapluie:
What I find really confusing is that there are prometheus_tsdb_ metrics coming from TSDB and other ones from prometheus, which sounds like namespace violation.

brian-brazil:
Yeah, that clash is a bit messy.

PR #551, opened 2019-03-17
Update Appender.AddFast method comment

zhulongcheng:
The method name in the comment for `AddFast` should be "AddFast".


krasi-georgiev:
Thanks

PR #556, opened 2019-03-18
Fix a typo: overlaping -> overlapping

zhulongcheng:
overlaping -> overlapping


krasi-georgiev:
I will add this fix as part of another PR as I will update that comment a bit more.

Issue #554, opened 2019-03-18
selectOverlappingDirs bug that misses some directories


krasi-georgiev:
`selectOverlappingDirs` doesn't work as expected for the followinf input
Here suppose we have 4 blocks, 1: [0, 10], 2: [9, 20], 3: [30, 40], 4: [39, 50]
This function will return **{1, 2, 4}**.
expect is **{1, 2, 3, 4}** ?

_Originally posted by @naivewong in https://github.com/prometheus/tsdb/pull/550#issuecomment-473490971_

krasi-georgiev:
I am working on a fix and adding the test case.

PR #559, opened 2019-03-18
fix for missing overlapping blocks from selectOverlappingDirs

krasi-georgiev:
fixes https://github.com/prometheus/tsdb/issues/554
Signed-off-by: Krasi Georgiev <kgeorgie@redhat.com>

PR #560, opened 2019-03-19
A simpler solution to #559 and fix comments in compact_test.go

naivewong:
fixes https://github.com/prometheus/tsdb/issues/554

alternative approach to https://github.com/prometheus/tsdb/issues/554


krasi-georgiev:
@naivewong nice! Didn't think about that approach.

Although it is more efficient I find it a bit more difficult to follow.

If you don't mind I would wait for one more opinion and will decide which one to merge

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## master / unreleased
  - [REMOVED] `chunks.NewReader` is removed as it wasn't used anywhere.
  - [REMOVED] `FromData` is considered unused so was removed.
+ - [BUGFIX] Vertical compaction was missing overlaping blocks.
```
codesome: I think it would be apt to call it a bugfix in `selectOverlappingDirs` (applies to #559 as well if that is to be merged) because vertical compaction didn't actually miss overlapping blocks.

To explain with the same example from here https://github.com/prometheus/tsdb/pull/550#issuecomment-473490971 :
* `db.compact()` runs in a loop till `db.compactor.Plan()` returns some dirs.
* In 1st loop trip {1,2,4} would be compacted. In 2nd loop trip {1,2,4} and {3} would be picked for compaction.
* Hence eventually all overlapping blocks are compacted at the end of db.compact().

codesome:
Sorry for catching this late, both this PR and #559 are slightly off from what is desired.

Taking example of this comment again https://github.com/prometheus/tsdb/pull/550#issuecomment-473490971, this is how it should work :
Only overlapping chain of blocks should be merged, so {1,2} should be picked from this set of blocks. And later {3,4} would get picked (`compact()` runs a loop). Note that {1,2} and {3,4} don't overlap. So the result should be {1,2} and {3,4}, and not {1,2,3,4}. 

I think this can be done by just breaking the loop where we are setting `chained` to `false`.

With current fix (and even before this fix!), we can run into an issue of compacting all the blocks into a single block (If not clear how, I can give example). That's a serious bug that you found out @naivewong, thanks!

compact_test.go
```diff
@@ -346,10 +346,23 @@ func TestLeveledCompactor_plan(t *testing.T) {
 			},
 			expected: []string{"5", "6", "7", "8"},
 		},
+		//|--------------|
+		//        |----------------|
+		// 				     	|--------------|
+		//       					|----------------|
```
krasi-georgiev: this one and some other graphs are not aligned properly.

compact.go
```diff
@@ -266,20 +266,27 @@ func (c *LeveledCompactor) selectDirs(ds []dirMeta) []dirMeta {
...
 	for i, d := range ds[1:] {
 		if d.meta.MinTime < globalMaxt {
-			if len(overlappingDirs) == 0 { // When it is the first overlap, need to add the last one as well.
+			if !chained {
 				overlappingDirs = append(overlappingDirs, ds[i].dir)
+				chained = true
 			}
 			overlappingDirs = append(overlappingDirs, d.dir)
+		} else {
+			if chained {
```
codesome: We can get rid of `chained` variable by having `len(overlappingDirs) > 0` check here instead.

codesome:
@naivewong you have got some conflicts.

compact_test.go
```diff
@@ -312,44 +300,69 @@ func TestLeveledCompactor_plan(t *testing.T) {
 			},
 			expected: []string{"7", "8"},
 		},
-		// For overlapping blocks.
-		{
+		//|--------------|
+		//              |----------------|
+		//			         |--------------|
```
krasi-georgiev: not aligned properly.

compact_test.go
```diff
@@ -312,44 +300,69 @@ func TestLeveledCompactor_plan(t *testing.T) {
...
 				metaRange("1", 0, 20, nil),
 				metaRange("2", 19, 40, nil),
 				metaRange("3", 40, 60, nil),
 			},
 			expected: []string{"1", "2"},
 		},
-		{
+		//|--------------|
+		//               |--------------|
+		//			 |--------------|
```
krasi-georgiev: not aligned properly.

compact_test.go
```diff
@@ -312,44 +300,69 @@ func TestLeveledCompactor_plan(t *testing.T) {
...
 				metaRange("1", 0, 20, nil),
 				metaRange("2", 20, 40, nil),
 				metaRange("3", 30, 50, nil),
 			},
 			expected: []string{"2", "3"},
 		},
-		{
+		//|--------------|
+		//        |--------------|
+		//		 |--------------|
```
krasi-georgiev: not aligned properly.

compact_test.go
```diff
@@ -312,44 +300,69 @@ func TestLeveledCompactor_plan(t *testing.T) {
...
+				metaRange("2", 10, 30, nil),
+				metaRange("3", 20, 40, nil),
 			},
 			expected: []string{"1", "2", "3"},
 		},
-		{
+		//|--------------|
+		//              |-------------------------------|
+		//		 |--------------|
+		//			        |--------------|
```
krasi-georgiev: not aligned properly.

compact_test.go
naivewong: Sorry I don't quite understand...I already made some changes to the range values and the graphs to make them aligned in the previous commits. Can you give me an example?

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## master / unreleased
  - [REMOVED] `chunks.NewReader` is removed as it wasn't used anywhere.
  - [REMOVED] `FromData` is considered unused so was removed.
+ - [BUGFIX] Bugfix in selectOverlappingDirs. Only return truly overlapping blocks.
```
bwplotka: what `truly` actually means?

compact.go
```diff
@@ -280,6 +280,8 @@ func (c *LeveledCompactor) selectOverlappingDirs(ds []dirMeta) []string {
 				overlappingDirs = append(overlappingDirs, ds[i].dir)
 			}
 			overlappingDirs = append(overlappingDirs, d.dir)
+		}else if len(overlappingDirs) > 0 {
```
bwplotka: wrong formatting?
```suggestion
		} else if len(overlappingDirs) > 0 {
```

bwplotka: is that right? Why we return only first overlap? Is this intented?

compact_test.go
```diff
@@ -312,44 +300,69 @@ func TestLeveledCompactor_plan(t *testing.T) {
...
+				metaRange("8", 40, 60, nil),
 			},
 			expected: []string{"5", "6", "7", "8"},
 		},
+		//|--------------|
+		//        |--------------|
+		// 				  |--------------|
+		//       				  |--------------|
+		"Overlapping blocks 5": {
+			metas: []dirMeta{
```
bwplotka: Is there any use case as this https://github.com/prometheus/tsdb/issues/554 actually?

bwplotka:
Where are we actually fixing https://github.com/prometheus/tsdb/issues/554 ?

compact_test.go
```diff
@@ -312,44 +300,69 @@ func TestLeveledCompactor_plan(t *testing.T) {
...
+				metaRange("6", 19, 61, nil),
+				metaRange("7", 20, 40, nil),
+				metaRange("8", 40, 60, nil),
 			},
 			expected: []string{"5", "6", "7", "8"},
 		},
+		//|--------------|
+		//        |--------------|
+		// 				  |--------------|
+		//       				  |--------------|
```
krasi-georgiev: not aligned properly.

krasi-georgiev:
> Where are we actually fixing #554 ?


with this one

```
metaRange("1", 0, 20, nil),
metaRange("2", 10, 30, nil),
metaRange("3", 40, 60, nil),
metaRange("4", 50, 70, nil),
```

CHANGELOG.md
naivewong: Oh I may want to express `first`, will change later.

compact.go
naivewong: hm you may have a look at the previous @codesome 's [comment](https://github.com/prometheus/tsdb/pull/560#issuecomment-476991384), if we return all, it can possibly result in compacting a very large block.

compact_test.go
naivewong: Sorry, I am a bit confused, why it is not aligned?.. Or do you mean I should use the case in #554 ?

compact_test.go
krasi-georgiev:
```
metaRange("1", 0, 20, nil),
metaRange("2", 19, 40, nil),
metaRange("3", 40, 60, nil),
```

the second block ends at 40 and the third block starts at 40 so the lines should be like
```
//|--------------|
//              |----------------|
//		             	         |--------------|
```

the same for the other places I commented.

codesome: The last sample of "2" and the first sample of "3" overlap I guess? So isn't the representation right?

codesome:
@bwplotka Description of #554 needs updating, so the issue is `selectOverlappingDirs` selects wrong dirs where there are 2 disjoint sets of overlapping dirs, and does not exactly miss any dir.

compact_test.go
krasi-georgiev: i don't see why not. the added blocks can have any ranges

compact_test.go
```diff
@@ -312,32 +300,44 @@ func TestLeveledCompactor_plan(t *testing.T) {
 			},
 			expected: []string{"7", "8"},
 		},
-		// For overlapping blocks.
-		{
+		//|--------------|
+		//              |----------------|
+		//			           |--------------|
```
krasi-georgiev: I expect these should be aligned like

```
//|--------------|
//              |----------------|
//		                         |--------------|
```

compact_test.go
```diff
@@ -312,32 +300,44 @@ func TestLeveledCompactor_plan(t *testing.T) {
...
 				metaRange("2", 19, 40, nil),
 				metaRange("3", 40, 60, nil),
 			},
 			expected: []string{"1", "2"},
 		},
-		{
+		//|--------------|
+		//                 |--------------|
+		//			    |--------------|
+		"Overlapping blocks 2": {
```
krasi-georgiev: //|--------------|
//               |--------------|
//			             |--------------|

krasi-georgiev:
@bwplotka 	do you still any problems in this PR?

krasi-georgiev:
Thanks. If @bwplotka has any other comments will open another PR to address these.

PR #562, opened 2019-03-21
Update wal.md

naivewong:
The format of sample is inconsistent with record.go.
The first row doesn't have field `value`.

### DIFF
--- a/docs/format/wal.md
+++ b/docs/format/wal.md
@@ -54,14 +54,16 @@ Series records encode the labels that identifies a series and its unique ID.
 
 Sample records encode samples as a list of triples `(series_id, timestamp, value)`.
 Series reference and timestamp are encoded as deltas w.r.t the first sample.
+The first row stores the starting id and the starting timestamp.
+The first sample record begins at the second row.
 
 ```
 ┌──────────────────────────────────────────────────────────────────┐
 │ type = 2 <1b>                                                    │
 ├──────────────────────────────────────────────────────────────────┤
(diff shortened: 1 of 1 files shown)


krasi-georgiev:
Can you point to the code for that.

naivewong:
```
// Samples appends the encoded samples to b and returns the resulting slice.
func (e *RecordEncoder) Samples(samples []RefSample, b []byte) []byte {
	buf := encoding.Encbuf{B: b}
	buf.PutByte(byte(RecordSamples))

	if len(samples) == 0 {
		return buf.Get()
	}

	// Store base timestamp and base reference number of first sample.
	// All samples encode their timestamp and ref as delta to those.
	first := samples[0]

	buf.PutBE64(first.Ref)
	buf.PutBE64int64(first.T)

	for _, s := range samples {
		buf.PutVarint64(int64(s.Ref) - int64(first.Ref))
		buf.PutVarint64(s.T - first.T)
		buf.PutBE64(math.Float64bits(s.V))
	}
	return buf.Get()
}
```
you can see the first row just contain starting ref and time.

krasi-georgiev:
yes you are right, I wonder what is the reason for that.Just asked in the dev channel for some more info and will probably add more comments to explain the reason.

btw you can copy [permalinks](https://help.github.com/en/articles/creating-a-permanent-link-to-a-code-snippet) instead of copy pasting the code. It makes it clickable and easier to find.

krasi-georgiev:
yep just confirmed that this is just a design decision and nothing special requires it.

PR #567, opened 2019-03-25
docs: add multi chunks layout in chunks file

zhulongcheng:
docs: add multi chunks layout in chunks file

### DIFF
--- a/docs/format/chunks.md
+++ b/docs/format/chunks.md
@@ -1,15 +1,31 @@
 # Chunks Disk Format
 
-The following describes the format of a single chunks file, which is created in the `chunks/` directory of a block. The maximum size per segment file is 512MiB.
+The following describes the format of a chunks file,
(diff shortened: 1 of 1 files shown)


docs/format/chunks.md
```diff
@@ -1,15 +1,31 @@
 # Chunks Disk Format
 
-The following describes the format of a single chunks file, which is created in the `chunks/` directory of a block. The maximum size per segment file is 512MiB.
+The following describes the format of a chunks file, which is created
```
krasi-georgiev: maybe best for new line as:
```
The following describes the format of a single chunks file, 
which is created in the `chunks/` directory of a block. 
The maximum size per segment file is 512MiB.
```

zhulongcheng: updated.

krasi-georgiev:
Thanks!

PR #568, opened 2019-03-25
Add filtering on ref of samples and tombstones in checkpoint.go

naivewong:
https://github.com/prometheus/tsdb/blob/db9177de0cc7fec169bda0ede005ffd6d6fd1d80/checkpoint.go#L102
As title, I think `keep` function should also be applied on samples and tombstones.

### DIFF
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -193,7 +193,7 @@ func Checkpoint(w *wal.WAL, from, to int, keep func(id uint64) bool, mint int64)
 			// Drop irrelevant samples in place.
 			repl := samples[:0]
 			for _, s := range samples {
-				if s.T >= mint {
+				if keep(s.Ref) && s.T >= mint {
 					repl = append(repl, s)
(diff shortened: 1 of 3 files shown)


krasi-georgiev:
this will decrease performance, but it is the correct behaviour.

LGTM

ping @gouthamve , @codesome , @bwplotka

krasi-georgiev:
btw @naivewong  don't forget that tomorrow (the 9th) is the last day for the GSOC proposal submissions. Feel free  to ping me on IRC with yours so I can give some feedback before submitting.

brian-brazil:
This will lead to data being silently missing for the wal watcher, this change should not be accepted.

krasi-georgiev:
after some IRC discussions the correct behaviour is to keep all series references until the last segment has samples referencing these.

closing this PR in favour of the correct fix in 
https://github.com/prometheus/tsdb/pull/577

PR #570, opened 2019-03-26
cleanup tmp files

### LINKED ISSUE #565: Cleanup temp file if rename/replace failed
https://github.com/prometheus/tsdb/blob/3ab5f4e57902eea758f1940dd852e954b5d5ba6b/compact.go#L1015-L1034

If `to` is removed successfully but failed to rename `from` to `to`,  data in `to` will be lost.

Ex: [OpenBlock](https://github.com/prometheus/tsdb/blob/master/block.go#L280-L334) -> [writeMetaFile](https://github.com/prometheus/tsdb/blob/master/block.go#L233-L258) (re-create a meta file) -> [renameFile](https://github.com/prometheus/tsdb/blob/master/compact.go#L1015-L1034)

The `meta.json` file is removed.
```
drwxr-xr-x  6 logan  staff   192B Mar 26 11:53 ./
drwx------  3 logan  staff    96B Mar 26 11:53 ../
drwxr-xr-x  3 logan  staff    96B Mar 26 11:53 chunks/
-rw-r--r--  1 logan  staff   315B Mar 26 11:53 index
-rw-r--r--  1 logan  staff   264B Mar 26 11:53 meta.json.tmp
-rw-r--r--  1 logan  staff     9B Mar 26 11:53 tombstones
```

---
The callers of [renameFile](https://github.com/prometheus/tsdb/blob/3ab5f4e57902eea758f1940dd852e954b5d5ba6b/compact.go#L1015-L1034) should check and cleanup temp file if `renameFile` return an error.

https://github.com/prometheus/tsdb/blob/3ab5f4e57902eea758f1940dd852e954b5d5ba6b/block.go#L233-L258

---

TODO:
- [ ] a safer way to replace files.
- [ ] remove the temp file.

zhulongcheng:
Closes #565

### DIFF
--- a/block.go
+++ b/block.go
@@ -29,6 +29,7 @@ import (
 	"github.com/prometheus/tsdb/chunkenc"
 	"github.com/prometheus/tsdb/chunks"
 	tsdb_errors "github.com/prometheus/tsdb/errors"
+	"github.com/prometheus/tsdb/fileutil"
 	"github.com/prometheus/tsdb/index"
 	"github.com/prometheus/tsdb/labels"
 )
@@ -230,12 +231,17 @@ func readMetaFile(dir string) (*BlockMeta, error) {
 	return &m, nil
 }
 
-func writeMetaFile(dir string, meta *BlockMeta) error {
+func writeMetaFile(logger log.Logger, dir string, meta *BlockMeta) error {
 	meta.Version = 1
 
 	// Make any changes to the file appear atomic.
 	path := filepath.Join(dir, metaFilename)
 	tmp := path + ".tmp"
+	defer func() {
+		if err := os.RemoveAll(tmp); err != nil {
+			level.Error(logger).Log("msg", "remove tmp file", "err", err.Error())
+		}
+	}()
 
 	f, err := os.Create(tmp)
 	if err != nil {
@@ -246,7 +252,6 @@ func writeMetaFile(dir string, meta *BlockMeta) error {
 	enc.SetIndent("", "\t")
 
 	var merr tsdb_errors.MultiError
-
 	if merr.Add(enc.Encode(meta)); merr.Err() != nil {
 		merr.Add(f.Close())
 		return merr.Err()
@@ -259,7 +264,7 @@ func writeMetaFile(dir string, meta *BlockMeta) error {
 	if err := f.Close(); err != nil {
 		return err
 	}
-	return renameFile(tmp, path)
+	return fileutil.Replace(tmp, path)
 }
 
 // Block represents a directory of time series data covering a continuous time range.
@@ -278,6 +283,8 @@ type Block struct {
 	chunkr     ChunkReader
 	indexr     IndexReader
 	tombstones TombstoneReader
+
+	logger log.Logger
 }
 
 // OpenBlock opens the block in the directory. It can be passed a chunk pool, which is used
@@ -322,7 +329,7 @@ func OpenBlock(logger log.Logger, dir string, pool chunkenc.Pool) (pb *Block, er
 	// that would be the logical place for a block size to be calculated.
 	bs := blockSize(cr, ir, tsr)
 	meta.Stats.NumBytes = bs
-	err = writeMetaFile(dir, meta)
+	err = writeMetaFile(logger, dir, meta)
 	if err != nil {
 		level.Warn(logger).Log("msg", "couldn't write the meta file for the block size", "block", dir, "err", err)
 	}
@@ -334,6 +341,7 @@ func OpenBlock(logger log.Logger, dir string, pool chunkenc.Pool) (pb *Block, er
 		indexr:          ir,
 		tombstones:      tr,
 		symbolTableSize: ir.SymbolTableSize(),
+		logger:          logger,
 	}
 	return pb, nil
 }
@@ -429,7 +437,7 @@ func (pb *Block) GetSymbolTableSize() uint64 {
 
 func (pb *Block) setCompactionFailed() error {
 	pb.meta.Compaction.Failed = true
-	return writeMetaFile(pb.dir, &pb.meta)
+	return writeMetaFile(pb.logger, pb.dir, &pb.meta)
 }
 
 type blockIndexReader struct {
@@ -553,10 +561,10 @@ Outer:
 	pb.tombstones = stones
 	pb.meta.Stats.NumTombstones = pb.tombstones.Total()
 
-	if err := writeTombstoneFile(pb.dir, pb.tombstones); err != nil {
+	if err := writeTombstoneFile(pb.logger, pb.dir, pb.tombstones); err != nil {
 		return err
 	}
-	return writeMetaFile(pb.dir, &pb.meta)
+	return writeMetaFile(pb.logger, pb.dir, &pb.meta)
 }
 
 // CleanTombstones will remove the tombstones and rewrite the block (only if there are any tombstones).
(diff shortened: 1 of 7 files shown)


krasi-georgiev:
Here is another related bug which we should fix here or in a separate PR
https://github.com/prometheus/prometheus/issues/4058#issuecomment-476544672

krasi-georgiev:
how is it going with this one? 

I received you ping on IRC and replied, but maybe you were offline. I suggest using Riot to never miss IRC messaged.

zhulongcheng:
> I suggest using Riot to never miss IRC messaged.

Thanks Krasi

block.go
```diff
@@ -259,7 +260,14 @@ func writeMetaFile(dir string, meta *BlockMeta) error {
 	if err := f.Close(); err != nil {
 		return err
 	}
-	return renameFile(tmp, path)
+
+	if err := fileutil.Rename(tmp, path); err != nil {
```
zhulongcheng: @krasi-georgiev should replace `renameFile` with `fileutil.Rename`?

Maybe we should check all callers of `renameFile`:
1. repalce `renameFile` with `fileutil.Rename` for renaming a file.
2. repalce `renameFile` with `fileutil.Replace` for renaming a dir.

krasi-georgiev: Ahh good catch. I think option 1 looks best here.
Replace all instances of `renameFile` with `filetuil.Rename`, 
remove `renameFile`,
check that all callers of `filetuil.Rename` handle an error properly by deleting any left over files/dirs.



not sure why `renameFile` has this 
```
if err := os.RemoveAll(to); err != nil {
		return err
	}
```
`os.Rename` should handle this so it shouldn't be needed. Maybe double check to confirm.

zhulongcheng: updated.

krasi-georgiev: I think best to put it in a defer statement similar to the one in `func (c *LeveledCompactor) write`

zhulongcheng: ok

krasi-georgiev:
btw @zhulongcheng don't forget that tomorrow the 9th is the last day for the GSOC proposal submissions. Feel free  to ping me on IRC with yours so I can give some feedback before submitting.

krasi-georgiev:
a file conflict and did you test that

``
if err := os.RemoveAll(to); err != nil {
		return err
	}
```
is not needed?

**Comment by zhulongcheng** [comment]
Will rebase ASAP.

About test:
For replacing files, `os.RemoveAll` is not needed.
For replacing dirs, `os.RemoveAll` is required to ensure the dest dir not exists.

**Comment by krasi-georgiev** [comment]
so IIUC we do need it since fileutil.Rename should also handle dirs.

**Comment by zhulongcheng** [comment]
> so IIUC we do need it since fileutil.Rename should also handle dirs.

How about `fileutil.Replace`?
https://github.com/prometheus/tsdb/blob/4b3a5ac5d36e5262d2656c8d149e137c2d1fab12/fileutil/fileutil.go#L128-L149

**Comment by krasi-georgiev** [comment]
do you mean to use Rename only for files and Replace for files or dirs?

is there no way to combine these into a single func?

File: compact.go [hunk 2]
```diff
@@ -639,7 +639,7 @@ func (c *LeveledCompactor) write(dest string, meta *BlockMeta, blocks ...BlockRe
 	df = nil
 
 	// Block successfully written, make visible and remove old ones.
-	if err := renameFile(tmp, dir); err != nil {
+	if err := fileutil.Replace(tmp, dir); err != nil {
```
krasi-georgiev: why Replace and not Rename?

repair.go
```diff
@@ -70,21 +72,30 @@ func repairBadIndexVersion(logger log.Logger, dir string) error {
...
 			return wrapErr(err, d)
 		}
 		if err := broken.Close(); err != nil {
 			return wrapErr(err, d)
 		}
-		if err := renameFile(repl.Name(), broken.Name()); err != nil {
-			return wrapErr(err, d)
+		if err := fileutil.Rename(repl.Name(), broken.Name()); err != nil {
+			merr.Add(wrapErr(err, d))
+			merr.Add(os.RemoveAll(repl.Name()))
```
krasi-georgiev: shouldn't repl be removed after any error 
```
repl, err := os.Create(filepath.Join(d, "index.repaired"))
if err != nil {
	return wrapErr(err, d)
}

defer func() {
	os.RemoveAll(repl.Name())
}()
```

wal.go
```diff
@@ -389,9 +393,10 @@ func (w *SegmentWAL) Truncate(mint int64, keep func(uint64) bool) error {
 	csf.Close()
 
 	candidates[0].Close() // need close before remove on platform windows
-	if err := renameFile(csf.Name(), candidates[0].Name()); err != nil {
+	if err := fileutil.Rename(csf.Name(), candidates[0].Name()); err != nil {
```
krasi-georgiev: this one looks like should be `fileutil.Replace` as `candidates[0].Name()` might already exist.

repair.go
zhulongcheng: fixed

compact.go
zhulongcheng: fixed

wal.go
zhulongcheng: fixed

block.go
```diff
@@ -259,7 +263,12 @@ func writeMetaFile(dir string, meta *BlockMeta) error {
...
```
krasi-georgiev: this one should also be Replace as in some cases the destination file might already exist

like in : https://github.com/prometheus/tsdb/blob/e46ec89de67d3d8fc91720ae3cba6f9c7a3320b0/block.go#L334

tombstones.go
```diff
@@ -64,6 +65,7 @@ func writeTombstoneFile(dir string, tr TombstoneReader) error {
 		if f != nil {
 			f.Close()
 		}
+		os.RemoveAll(tmp)
```
krasi-georgiev: lets add some logging if this fails like you did in the other place:
```
				level.Error(logger).Log("msg", "remove tmp file", "err", err.Error())
```

block.go
zhulongcheng: ok.

zhulongcheng:
Maybe we need to reconsider whether we still need the `fileutil.Rename` function?

If  the destination file/dir exists, should use `fileutil.Replace`. 
If  the destination file/dir not exists, use `fileutil.Replace`, and `os.RemoveAll` (in `fileutil.Replace` func) just does nothing.

krasi-georgiev:
I am concerned that this might lead to unexpected data loss. For example if the generated destination ULID of a block collapses with an existing block Replace will just delete the destination without us even knowing about this leading to data loss.

zhulongcheng:
I want to simplify this pr.
just do two things:
1. replace `renameFile` with `fileutil.Replace`, and remove the `renameFile` func.
2. cleanup tmp files.

because the implementation of  `renameFile` is identical to `fileutil.Replace`, the replace operation is safe.

this will make this pr clear and be reviewed friendly.

zhulongcheng:
new a separate issue to discuss implementations and usages of Rename and Replace.

krasi-georgiev:
ok

krasi-georgiev:
if https://github.com/prometheus/tsdb/pull/583 turns out to be a good solution to inject FS errors can use it to write a test for this PR.

krasi-georgiev:
LGTM
no need to wait for #583 as it would probably take a while and we can write that test later.

ping @gouthamve , @codesome , @bwplotka

krasi-georgiev:
since it is a low risk PR , will merge and if anymore changes are needed will open a new one.

PR #256, opened 2018-01-13
bench: Fix path to default sample file

mattbostock:
The sample file used for benchmarking was renamed in 8326e410d01a5 but
the `--file` flag default was not updated.

fixes #305

PR #571, opened 2019-03-27
Fixed default benchmark test data path

VladMasarik:
Default benchmark test was not working although the test data is present.

Might not be groundbreaking, but might saves some confusion in case you execute `go build` in `cmd/tsdb`. Also, it seems that this was introduced in #428 ; before #256

I could not find an Issue for this "bug".


krasi-georgiev:
wondering whether we should have relative paths at all. Since you can build and run it even from the root folder like `go build cmd/tsdb/main.go` in which case the path should be just as `filepath.Join("testdata", "20kseries.json")`

but anyway lets merge as is and will change again if needed.

krasi-georgiev:
Thanks

PR #573, opened 2019-04-01
use fsync to force persisting the meta file.

krasi-georgiev:
fixes https://github.com/prometheus/prometheus/issues/4058

### DIFF
--- a/block.go
+++ b/block.go
@@ -252,7 +252,7 @@ func writeMetaFile(dir string, meta *BlockMeta) error {
 		return merr.Err()
 	}
 	// Force the kernel to persist the file on disk to avoid data loss if the host crashes.
-	if err := f.Sync(); err != nil {
+	if merr.Add(f.Sync()); merr.Err() != nil {
 		merr.Add(f.Close())
 		return merr.Err()
 	}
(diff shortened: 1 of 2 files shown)


block.go
```diff
@@ -251,6 +251,10 @@ func writeMetaFile(dir string, meta *BlockMeta) error {
 		merr.Add(f.Close())
 		return merr.Err()
 	}
+	// Force the kernel to persist the file on disk to avoid data loss if the host crashes.
+	if err := f.Sync(); err != nil {
```
pborzenkov: Hmm. Shouldn't you close the file in this case to avoid fd leak?

krasi-georgiev: yep! updated

pborzenkov: There is also a tsdb.fileutil.Sync() function which issues correct sync on macOS (default fsync doesn't issue disk barrier). It's probably better to use it.
This has been fixed in Go 1.12, so if Prometheus/TSDB require at least 1.12 it's not a big deal.

krasi-georgiev: yeah I noticed this and some parts in the code use fileutil.Fsync and others use the regular one.
I will check if there is any blocker to require 1.12 for tsdb/Prometheus and will probably open another PR to remove the workaround.

block.go
```diff
@@ -251,6 +251,11 @@ func writeMetaFile(dir string, meta *BlockMeta) error {
...
```
gouthamve: We should also do `merr.Add(err)` so that we don't drop the error from fsync.

PR #574, opened 2019-04-01
remove Fsync workaround for macos.

krasi-georgiev:
since golang 1.12 no special handling is required for file.Sync()

@pborzenkov thanks for the pointer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,6 @@
 
 This repository contains the Prometheus storage layer that is used in its 2.x releases.
 
-It requires golang 1.12 or higher.
-
 A writeup of its design can be found [here](https://fabxc.org/blog/2017-04-10-writing-a-tsdb/).
 
 Based on the Gorilla TSDB [white papers](http://www.vldb.org/pvldb/vol8/p1816-teller.pdf).
--- a/README.md
+++ b/README.md
@@ -6,6 +6,8 @@
 
 This repository contains the Prometheus storage layer that is used in its 2.x releases.
 
+It requires golang 1.12 or higher.
+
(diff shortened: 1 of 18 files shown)


README.md
```diff
@@ -6,6 +6,8 @@
 
 This repository contains the Prometheus storage layer that is used in its 2.x releases.
 
+It requires golang 1.12 or higher.
```
brian-brazil: Will we remember to keep this up to date?

krasi-georgiev: Is there any other way to enforce or notify the users of the the library that it requires 1.12 or higher?

krasi-georgiev: just found this and will try it
https://medium.com/@theckman/version-constraints-and-go-c9309be15773

simonpasquier: You could add `go 1.12` to `go.mod` though it won't prevent anybody to try building the library with an older version of Go. From https://golang.org/doc/go1.12#modules

> The go directive in a go.mod file now indicates the version of the language used by the files within that module. It will be set to the current release (go 1.12) if no existing version is present. If the go directive for a module specifies a version newer than the toolchain in use, the go command will attempt to build the packages regardless, and will note the mismatch only if that build fails.

krasi-georgiev: Aaah that is nice, thanks Simon.  
The description in godoc is a bit confusing though.

I implemented the suggested method from the blog post which will enforce a build failure for versions older than 1.12. 

Another approach would be to print a warning log in an init() func if the golang version is older than 1.12.

krasi-georgiev: I think best to leave the current enforcement and once golang progresses few more versions will switch to the go modules way.

PR #575, opened 2019-04-02
chunks: close all opened MmapFiles on NewDirReader error path

pborzenkov:


pborzenkov:
CI failed on go1.10/macOS:

```
>> getting dependencies
go get  -t ./...
# golang.org/x/sys/unix
../../../golang.org/x/sys/unix/zsyscall_darwin_amd64.1_11.go:687:22: undefined: SYS_CLOCK_GETTIME
make: *** [common-deps] Error 2
```

krasi-georgiev:
I wouldn't worry about that since we won't support 1.10 anyway.

Wondering whether a test here would be useful. 
Something like : 
open a dir with a single corrupted file to trigger the error.
Next try to delete all files. Without this fix the test would fail under windows with an error that the file is still open.

pborzenkov:
@krasi-georgiev Hmm, the problem is that we need to trigger an error in OpenMmapFile, which doesn't perform any validation at all. So either some kind of fault injection is required, or an OpenMmapFile mock.

krasi-georgiev:
Had a quick look and don't have any good ideas either so lets leave it as is for now.

krasi-georgiev:
Thanks for the bugfix!

Issue #21, opened 2017-03-24
Unknown series references in WAL after crash


fabxc:
After ungraceful termination sometimes samples in the WAL reference series that do not exist yet. This should strictly not happen as all newly created series are added (in-order) to the WAL before any of their samples.

Needs investigation but no critical impact at this stage.

devnev:
We're running a prometheus2 instance alongside our prometheus instances to observe the system's behaviour as we'd like to switch to P2 as soon as its ready. This crash seems insofar impacting as it can get a prometheus2 completely stuck in a crashloop: Our instance logs `Opening storage failed: open block /prometheus-data/01BJ772AN91568QNCGANK95GM7: open head block /prometheus-data/01BJ772AN91568QNCGANK95GM7: unknown series reference 101615 (max 101608); abort WAL restore" source="main.go:89"` on startup and terminates, to then be restarted by k8s and repeat.

mwitkow:
https://github.com/prometheus/prometheus/issues/2795 is the same thing :(

fabxc:
Did either of you ever observe this again?

WuLeeX:
Any updates on this？

roidelapluie:
I observe that behaviour too after arn early restart of prometheus

roidelapluie:
Jul 19 13:59:50 nip-prometheus1 prometheus[27159]: level=warn ts=2018-07-19T11:59:50.793064439Z caller=head.go:320 component=tsdb msg="unknown series references in WAL samples" count=84

roidelapluie:
I have an impact as some recording rules do not have the correct timestamp it seems. Like they are delayed by a few minutes.

krasi-georgiev:
Initially I thought that this is caused by more than one prometheus writing to the same data folder, but @brian-brazil found some parts of the code that are more likely to cause this.
https://github.com/prometheus/tsdb/pull/332#discussion_r195148633

It is best to get the WAL rewrite merged and than work on this bug after.

JorritSalverda:
Seeing this in 2.6.0 as well, after having upgraded from 2.3.2 a couple of days ago.

```
level=info ts=2019-01-16T19:14:28.556334137Z caller=repair.go:48 component=tsdb msg="found healthy block" mint=1546545600000 maxt=1546552800000 ulid=01D18NC97NQE2TMPDRPS77T1KJ
[...]

PR #577, opened 2019-04-04
Keep series that are still in WAL in checkpoints

brian-brazil:
If all the samples are deleted for a series,
we should still keep the series in the WAL as
anything else reading the WAL will still care
about it in order to understand the samples.

fixes: https://github.com/prometheus/tsdb/issues/21

Signed-off-by: Brian Brazil <brian.brazil@robustperception.io>

### DIFF
--- a/head.go
+++ b/head.go
@@ -75,6 +75,9 @@ type Head struct {
 	symbols map[string]struct{}
 	values  map[string]stringset // label names to possible values
 
+	deletedMtx sync.Mutex
+	deleted    map[uint64]int // Deleted series, and what WAL segment they must be kept until.
+
 	postings *index.MemPostings // postings lists for terms
 }
 
@@ -234,6 +237,7 @@ func NewHead(r prometheus.Registerer, l log.Logger, wal *wal.WAL, chunkRange int
 		values:     map[string]stringset{},
 		symbols:    map[string]struct{}{},
 		postings:   index.NewUnorderedMemPostings(),
+		deleted:    map[uint64]int{},
 	}
 	h.metrics = newHeadMetrics(h, r)
 
@@ -557,7 +561,13 @@ func (h *Head) Truncate(mint int64) (err error) {
 	}
 
 	keep := func(id uint64) bool {
-		return h.series.getByID(id) != nil
+		if h.series.getByID(id) != nil {
+			return true
+		}
+		h.deletedMtx.Lock()
+		_, ok := h.deleted[id]
+		h.deletedMtx.Unlock()
+		return ok
 	}
 	h.metrics.checkpointCreationTotal.Inc()
 	if _, err = Checkpoint(h.wal, first, last, keep, mint); err != nil {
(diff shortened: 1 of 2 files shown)


head.go
```diff
@@ -953,6 +974,15 @@ func (h *Head) gc() {
 	// Remove deleted series IDs from the postings lists.
 	h.postings.Delete(deleted)
 
+	if h.wal != nil {
+		_, last, _ := h.wal.Segments()
+		h.deletedMtx.Lock()
+		for ref := range deleted {
```
cstyan: I think it would be worth including a comment here:
```suggestion
                // Keep series records until we're past segment 'last' because the WAL will still have
                // samples records with this ref ID. If we didn't keep these series records then on start up
                // when we replay the WAL, or any other code that reads the WAL, wouldn't be able to use
                // those samples since we would have no labels for that ref ID.
		for ref := range deleted {
```

krasi-georgiev:
@brian-brazil that should also fix https://github.com/prometheus/tsdb/issues/21 right?

krasi-georgiev:
I think instead of this we should just drop all irrelevant samples as per this PR
https://github.com/prometheus/tsdb/pull/568

brian-brazil:
I believe this will fix #21. There should be no irrelevant samples.

head.go
brian-brazil: Done

krasi-georgiev:
> I think instead of this we should just drop all irrelevant samples as per this PR
> #568

I think this should be the correct fix for this. Instead of keeping series we should just delete the irrelevant samples.

head.go
```diff
@@ -570,6 +580,17 @@ func (h *Head) Truncate(mint int64) (err error) {
 		// that supersedes them.
 		level.Error(h.logger).Log("msg", "truncating segments failed", "err", err)
 	}
+
+	// The checkpoint is written and segments before it truncated, so we no
```
krasi-georgiev:
```suggestion
	// The checkpoint is written and segments before it is truncated, so we no
```

head.go
```diff
@@ -75,6 +75,9 @@ type Head struct {
 	symbols map[string]struct{}
 	values  map[string]stringset // label names to possible values
 
+	deletedMtx sync.Mutex
+	deleted    map[uint64]int // Deleted series, and what WAL segment they must be kept until.
```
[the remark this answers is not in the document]
brian-brazil: Your suggestion doesn't make sense. This is a data structure, not an algorithm.

head_test.go
```diff
@@ -501,6 +501,57 @@ func TestDeleteUntilCurMax(t *testing.T) {
 	testutil.Ok(t, err)
 	testutil.Equals(t, []tsdbutil.Sample{sample{11, 1}}, ressmpls)
 }
+
+func TestDeletedSamplesAndSeriesStillInWALAfterCheckpoint(t *testing.T) {
```
krasi-georgiev: Maybe extend the test a bit like:
After the checkpoint add more samples for existing series. Read the wal from scratch and ensure that there are no samples with unknown series references.

head.go
brian-brazil: Done

head_test.go
```diff
@@ -501,6 +501,57 @@ func TestDeleteUntilCurMax(t *testing.T) {
...
+	numSamples := 10000
+	hb, err := NewHead(nil, nil, wlog, int64(numSamples)*10)
+	testutil.Ok(t, err)
+	defer hb.Close()
+	for i := 0; i < numSamples; i++ {
+		app := hb.Appender()
+		_, err := app.Add(labels.Labels{{"a", "b"}}, int64(i), 0)
+		testutil.Ok(t, err)
+		testutil.Ok(t, app.Commit())
+	}
```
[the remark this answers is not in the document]
brian-brazil: That'd clash with the existing name, plus that function doesn't have the WAL handling I need.

krasi-georgiev:
LGTM

PR #581, opened 2019-04-10
make Close methods for the querier safe to call more than once.

krasi-georgiev:
from: https://github.com/prometheus/prometheus/issues/5408

`querier.Close` is called more than once somewhere in the rules Manager. I will still troubleshoot why this happens.

Regardless of this the method should be safe to be called more than once and this PR fixes this and also adds tests for the WAL and querier Close methods.

### DIFF
--- a/querier.go
+++ b/querier.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"sync"
 
 	"github.com/pkg/errors"
 	"github.com/prometheus/tsdb/chunkenc"
@@ -205,6 +206,8 @@ type blockQuerier struct {
 	chunks     ChunkReader
 	tombstones TombstoneReader
 
+	close sync.Once
+
 	mint, maxt int64
 }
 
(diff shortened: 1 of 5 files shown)


bwplotka:
Unless it is urgent fix, but hopefully we can find the root cause, let me know, can try to debug this with you.

krasi-georgiev:
Yes I will continue looking for the cause, but regardless, calling any public API should not cause a panic on a library.

querier.go
```diff
@@ -253,12 +255,14 @@ func (q *blockQuerier) LabelValuesFor(string, labels.Label) ([]string, error) {
 
 func (q *blockQuerier) Close() error {
 	var merr tsdb_errors.MultiError
-
-	merr.Add(q.index.Close())
-	merr.Add(q.chunks.Close())
-	merr.Add(q.tombstones.Close())
-
-	return merr.Err()
+	if !q.closed {
```
bwplotka: Usually in Golang you error out things first to have larger code blocks in "normal" flow. Can we fix that here too? Also `merr` variable should be definited just before usage.

krasi-georgiev: updated, thanks!

Issue #579, opened 2019-04-06
Tests with injected file system errors.


krasi-georgiev:
One case it to catch bugs like:
https://github.com/prometheus/tsdb/pull/575
https://github.com/prometheus/tsdb/issues/565 - A test for this woule look like. Inject an error at some point for these funcs and check that no tmp files is left behind.
An example test:
https://github.com/prometheus/tsdb/pull/576

This article is a good start.
https://medium.com/@siddontang/use-fuse-to-inject-failure-to-i-o-deb5f2e7800a

geekodour:
would like to work on this as it's not a high priority one.

pborzenkov:
@krasi-georgiev Alternative approach would be to use a filesystem abstraction interface (something like https://github.com/spf13/afero) and instantiate it with specially crafted fault-injecting implementation for tests.

krasi-georgiev:
@pborzenkov yes I looked at it , but didn't see anything in the interface to trigger different faults. It looks like its main purpose is to NOT use a real file system and avoid the cleanup after each test.

@pborzenkov can you point to some examples how can afero be used for our use case?

pborzenkov:
@krasi-georgiev I meant something like this:
  - switch all TSDB code that accesses FS to use afero interface
  - in normal conditions, instantiate the interface with real FS implementation (taken from afero)
  - in test, instantiate it with some custom (yet to be written) implementation that allows to inject faults (via callback, random or any other method).

I haven't seen such implementation in the wild, just wanted to say that it might be more useful as it may provide much more granular control of where and when to inject faults.

krasi-georgiev:
yes that is a good suggestion, but I would like to see if we can do this without any  additional intrerfaces. Using afero in the actual tsdb code would add using interfaces and in some hot paths this will have some negative performance impact because of  the interface dereferencing. I have noticed that Fabian removed some interfaces in the past to improve some performance. I think the WAL was an interface and now is passed as a simple struct.

pborzenkov:
Yeah, in this case it's probably better to avoid interfaces. Another implementation of granular fault injections might be something like this:
  - insert static faults into points of interest
  - make them noop by default (via build tags)
  - in tests, build with faults enabled and control which ones to enable

E.g. something like this:

`
func OpenMmapFile(path string) (*MmapFile, error) {
    if err := faults.IsFailed("OpenMmapFile"); err != nil {
        return nil, err
    }

    <normal implementation>
}
`

[...]

PR #582, opened 2019-04-15
fix wal panic when page flush fails.

krasi-georgiev:
fixes: https://github.com/prometheus/prometheus/issues/3283

New records should be added to the page only when the last flush
succeeded. Otherwise the page would be full and panics when trying to
add a new record.

to replicate the bug:
```
dd if=/dev/zero of=tmp/data.img bs=1M count=20 // create a 20mb disk image
mkfs.ext4 /tmp/data.img
mkdir /tmp/prometheus/
sudo mount -t ext4 -o loop /tmp/data.img /tmp/prometheus/
GO111MODULE=on go run cmd/prometheus/main.go   --config.file=.local/simpleMany.yaml --storage.tsdb.path=/tmp/prometheus   --storage.tsdb.wal-segment-size=1MB
```

before the fix this will panic when the disk image is full because it is trying to add more records to the page when it is already full.

At the moment can't think of simple test to test for this, but will add it to https://github.com/prometheus/tsdb/issues/579 if we figure out how to inject file system faults.

### DIFF
--- a/wal/wal.go
+++ b/wal/wal.go
@@ -429,7 +429,6 @@ func (w *WAL) flushPage(clear bool) error {
 	// No more data will fit into the page. Enqueue and clear it.
 	if clear {
 		p.alloc = pageSize // Write till end of page.
-		w.pageCompletions.Inc()
 	}
 	n, err := w.segment.Write(p.buf[p.flushed:p.alloc])
 	if err != nil {
@@ -445,6 +444,7 @@ func (w *WAL) flushPage(clear bool) error {
 		p.alloc = 0
 		p.flushed = 0
 		w.donePages++
+		w.pageCompletions.Inc()
 	}
 	return nil
 }
@@ -499,6 +499,13 @@ func (w *WAL) Log(recs ...[]byte) error {
 // the final record of a batch, the record is bigger than the page size or
 // the current page is full.
 func (w *WAL) log(rec []byte, final bool) error {
+	// When the last page flush failed the page will remain full.
+	// When the page is full, need to flush it before trying to add more records to it.
+	if w.page.full() {
+		if err := w.flushPage(true); err != nil {
+			return err
+		}
+	}
 	// If the record is too big to fit within the active page in the current
 	// segment, terminate the active segment and advance to the next one.
 	// This ensures that records do not cross segment boundaries.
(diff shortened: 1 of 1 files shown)


wal/wal.go
```diff
@@ -429,7 +429,6 @@ func (w *WAL) flushPage(clear bool) error {
 	// No more data will fit into the page. Enqueue and clear it.
```
bwplotka: Comment is not strictly true -> clear bool also matters (:

wal/wal.go
```diff
@@ -499,6 +499,13 @@ func (w *WAL) Log(recs ...[]byte) error {
 // the final record of a batch, the record is bigger than the page size or
```
bwplotka: Can we modify this comment to be bullet list ? Would easier to enumerate those 3 options.

wal/wal.go
```diff
@@ -429,7 +429,6 @@ func (w *WAL) flushPage(clear bool) error {
 	// No more data will fit into the page. Enqueue and clear it.
 	if clear {
 		p.alloc = pageSize // Write till end of page.
-		w.pageCompletions.Inc()
 	}
 	n, err := w.segment.Write(p.buf[p.flushed:p.alloc])
 	if err != nil {
```
bwplotka: Line 437: What if the clear is true, and we do `Write` and `n` is not actually `p.alloc`? I think we would then drop some bytes, right? Now is the question if the implementation of `segment.Write` does partial writes in any case (`n`).  Worth to check?

bwplotka: I know it's not your bug but: https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html

krasi-georgiev: > I think we would then drop some bytes, right? 

Not sure what you mean by this. Can you add more details.

```
n, err := w.segment.Write(p.buf[p.flushed:p.alloc])
	if err != nil {
		return err
	}
	p.flushed += n
```
When `Write` returns an error tsdb will exit(leaving the WAL in corrupted state) and will run a repair  on the next start up.

wal/wal.go
krasi-georgiev: updated

wal/wal.go
krasi-georgiev: updated

wal/wal.go
krasi-georgiev: ping @bwplotka

krasi-georgiev:
merging as is.
@bwplotka if you still think there is a problem please let me know and I will open another PR.

PR #585, opened 2019-04-18
Avoid potential collision of createIdxChkReaders in querier_test.go

naivewong:
as title, make ref a monotonically increasing number to avoid the potential collision
https://github.com/prometheus/tsdb/blob/4b3a5ac5d36e5262d2656c8d149e137c2d1fab12/querier_test.go#L213-L214


krasi-georgiev:
Thanks

PR #588, opened 2019-04-24
Open db in Read only mode

### LINKED ISSUE #516: TSDB CLI tool doesn't work on live Prometheus

krasi-georgiev:
a better approach than #541 (Thanks to Bartek's review.)
 
fixes #516

TODO:
 - [ ] open an issue for the TODO comments refactoring.


db.go
```diff
@@ -240,6 +245,94 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
...
+
+	if len(loadable) == 0 {
+		return nil, errors.New("no blocks found")
+	}
+
+	sort.Slice(loadable, func(i, j int) bool {
+		return loadable[i].Meta().MinTime < loadable[j].Meta().MinTime
+	})
+
+	db.db.blocks = loadable
```
krasi-georgiev: I will think how to remove the `db.db` stutter. Suggestions are welcome.

db.go
```diff
@@ -240,6 +245,94 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
 	return m
 }
 
+// DBReadOnly provides APIs for read only operations on a database.
+type DBReadOnly struct {
+	db *DB
+}
+
+// OpenReadOnly returns a new DB in the given directory only for read operations.
+func OpenReadOnly(dir string, l log.Logger, r prometheus.Registerer) (db *DBReadOnly, err error) {
```
brian-brazil: This duplicates a fair bit of code compared to the main open function, how will we ensure they stay in sync?

krasi-georgiev: Don't think we need to as it has completely different purpose.

db.go
```diff
@@ -240,6 +245,94 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
 	return m
 }
 
+// DBReadOnly provides APIs for read only operations on a database.
+type DBReadOnly struct {
+	db *DB
```
bwplotka: I agree with @brian-brazil and I think this is the problematic bit. I think we can reuse more if we would do vice versa. Instead of this we would have some `Querier` and `Blocker` (or `RODB`) interfaces within `DB` (: 

This has also negative effect of DBReadOnly code having an access to write methods into underlying `db`. This might cause future bugs as well.

Wonder if that is doable without any major refactors, let me experiment for a bit.

bwplotka: something like `DBReader` with `Querier()` and `Block()` methods

db.go
```diff
@@ -240,6 +245,94 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
...
+func OpenReadOnly(dir string, l log.Logger, r prometheus.Registerer) (db *DBReadOnly, err error) {
+	if l == nil {
+		l = log.NewNopLogger()
+	}
+
+	if _, err := os.Stat(dir); err != nil {
+		return nil, err
+	}
+
+	head, err := NewHead(r, l, nil, 1)
```
bwplotka: I think you meant to add 
```
var wlog *wal.WAL
	segmentSize := wal.DefaultSegmentSize
	// Wal is enabled.
	if opts.WALSegmentSize >= 0 {
		// Wal is set to a custom size.
		if opts.WALSegmentSize > 0 {
			segmentSize = opts.WALSegmentSize
		}
		wlog, err = wal.NewSize(l, r, filepath.Join(dir, "wal"), segmentSize)
		if err != nil {
			return nil, err
		}
	}
```

bwplotka:
Also we need to think about our use case. In current design we have only snaphot reader. There is no one syncing blocks if some writer changes it

db.go
bwplotka: Alternative is not use any relation. I think this is pretty straightforward then if we need only blocks and Querier

bwplotka: https://github.com/prometheus/tsdb/pull/589

bwplotka:
Sorry for another PR: https://github.com/prometheus/tsdb/pull/589

db.go
```diff
@@ -247,34 +247,73 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
 
 // DBReadOnly provides APIs for read only operations on a database.
 type DBReadOnly struct {
-	db *DB
+	logger     log.Logger
+	dir        string
+	registerer prometheus.Registerer
```
bwplotka: Do you need metrics really? All the CLI jobs are batch jobs.

db.go
```diff
@@ -322,8 +428,17 @@ func Open(dir string, l log.Logger, r prometheus.Registerer, opts *Options) (db
 		minValidTime = blocks[len(blocks)-1].Meta().MaxTime
 	}
 
-	if err := db.head.Init(minValidTime); err != nil {
-		return nil, errors.Wrap(err, "read WAL")
+	if initErr := db.head.Init(minValidTime); initErr != nil {
+		err := errors.Cause(initErr) // So that we can pick up errors even when wrapped.
+		if _, ok := err.(*wal.CorruptionErr); ok {
```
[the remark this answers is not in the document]
bwplotka: Can we have separate PR for this?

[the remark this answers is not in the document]
krasi-georgiev: No there is no error, this should be part of this PR. 
`db.head.Init` was doing a repair when the WAL is corrupted and with this change this repair is now moved in tsdb.Open. 
This is needed so that when we create a querier need to load the WAL for the Head but not try to repair it when corrupted.

If you look at the refactored test it should make it more clear  why this change was needed.

db.go
```diff
@@ -240,6 +245,107 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
...
+	if err := db.head.Init(minValidTime); err != nil {
+		return nil, errors.Wrap(err, "read WAL")
+	}
+
+	return db.Querier(mint, maxt)
+}
+
+// Blocks returns the databases persisted blocks.
+func (dbRead *DBView) Blocks() ([]*Block, error) {
+	db := &DB{
```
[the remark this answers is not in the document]
krasi-georgiev: I would say lets leave as is and will refactor if/when needed.

db.go
krasi-georgiev: Yeah I have removed it and will add when/if needed.

I use these for tests all the time when I want to check some internal state during the execution, but there is no immediate need for it right now.

db.go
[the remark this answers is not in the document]
bwplotka: I disagree, and won't approve this PR with this. This is a blocker IMO

db.go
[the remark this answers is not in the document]
bwplotka: OK, can agree, will rereview

db.go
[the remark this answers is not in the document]
krasi-georgiev: I just double checked and It feels that the refactoring that is needed to completely detach it from the db struct might allow more unexpected bugs than keeping it detached. For example the shared `chunkenc.Pool` , checking if a block is not already open etc.

I am thinking if I add a test that will ensure that running the `DBReadOnly` funcs don't modify any files on disk should be sufficient to catch write operations.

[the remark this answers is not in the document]
bwplotka: I doubt that this much of refactoring is needed. What exactly is required? Even my proposal had low impact and I was able to not use `db`: https://github.com/prometheus/tsdb/pull/589

> I am thinking if I add a test that will ensure that running the DBReadOnly funcs don't modify any files on disk should be sufficient to catch write operations.

This is kind of wrong and overeningeering IMO. Why we are trying to spend time on not needed tests instead of doing this properly? In the very worst scenario just copied needed method like `getBlock` etc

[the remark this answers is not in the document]
krasi-georgiev: I see both methods as equally wrong and right and tests will be needed sooner or later so I will sit on it few days and will revisit.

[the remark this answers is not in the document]
krasi-georgiev: I revisited this and although I think both methods have dissadvantages, I have refactored it to your suggestions and will try to revisit again at some point.

krasi-georgiev:
@bwplotka , @brancz 

I did a bit more work on this and I think it is ready for a final review.

db.go
```diff
@@ -602,24 +741,24 @@ func (db *DB) reload() (err error) {
 	return errors.Wrap(db.head.Truncate(maxt), "head truncate failed")
 }
 
-func (db *DB) openBlocks() (blocks []*Block, corrupted map[ulid.ULID]error, err error) {
-	dirs, err := blockDirs(db.dir)
+func openBlocks(l log.Logger, dir string, loaded []*Block, chunkPool chunkenc.Pool) (blocks []*Block, corrupted map[ulid.ULID]error, err error) {
+	blockDirs, err := blockDirs(dir)
```
codesome: I prefer keeping different names for the variable and the function to avoid confusion.

db.go
```diff
@@ -240,6 +245,131 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
...
+}
+
+// Blocks returns all persisted blocks.
+func (db *DBReadOnly) Blocks() ([]*Block, error) {
+	loadable, corrupted, err := openBlocks(db.logger, db.dir, nil, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	// Corrupted blocks that have been replaced by parents can be safely ignored.
```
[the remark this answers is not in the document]
krasi-georgiev: nope it is exactly as my comment.

krasi-georgiev:
> Are we planning to guarantee "read only" here? Because it is still possible to call Delete() and CleanTombstones() from a Block. If we want to assure read-only at all levels, we may need a BlockReadOnly.

good point , updated with a read only interface for a block.

head.go
```diff
@@ -329,6 +323,18 @@ func (h *Head) loadWAL(r *wal.Reader) error {
 	)
 	wg.Add(n)
 
+	defer func() {
+		// For CorruptionErr ensure to terminate all workers before exiting.
+		if _, ok := err.(*wal.CorruptionErr); ok {
```
codesome: Should we check for `err != nil` before?

db.go
```diff
@@ -246,6 +251,145 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
...
+	// Close all previously open blocks and add the new ones to the catche.
+	for _, b := range db.blocks {
+		b.Close()
+	}
+
+	blocks := make([]BlockReadOnly, len(loadable))
+	for i, b := range loadable {
+		blocks[i] = b
+	}
+	db.blocks = blocks
```
codesome: For the current use case, this looks fine. We can run into race conditions if it used concurrently by multiple goroutines in the future (more like wherever we are using `db.blocks`). We can address this in the future if needed, just adding a note here.

db.go
```diff
@@ -328,8 +472,19 @@ func Open(dir string, l log.Logger, r prometheus.Registerer, opts *Options) (db
 		minValidTime = blocks[len(blocks)-1].Meta().MaxTime
 	}
 
-	if err := db.head.Init(minValidTime); err != nil {
-		return nil, errors.Wrap(err, "read WAL")
+	if initErr := db.head.Init(minValidTime); initErr != nil {
+		// So that we can pick up errors even when wrapped.
+		if err := errors.Cause(initErr); err != nil {
```
codesome: If `err` turns out to be `nil` (which I guess not), shouldn't we return `errors.Wrap(err, "read WAL")` like before?

head.go
codesome: I guess it changed again the in last commit.

krasi-georgiev: don't think it is necessary, but doesn't hurt so I updated it.

db.go
```diff
@@ -246,6 +251,145 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
...
+
+	db := &DBReadOnly{
+		logger: l,
+		dir:    dir,
+	}
+
+	return db, nil
+}
+
+// Querier loads the wal and returns a new querier over the data partition for the given time range.
```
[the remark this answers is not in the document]
krasi-georgiev: maybe shouldn't even need to be that specific as this describes the implementation. I think it should be just 
`// Querier returns a new querier over the data partition for the given time range.`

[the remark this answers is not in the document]
codesome: Should be fine. Also, I assume the reason to load WAL and blocks every time is because we would use it with already running Prometheus.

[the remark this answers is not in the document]
krasi-georgiev: correct.

head.go
```diff
@@ -520,15 +530,13 @@ func (h *Head) Init(minValidTime int64) error {
...
+		}
 		if err == nil {
 			continue
 		}
-		level.Warn(h.logger).Log("msg", "encountered WAL error, attempting repair", "err", err)
-		h.metrics.walCorruptionsTotal.Inc()
-		if err := h.wal.Repair(err); err != nil {
-			return errors.Wrap(err, "repair corrupted WAL")
-		}
+		return err
```
codesome: If I am not wrong, if we were able to repair here before, we would continue to loading next segment. But with this, if we encounter corruption, we return will repair it in `DB.Open()`, but if that was a success, we miss the further segments as we don't load them.

db.go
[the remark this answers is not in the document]
krasi-georgiev: updated

db.go
krasi-georgiev: added a lock

db_test.go
```diff
@@ -2191,3 +2191,94 @@ func TestBlockRanges(t *testing.T) {
...
+			testutil.Equals(t, expBlock.MaxTime(), blocks[i].MaxTime(), "block MaxTime mismatch")
+		}
+
+		q, err := dbReadOnly.Querier(math.MinInt64, math.MaxInt64)
+		testutil.Ok(t, err)
+		readOnlySeries := query(t, q, matchAll)
+		readOnlyDBSize, err := testutil.DirSize(dbDir)
+		testutil.Ok(t, err)
+
+		testutil.Assert(t, len(readOnlySeries) > 0, "querier should return some series")
```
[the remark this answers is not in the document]
krasi-georgiev: updated.
actually it is 2 series as `genSeries(1, 1, m.MinTime, m.MaxTime)` always creates the same series.

CHANGELOG.md
```diff
@@ -1,5 +1,5 @@
 ## master / unreleased
-
+ - [FEATURE] Added `DBReadOnly` to allow opening a database in read only mode.
```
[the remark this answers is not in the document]
krasi-georgiev: done

db.go
krasi-georgiev: I refactored a bit to simplify this.

head.go
krasi-georgiev: after a short discussion we figured that this is a bug in the current code and this PR accidentally fixes it.

All WAL files after the corrupted one should be ignored to avoid adding misleading data.

head.go
krasi-georgiev: confirmed that typecasting on nil interface doesn't panic
https://play.golang.org/p/plfVNzBEclS

db.go
```diff
@@ -246,6 +246,150 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
...
+// A goroutine must not handle more than one open Querier.
+func (db *DBReadOnly) Querier(mint, maxt int64) (Querier, error) {
+	w, err := wal.NewReadOnly(db.logger, nil, filepath.Join(db.dir, "wal"))
+	if err != nil {
+		return nil, err
+	}
+	head, err := NewHead(nil, db.logger, w, 1)
```
codesome: We should close this head somewhere.

db_test.go
```diff
@@ -2191,3 +2191,97 @@ func TestBlockRanges(t *testing.T) {
...
+
+		q, err := dbReadOnly.Querier(math.MinInt64, math.MaxInt64)
+		testutil.Ok(t, err)
+		readOnlySeries := query(t, q, matchAll)
+		readOnlyDBSize, err := testutil.DirSize(dbDir)
+		testutil.Ok(t, err)
+
+		testutil.Equals(t, expSeriesCount, len(readOnlySeries), "total series mismatch")
+		testutil.Equals(t, expSeries, readOnlySeries, "series mismatch")
+		testutil.Equals(t, expDbSize, readOnlyDBSize, "after all read operations the db size should remain the same")
```
codesome: I was wondering if comparing size would assure equality. A good test for equality would be hashing all the file data + file/dir names as we walk the directory and compare them.
Hashing _could_ be an overkill, so this check should be fine.

krasi-georgiev: This is actually a good idea and I added it. 
It indeed found another problem - when we open a block it rewrites the mets.json file

I will look into some way to refactor it to make Opening a block a read only operation.

krasi-georgiev:
@codesome 	@bwplotka 	this is ready for a review. now that opening a block doesn't rewrite the meta file.

testutil/directory.go
```diff
@@ -143,3 +146,41 @@ func DirSize(path string) (int64, error) {
...
 }
+
+// DirHash returns a hash of all files attribites and their content within a directory.
+func DirHash(path string) ([]byte, error) {
+	hash := md5.New()
+	err := filepath.Walk(path, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if !info.IsDir() {
```
bwplotka: You can continue early and safe one indent for long body below

krasi-georgiev: yep , updated

krasi-georgiev:
ping @bwplotka

testutil/directory.go
```diff
@@ -143,3 +146,43 @@ func DirSize(path string) (int64, error) {
 	})
 	return size, err
 }
+
+// DirHash returns a hash of all files attribites and their content within a directory.
+func DirHash(path string) ([]byte, error) {
+	hash := md5.New()
```
codesome: We can change it to something better, like `sha256`.

cmd/tsdb/main.go
```diff
@@ -73,18 +70,37 @@ func main() {
...
 		wb.run()
 	case listCmd.FullCommand():
-		db, err := tsdb.Open(*listPath, nil, nil, &safeDBOptions)
+		db, err := tsdb.OpenDBReadOnly(*listPath, nil)
+		if err != nil {
+			exitWithError(err)
+		}
+		defer func() {
+			err := db.Close()
+			if err != nil {
```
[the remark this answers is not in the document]
krasi-georgiev: updated

db.go
```diff
@@ -250,6 +250,161 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
...
+	blocks := make([]*Block, len(blocksReaders))
+	for i, b := range blocksReaders {
+		b, ok := b.(*Block)
+		if !ok {
+			return nil, errors.New("unable to convert a read only block to a normal block")
+		}
+		blocks[i] = b
+	}
+
+	var head *Head
```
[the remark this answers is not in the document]
krasi-georgiev: good point. updated

db.go
```diff
@@ -250,6 +250,161 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
...
+		if err != nil {
+			return nil, err
+		}
+		// Set the min valid time for the ingested wal samples
+		// to be no lower than the maxt of the last block.
+		if err := head.Init(maxBlockTime); err != nil {
+			return nil, errors.Wrap(err, "read WAL")
+		}
+		// Set the wal to nil to disable all wal operations.
+		// This is mainly to avoid blocking when closing the head.
```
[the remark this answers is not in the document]
krasi-georgiev: Closing the head might lead to some write operations and we want to avoid this.

db.go
```diff
@@ -250,6 +250,161 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
...
+	for i, b := range loadable {
+		blockClosers[i] = b
+		blockReaders[i] = b
+	}
+	db.closers = blockClosers
+
+	return blockReaders, nil
+}
+
+// Close all db blocks to release the locks.
```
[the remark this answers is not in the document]
krasi-georgiev: to close all block readers. 
updated to make it more clear.

head.go
```diff
@@ -1101,6 +1110,17 @@ func (h *Head) chunksRange(mint, maxt int64) *headChunkReader {
 	return &headChunkReader{head: h, mint: mint, maxt: maxt}
 }
 
+// Meta returns meta information about the head.
+func (h *Head) Meta() BlockMeta {
+	var id [16]byte
+	copy(id[:], "______head______")
```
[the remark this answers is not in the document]
krasi-georgiev: > If the id is going to remain same, why not have it as a global const? Would avoid allocation and copying every time it is called.

just double checked and can't see how to save a string/byte array into a const.


>PS: I am not sure what Meta for Head should actually mean as only MinTime and MaxTime are relevant here. I would be careful about adding Meta() to BlockReader, for example if I try to use BlockStats from the Meta() somewhere and if it happens to be Head, I would get wrong numbers.

yes I was thinking the same, but really want to use the BlockReader interface instead of blocks directly to make sure it doesn't provide write operations and the `analyzeBlock` func requires the meta do display the block UUID. 
I think if we run into some sort of problems can refactor later.

testutil/directory.go
krasi-georgiev: updated

cmd/tsdb/main.go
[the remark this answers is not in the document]
codesome: https://github.com/prometheus/tsdb/pull/588/files#diff-beaca4df72bb224a9c7b294c84b2b493R599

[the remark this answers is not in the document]
krasi-georgiev: sorry! updated

head.go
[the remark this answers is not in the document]
codesome: > just checked and can't see how to save a string/byte array into a const.

I think that's fine then.

> I think if we run into some sort of problems can refactor later.

Hope we don't run into any problem!

krasi-georgiev:
ping @codesome 	@bwplotka 	for a review.

testutil/directory.go
```diff
@@ -143,3 +146,43 @@ func DirSize(path string) (int64, error) {
 	})
 	return size, err
 }
+
+// DirHash returns a hash of all files attribites and their content within a directory.
+func DirHash(path string) ([]byte, error) {
```
codesome: As this is a `testutil`, it can take `testing.TB` object and call `Ok(...)` method for errors instead of `err != nil` checks and returning of error.

krasi-georgiev: updated, also updated the DirSize func to make it consistent.

gouthamve:
Can you hold off one more day? I plan to review this tomorrow but if I don't do anything tomorrow, feel free to merge, I don't forsee any issues :)

krasi-georgiev:
@gouthamve 	thanks! waiting for the review.

bwplotka:
Reviewing now (:

db.go
```diff
@@ -250,6 +250,164 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
...
+	dir     string
+	closers []io.Closer
+	mtx     sync.Mutex
+	closed  bool
+}
+
+// OpenDBReadOnly opens DB in the given directory for read only operations.
+func OpenDBReadOnly(dir string, l log.Logger) (*DBReadOnly, error) {
+	if _, err := os.Stat(dir); err != nil {
+		return nil, err
```
bwplotka: some wrap would be nice? to show it is from `read only`?

db.go
```diff
@@ -250,6 +250,164 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
...
+	}
+
+	if l == nil {
+		l = log.NewNopLogger()
+	}
+
+	db := &DBReadOnly{
```
bwplotka: you could inline this, to minimize complexity, but not a blocker

db.go
```diff
@@ -250,6 +250,164 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
...
+// Querier loads the wal and returns a new querier over the data partition for the given time range.
+// A goroutine must not handle more than one open Querier.
+func (db *DBReadOnly) Querier(mint, maxt int64) (Querier, error) {
+	blocksReaders, err := db.Blocks()
+	if err != nil {
+		return nil, err
+	}
+	blocks := make([]*Block, len(blocksReaders))
+	for i, b := range blocksReaders {
+		b, ok := b.(*Block)
```
bwplotka: Why we do that?

db.go
```diff
@@ -250,6 +250,164 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
...
+		// Set the wal to nil to disable all wal operations.
+		// This is mainly to avoid blocking when closing the head.
+		head.wal = nil
+
+		db.mtx.Lock()
+		db.closers = append(db.closers, head)
+		db.mtx.Unlock()
+	}
+
+	dbWritable := &DB{
```
bwplotka: I don't think it makes sense for read only instance to create write instance inside to read something.. It does not sound right.

I think I am fine for now to unblock read only as no write method is exposed, but can we at least put some `TODO` or comment that we could improve it? There should be `dbReader` IMO that allows only Querying and separe writer layer on top that allows both read and write. 

Reason: we should isolate those things.

db.go
bwplotka: I mean why have interface if we literally have one implementation we require everywhere... ?

db.go
```diff
@@ -250,6 +250,164 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
...
+			delete(corrupted, b.ULID)
+		}
+	}
+	if len(corrupted) > 0 {
+		for _, b := range loadable {
+			if err := b.Close(); err != nil {
+				level.Warn(db.logger).Log("msg", "closing a block", err)
+			}
+		}
+		return nil, fmt.Errorf("unexpected corrupted block:%v", corrupted)
```
bwplotka: I think we used `pkg/errors` everywhere so it should be `errors.Errorf` here?

db.go
bwplotka: And it's even us who implement `db.Blocks` so can we return `[]*Block` in `db.Blocks` then?

db.go
```diff
@@ -250,6 +250,164 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
...
+}
+
+// Close all block readers.
+func (db *DBReadOnly) Close() error {
+	if db.closed {
+		return errors.New("db already closed")
+	}
+	var merr tsdb_errors.MultiError
+
+	db.mtx.Lock()
```
bwplotka: Do we need lock here? Do we expect readers to be created after Close? This will then create readers which will be never closed. 

I would suggest moving `db.closed` under this lock as well or make `db.closed` a channel and respect it on `db.Block`  as well.

db.go
```diff
@@ -250,6 +250,164 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
...
+	db := &DBReadOnly{
+		logger: l,
+		dir:    dir,
+	}
+
+	return db, nil
+}
+
+// Querier loads the wal and returns a new querier over the data partition for the given time range.
+// A goroutine must not handle more than one open Querier.
```
bwplotka: What do you mean? It is not thread safe? You can only invoke Querier once from single `DbReadOnly`? Can reword comment for that then?

db.go
```diff
@@ -250,6 +250,164 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
...
+	for _, b := range loadable {
+		blockMetas = append(blockMetas, b.Meta())
+	}
+	if overlaps := OverlappingBlocks(blockMetas); len(overlaps) > 0 {
+		level.Warn(db.logger).Log("msg", "overlapping blocks found during opening", "detail", overlaps.String())
+	}
+
+	db.mtx.Lock()
+	defer db.mtx.Unlock()
+	// Close all previously open readers and add the new ones to the cache.
```
bwplotka: Really? Even the pending ones from multiple invokes of `DbReadOnly.Querier` that still might be pending? I

EDIT: Ok, I can see the comment about Querier that it can be only one pending right? If that is the case this is fine, but we literally don't need lock here - just making `db.closed` to channel and checking if it's closed

db.go
bwplotka: I don't think we need lock at all then (only channel for `db.closed`) if I understand the

`// A goroutine must not handle more than one open Querier.` right

db.go
```diff
@@ -514,13 +672,13 @@ func (db *DB) compact() (err error) {
 	return nil
 }
 
-func (db *DB) getBlock(id ulid.ULID) (*Block, bool) {
-	for _, b := range db.blocks {
+func getBlock(loaded []*Block, id ulid.ULID) *Block {
```
bwplotka: I preferred previous approach more, because it gives the guarantee that block cannot be nil. It's more explicit. But not a blocker.

head.go
```diff
@@ -698,6 +702,21 @@ func (h *rangeHead) MaxTime() int64 {
 	return h.maxt
 }
 
+func (h *rangeHead) NumSeries() uint64 {
+	return h.head.NumSeries()
+}
+
+func (h *rangeHead) Meta() BlockMeta {
+	return BlockMeta{
+		MinTime: h.MinTime(),
```
bwplotka: wonder if anyone is using those `MinTime` / `MaxTime` methods apart here?

head.go
```diff
@@ -1022,9 +1041,10 @@ func (h *Head) gc() {
 	seriesRemoved := len(deleted)
 
 	h.metrics.seriesRemoved.Add(float64(seriesRemoved))
-	h.metrics.series.Sub(float64(seriesRemoved))
 	h.metrics.chunksRemoved.Add(float64(chunksRemoved))
 	h.metrics.chunks.Sub(float64(chunksRemoved))
+	// Ref: https://golang.org/pkg/sync/atomic/#AddUint64
```
bwplotka:
```suggestion
	// Using AddUint64 to substract series removed. See: https://golang.org/pkg/sync/atomic/#AddUint64
```

head.go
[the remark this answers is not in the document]
bwplotka: > PS: I am not sure what Meta for Head should actually mean as only MinTime and MaxTime are relevant here. I would be careful about adding Meta() to BlockReader, for example if I try to use BlockStats from the Meta() somewhere and if it happens to be Head, I would get wrong numbers.

Yea this is bit weird. Maybe splitting `Meta()` to actually `BlockStats` which would be current snapshot of things might be nicer? not sure, worth to comment on `Meta` interface method for sure. Head is already wrapped with `BlockReader` even though it is not a block so it's fine I think.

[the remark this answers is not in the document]
bwplotka: Also please move this string to constant at least `const headBlockID = "______head______"` I think it might nice to keep in single place. Although we don't depend on it now so we can refactor later.

block.go
```diff
@@ -138,11 +138,8 @@ type BlockReader interface {
 	// Tombstones returns a TombstoneReader over the block's deleted data.
 	Tombstones() (TombstoneReader, error)
 
-	// MinTime returns the min time of the block.
-	MinTime() int64
-
-	// MaxTime returns the max time of the block.
-	MaxTime() int64
+	// Meta provides meta information about the block reader.
+	Meta() BlockMeta
```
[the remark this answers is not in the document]
krasi-georgiev: topic for another PR :)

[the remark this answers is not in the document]
bwplotka: :+1:

db.go
krasi-georgiev: Great idea for the channel use.
I refactored using channels, much better, thanks. 

I removed the mutex. This makes is unsafe i called in separate go routines, but I think the API is clear enough to make it obvious that these should not be called in separate go routines.

db.go
```diff
@@ -250,6 +250,171 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
...
+		blockReaders[i] = b
+	}
+	db.closers = blockClosers
+
+	return blockReaders, nil
+}
+
+// Close all block readers.
+func (db *DBReadOnly) Close() error {
+	select {
```
bwplotka: :+1:

db.go
bwplotka: `func (db *DB) getBlock(id ulid.ULID) (*Block, bool) {`

bwplotka: similar to `map` when you do `_, ok := someMap[...]`

head.go
bwplotka: What do you mean? 

We literally removed this from interface, so if TSDB code base does not use `MinTime`  from outside then it's wrong to leave them as there is no way anyone can use them outside. (:

head.go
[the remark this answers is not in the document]
bwplotka: resolved without addressing my questions.

db.go
krasi-georgiev: Thanks. I added a comment. Different wording, but I think it is quite clear.
Will open an issue once this is merged as the idea might change along the way.

db.go
krasi-georgiev: Yep, now looking at the code I agree, it would be better with the old approach and more familiar to the behavior of maps.

db.go
krasi-georgiev: hm , I am still unsure what you mean here.

db.go
krasi-georgiev: Yep, only a single Querier and no concurrency for now. Added a comment.

head.go
krasi-georgiev: you can still use it when you open a writable database like:
db:=tsdb.Open(...)
blocks:=db.Blocks()
blocks[0].MinTime()


or you mean something different?

head.go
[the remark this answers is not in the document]
krasi-georgiev: >resolved without addressing my questions.

not sure what happened, wasn't my intention to leave unanswered.

> Yea this is bit weird. Maybe splitting Meta() to actually BlockStats which would be current snapshot of things might be nicer?

that is a good idea, but how can we do it without braking users and the current meta json implementation?

> not sure, worth to comment on Meta interface method for sure. 

What would this comment say and why/how would it be useful?



>Also please move this string to constant at least const headBlockID = "______head______" I think it might nice to keep in single place. Although we don't depend on it now so we can refactor later.

Why not do it when/if we actually need to use it in more than a single place?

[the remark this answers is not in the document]
bwplotka: > Why not do it when/if we actually need to use it in more than a single place?

Sure, just here this is quite vague ID. Putting in constant would allow to name it `headBlockID` so it would maybe more obvious (not a blocker)

> that is a good idea, but how can we do it without braking users and the current meta json implementation?

I think we misunderstood -> I meant rather having `MinTime` and `MaxTime` and `BlockStats` as separate methods. Anyway, same problem, so ignore me.

> What would this comment say and why/how would it be useful?
> > not sure, worth to comment on Meta interface method for sure.

That for head block this meta can give "dynamic" result for both max time and stats.

db.go
bwplotka: :+1: Not sure either, I think it might mean similar one querier = one goroutine, but no sure 100% (: 
That will work for now.

db.go
bwplotka: :+1:

head.go
bwplotka: I can't because you just removed if from interface, righ? ;p

head.go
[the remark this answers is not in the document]
krasi-georgiev: updated.

db.go
krasi-georgiev: updated

head.go
krasi-georgiev: We removed the MinTime/MaxTime for the BlockReader interface, but you can still use it on blocks.

cmd/tsdb/main.go
```diff
@@ -61,30 +68,43 @@ func main() {
...
 	case listCmd.FullCommand():
-		db, err := tsdb.Open(*listPath, nil, nil, &safeDBOptions)
+		db, err := tsdb.OpenDBReadOnly(*listPath, nil)
 		if err != nil {
-			exitWithError(err)
+			return err
 		}
-		printBlocks(db.Blocks(), listCmdHumanReadable)
+		defer func() {
+			err = db.Close()
```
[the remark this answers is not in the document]
krasi-georgiev: updated

head.go
krasi-georgiev: or you mean remove it for the blocks as well?

head.go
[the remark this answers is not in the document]
bwplotka: What exactly? (:

bwplotka:
Can see comment https://github.com/prometheus/tsdb/pull/588/files#diff-0ced3454d447442a96a77b33ebe87300R1131 :+1: 

Thanks

PR #594, opened 2019-04-26
Correctly handle empty labels.

brian-brazil:
Currently a time series with empty labels is not treated the same
as one with missing labels. Currently this can only come from
ALERTS&ALERT_FOR_STATE so it's unlikely anyone has actually hit it.

@beorn7 FYI

### DIFF
--- a/labels/labels.go
+++ b/labels/labels.go
@@ -103,6 +103,23 @@ func (ls Labels) Map() map[string]string {
 	return m
 }
 
+// WithoutEmpty returns the labelset without empty labels.
+// May return the same labelset.
+func (ls Labels) WithoutEmpty() Labels {
+	for _, v := range ls {
+		if v.Value == "" {
+			els := make(Labels, 0, len(ls)-1)
+			for _, v := range ls {
+				if v.Value != "" {
+					els = append(els, v)
+				}
+			}
+			return els
+		}
+	}
+	return ls
+}
+
 // New returns a sorted Labels from the given labels.
 // The caller has to guarantee that all label names are unique.
 func New(ls ...Label) Labels {
@@ -119,7 +136,9 @@ func New(ls ...Label) Labels {
 func FromMap(m map[string]string) Labels {
 	l := make(Labels, 0, len(m))
 	for k, v := range m {
-		l = append(l, Label{Name: k, Value: v})
+		if v != "" {
+			l = append(l, Label{Name: k, Value: v})
+		}
 	}
 	sort.Sort(l)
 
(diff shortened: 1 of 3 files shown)


labels/labels.go
```diff
@@ -103,6 +103,23 @@ func (ls Labels) Map() map[string]string {
 	return m
 }
 
+// WithoutEmpty returns the labelset without empty labels.
+// May return the same the same labelset..
```
simonpasquier: "the same the same"?

brian-brazil: the same

labels/labels.go
```diff
@@ -103,6 +103,23 @@ func (ls Labels) Map() map[string]string {
 	return m
 }
 
+// WithoutEmpty returns the labelset without empty labels.
+// May return the same labelset.
+func (ls Labels) WithoutEmpty() Labels {
```
bwplotka: Maybe me personal opinion, but I would say just function is better here as it has to be done only once at certain code logic.

Having this as Labels method indicates you need to do it a lot... but we don't want unnecceary run through all labels if not needed.

labels/labels.go
```diff
@@ -103,6 +103,23 @@ func (ls Labels) Map() map[string]string {
...
+	for _, v := range ls {
+		if v.Value == "" {
+			els := make(Labels, 0, len(ls))
```
bwplotka: We might be able to do it in less indent (more verbose), but not a blocker.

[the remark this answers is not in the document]
bwplotka: Agree, but potentially does not matter as there might be even more empty values in set.

brian-brazil:
> Does the querier follow the same semantics?

Yes, we go to a good bit of effort to do so. These are the general semantics of PromQL.

labels/labels.go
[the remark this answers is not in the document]
beorn7: Even then we waste one less element if using `len(ls)-1`. That's what I mean by “strictly better”.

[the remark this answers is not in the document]
beorn7: Not that it matters a lot, of course. Performance-wise, that is.

In my way of viewing things, doing `len(ls)-1` is actually a readability improvement. As the code reader, I start to think about possible reasons why `len(ls)` was used and why my understanding of the code (which suggests that at most `len(ls)-1` will be needed) might be wrong.

If performance wouldn't matter at all, I would not pre-allocate anything for maximum readability.
In this case, I think `len(ls)-1` > no pre-allocation > `len(ls)`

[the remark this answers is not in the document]
bwplotka: Thanks for clarification, we are definitely on the same page @beorn7  (:

labels/labels.go
brian-brazil: If this turns out to be a bigger problem than I think, we'll need to sprinkle this across the codebase.

labels/labels.go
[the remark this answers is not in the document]
brian-brazil: Done

labels/labels.go
beorn7: I don't feel strongly either way.

brian-brazil:
Are there any other comments on this?

PR #601, opened 2019-05-09
Export the current segment index as a metic.

cstyan:
Add a metric for the current segment index that TSDB will use for the WAL. Useful for looking into WAL watcher/tailer issues.

### DIFF
--- a/wal/wal.go
+++ b/wal/wal.go
@@ -171,6 +171,7 @@ type WAL struct {
 	pageCompletions prometheus.Counter
 	truncateFail    prometheus.Counter
 	truncateTotal   prometheus.Counter
+	currentSegment  prometheus.Gauge
 }
 
 // New returns a new WAL over the given directory.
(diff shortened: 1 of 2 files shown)


wal/wal.go
```diff
@@ -218,8 +219,12 @@ func NewSize(logger log.Logger, reg prometheus.Registerer, dir string, segmentSi
 		Name: "prometheus_tsdb_wal_truncations_total",
 		Help: "Total number of WAL truncations attempted.",
 	})
+	w.currentSegment = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "prometheus_tsdb_wal_current_segment",
```
brian-brazil: segment_current would be more common

cstyan: fixed

krasi-georgiev:
The current index is always the one with the highest number, why do you need a metric for that?

brian-brazil:
:+1:

krasi-georgiev:
and if we decide to add this please don't forget to add a test for that metric.

brian-brazil:
We don't usually add tests for metrics.

krasi-georgiev:
I prefer if we have these as we had  few broken ones here and in Prometheus with some refactoring.

krasi-georgiev:
look at the wal custom size test on how to easily trigger creation of a new wal file so you can check the metric is incremented

cstyan:
@krasi-georgiev 

> The current segment is always the one with the highest number, why do you need a metric for that?

The WAL watcher code exports a metric for what segment it's currently reading from. There's a WAL watcher per remote write queue. If there's an issue with one of the remote write queue's it's useful to know whether the segment that's currently being read is the segment TSDB is currently writing to, or how far behind the queue is.

cstyan:
@krasi-georgiev any objections to merging?

krasi-georgiev:
no objection, if there is a real use case for it.

PR #467, opened 2018-12-10
Add e2e  benchmarks for head and block Queries

krasi-georgiev:
closes #240 
~closes #278~ already closed in https://github.com/prometheus/tsdb/pull/572

combined and refactored the above PRs.

will post results from the benchmarks shortly.

Signed-off-by: Krasi Georgiev <kgeorgie@redhat.com>

PR #602, opened 2019-05-14
Optimize queries using regex matchers for set lookups

naivewong:
fixes https://github.com/prometheus/prometheus/issues/2651

### DIFF
--- a/block_test.go
+++ b/block_test.go
@@ -21,6 +21,7 @@ import (
 	"math/rand"
 	"os"
 	"path/filepath"
+	"strconv"
 	"testing"
 
 	"github.com/go-kit/kit/log"
@@ -184,6 +185,11 @@ func createBlock(tb testing.TB, dir string, series []Series) string {
 	return filepath.Join(dir, ulid.String())
 }
 
+const (
+	defaultLabelName  = "labelName"
+	defaultLabelValue = "labelValue"
+)
+
 // genSeries generates series with a given number of labels and values.
 func genSeries(totalSeries, labelCount int, mint, maxt int64) []Series {
 	if totalSeries == 0 || labelCount == 0 {
@@ -193,8 +199,9 @@ func genSeries(totalSeries, labelCount int, mint, maxt int64) []Series {
 	series := make([]Series, totalSeries)
 	for i := 0; i < totalSeries; i++ {
 		lbls := make(map[string]string, labelCount)
-		for len(lbls) < labelCount {
-			lbls[randString()] = randString()
+		lbls[defaultLabelName] = strconv.Itoa(i)
+		for j := 1; len(lbls) < labelCount; j++ {
+			lbls[defaultLabelName+strconv.Itoa(j)] = defaultLabelValue + strconv.Itoa(j)
 		}
 		samples := make([]tsdbutil.Sample, 0, maxt-mint+1)
 		for t := mint; t <= maxt; t++ {
@@ -223,32 +230,4 @@ func populateSeries(lbls []map[string]string, mint, maxt int64) []Series {
 		series = append(series, newSeries(lbl, samples))
 	}
 	return series
-}
-
-const letterBytes = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
-const (
-	letterIdxBits = 6                    // 6 bits to represent a letter index
-	letterIdxMask = 1<<letterIdxBits - 1 // All 1-bits, as many as letterIdxBits
-	letterIdxMax  = 63 / letterIdxBits   // # of letter indices fitting in 63 bits
-)
-
-// randString generates random string.
-func randString() string {
-	maxLength := int32(50)
-	length := rand.Int31n(maxLength)
-	b := make([]byte, length+1)
-	// A rand.Int63() generates 63 random bits, enough for letterIdxMax characters!
-	for i, cache, remain := length, rand.Int63(), letterIdxMax; i >= 0; {
-		if remain == 0 {
-			cache, remain = rand.Int63(), letterIdxMax
-		}
-		if idx := int(cache & letterIdxMask); idx < len(letterBytes) {
-			b[i] = letterBytes[idx]
-			i--
-		}
-		cache >>= letterIdxBits
-		remain--
-	}
-
-	return string(b)
 }
--- a/querier.go
+++ b/querier.go
@@ -266,6 +266,350 @@ func (q *blockQuerier) Close() error {
 	return merr.Err()
 }
 
+func addStrToLevel(level []*strings.Builder, strs []string, idx int) []*strings.Builder {
+	levelSize := len(level) - idx
+	for j := idx + levelSize; j < idx + levelSize * len(strs); j ++ {
+		level = append(level, &strings.Builder{})
+		level[j].WriteString(level[j % levelSize + idx].String())
+	}
+	for i, s := range strs {
+		for j := idx + i * levelSize; j < idx + (i + 1) * levelSize; j ++ {
+			level[j].WriteString(s)
+		}
+	}
+	return level
+}
+
+func combineLevels(left []*strings.Builder, right []*strings.Builder, idx int) []*strings.Builder {
+	levelSize := len(left) - idx
+	for j := idx + levelSize; j < idx + levelSize * len(right); j ++ {
+		left = append(left, &strings.Builder{})
+		left[j].WriteString(left[idx + j % levelSize].String())
+	}
+	for i, s := range right {
+		for j := idx + i * levelSize; j < idx + (i + 1) * levelSize; j ++ {
+			left[j].WriteString(s.String())
+		}
+	}
+	return left
+}
+
+func handleEnd(pattern string, idx int) int {
+	escaped := false
+	for idx < len(pattern) {
+		switch c := pattern[idx]; {
+		case c == '$':
+			break
+		case c == '\\':
+			escaped = true
+			break
+		case c == 'B' && escaped:
+			escaped = false
+			break
+		default:
+
+		}
+		idx += 1
+	}
+	if idx == len(pattern) {
+		return idx
+	} else {
+		return -1
+	}
+}
+
+// Return empty array if not found.
+func findSetMatches(pattern string) []string {
+	// To detect the character '\'.
+	escaped := false
+	insideBrackets := false
+	insideBraces := false
+	combinePending := false
+	matches := []string{}
+	// This is to handle the nested parentheses.
+	levels := [][]*strings.Builder{[]*strings.Builder{&strings.Builder{}}}
+	barIdx := map[int]int{}
+	bracket := []string{}
+	i := 0
+	// Handle the beginning part of the pattern.
+Loop:
+	for i < len(pattern) {
+		switch c := pattern[i]; {
+		case c == '^':
+			break
+		case c == '\\':
+			escaped = true
+			break
+		case c == 'b' && escaped:
+			escaped = false
+			break
+		default:
+			break Loop
+		}
+		i += 1
+	}
+	// Handle the middle part of the pattern.
+	for i < len(pattern) {
+		if escaped {
+			switch c := pattern[i]; {
+			case c == 'B':
+				escaped = false
+				i = handleEnd(pattern, i + 1)
+				if i == -1 {
+					return []string{}
+				}
+				break
+			case c == '(' || c == ')' || c == '[' || c == '{' || c == '$' || c == '*' || c == '+' || c == '.' || c == '?' || c == '\\' || c == '^' || c == '|':
+				escaped = false
+				if idx, ok := barIdx[len(levels) - 1]; ok {
+					levels[len(levels) - 1] = addStrToLevel(levels[len(levels) - 1], []string{string(c)}, idx)
+				} else {
+					levels[len(levels) - 1] = addStrToLevel(levels[len(levels) - 1], []string{string(c)}, 0)
+				}
+			default:
+				return []string{}
+			}
+		} else if insideBrackets {
+			switch c := pattern[i]; {
+			case c == '^':
+				return []string{}
+			case c == ']':
+				if !(i + 1 < len(pattern) && pattern[i + 1] == '{') {
+					if idx, ok := barIdx[len(levels) - 1]; ok {
+						levels[len(levels) - 1] = addStrToLevel(levels[len(levels) - 1], bracket, idx)
+					} else {
+						levels[len(levels) - 1] = addStrToLevel(levels[len(levels) - 1], bracket, 0)
+					}
+					bracket = bracket[:0]
+				}
+				insideBrackets = false
+				break
+			case c == '-':
+				for ch := pattern[i - 1] + 1; ch <= pattern[i + 1]; ch ++ {
+					bracket = append(bracket, string(ch))
+				}
+				i += 1
+				break;
+			default:
+				bracket = append(bracket, string(pattern[i]))
+				break
+			}
+		} else if insideBraces {
+			start := 0
+			end := 0
+			left := false
+			comma := false
+			right := false
+			j := i
+			for pattern[j] != '}' {
+				if '0' <= pattern[j] && pattern[j] <= '9' {
+					if !comma {
+						left = true
+						start = start * 10 + int(pattern[j] - '0')
+					} else {
+						right = true
+						end = end * 10 + int(pattern[j] - '0')
+					}
+				} else if pattern[j] == ',' {
+					comma = true
+				} else {
+					return []string{}
+				}
+				j += 1
+			}
+			if !left {
+				if len(bracket) > 0 {
+					if idx, ok := barIdx[len(levels) - 1]; ok {
+						levels[len(levels) - 1] = addStrToLevel(levels[len(levels) - 1], bracket, idx)
+					} else {
+						levels[len(levels) - 1] = addStrToLevel(levels[len(levels) - 1], bracket, 0)
+					}
+					bracket = bracket[:0]
+				}
+				if idx, ok := barIdx[len(levels) - 1]; ok {
+					levels[len(levels) - 1] = addStrToLevel(levels[len(levels) - 1], []string{pattern[i - 1: j + 1]}, idx)
+				} else {
+					levels[len(levels) - 1] = addStrToLevel(levels[len(levels) - 1], []string{pattern[i - 1: j + 1]}, 0)
+				}
+			} else if comma {
+				if !right {
+					return []string{}
+				} else {
+					// Need further discussion.
+					l := 1
+					if len(bracket) > l {
+						l = len(bracket)
+					}
+					if combinePending && len(levels[len(levels) - 1]) > l {
+						l = len(levels[len(levels) - 1])
+					}
+					addStrs := make([]string, 0, (end - start + 1) * l)
+					if combinePending {
+						idx, ok := barIdx[len(levels) - 1]
+						if !ok {
+							idx = 0
+						}
+						for k := start; k <= end; k ++ {
+							for _, s := range levels[len(levels) - 1] {
+								addStrs = append(addStrs, strings.Repeat(s.String(), k))
+							}
+						}
+						levels[len(levels) - 1] = levels[len(levels) - 1][: idx + 1]
+						levels[len(levels) - 1][idx].Reset()
+						// Delete the bar index on the top level.
+						if _, ok = barIdx[len(levels) - 1]; ok {
+							delete(barIdx, len(levels) - 1)
+						}
+						levels = levels[: len(levels) - 1]
+						combinePending = false
+					} else if len(bracket) > 0 {
+						for k := start; k <= end; k ++ {
+							for _, s := range bracket {
+								addStrs = append(addStrs, strings.Repeat(s, k))
+							}
+						}
+						bracket = bracket[:0]
+					} else {
+						for k := start; k <= end; k ++ {
+							addStrs = append(addStrs, strings.Repeat(string(pattern[i - 2]), k))
+						}
+					}
+					if idx, ok := barIdx[len(levels) - 1]; ok {
+						levels[len(levels) - 1] = addStrToLevel(levels[len(levels) - 1], addStrs, idx)
+					} else {
+						levels[len(levels) - 1] = addStrToLevel(levels[len(levels) - 1], addStrs, 0)
+					}
+				}
+			} else {
+				l := 1
+				if len(bracket) > l {
+					l = len(bracket)
+				}
+				if combinePending && len(levels[len(levels) - 1]) > l {
+					l = len(levels[len(levels) - 1])
+				}
+				addStrs := make([]string, 0, l)
+				if combinePending {
+					idx, ok := barIdx[len(levels) - 1]
+					if !ok {
+						idx = 0
+					}
+					for _, s := range levels[len(levels) - 1] {
+						addStrs = append(addStrs, strings.Repeat(s.String(), start))
+					}
+					levels[len(levels) - 1] = levels[len(levels) - 1][: idx + 1]
+					levels[len(levels) - 1][idx].Reset()
+					// Delete the bar index on the top level.
+					if _, ok = barIdx[len(levels) - 1]; ok {
+						delete(barIdx, len(levels) - 1)
+					}
+					levels = levels[: len(levels) - 1]
+					combinePending = false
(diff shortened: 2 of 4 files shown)


brian-brazil:
You need to check for and exclude the wrapper we put around the regex the user specifies.

naivewong:
Oh yes, I forgot to exclude the wrapper. But sorry what did you mean by "check for"?

naivewong:
Is that we exclude the wrapper only if we find it?

brian-brazil:
If the wrapper is missing, treat it as a regex.

brian-brazil:
That looks about right. Can you add it to the postings benchmark too?

I'd be interested to know if there's a point where it's better to use the regex rather than the set.

gouthamve:
> I'd be interested to know if there's a point where it's better to use the regex rather than the set.

I don't think that'll ever be the case, as with regex, first we'll get all the matching label values (the set) and doing what the set matcher does anyways. With set, we'll be avoiding a step.

brian-brazil:
At some point, us doing parsing of the strings might be slower than the regex. Would be good to check one way or the other.

bboreham:
What's the motivation for only allowing escaping of special characters?  What if I want to match `\t` or `\n` in a label value?

codesome:
Adding to @bboreham's comment, isn't it enough to just include any character that is escaped rather than checking for special character or '\\'?

naivewong:
Benchmark
```
BenchmarkSetMatcher/SetMatch,nSeries=15,pattern="^(?:1|2|3)$"-8         	  300000	      4579 ns/op	    3425 B/op	      59 allocs/op
BenchmarkSetMatcher/RegexMatch,nSeries=15,pattern="1|2|3"-8             	  200000	      6349 ns/op	    2836 B/op	      64 allocs/op
BenchmarkSetMatcher/SetMatch,nSeries=15,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-8         	  200000	      8249 ns/op	    5443 B/op	      96 allocs/op
BenchmarkSetMatcher/RegexMatch,nSeries=15,pattern="1|2|3|4|5|6|7|8|9|10"-8             	  200000	      8844 ns/op	    4208 B/op	      81 allocs/op
BenchmarkSetMatcher/SetMatch,nSeries=200,pattern="^(?:1|2|3)$"-8                       	  300000	      5011 ns/op	    3425 B/op	      59 allocs/op
BenchmarkSetMatcher/RegexMatch,nSeries=200,pattern="1|2|3"-8                           	   20000	     70441 ns/op	   31928 B/op	     545 allocs/op
BenchmarkSetMatcher/SetMatch,nSeries=200,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-8        	  200000	      9755 ns/op	    5546 B/op	      96 allocs/op
BenchmarkSetMatcher/RegexMatch,nSeries=200,pattern="1|2|3|4|5|6|7|8|9|10"-8            	   20000	     83923 ns/op	   35351 B/op	     644 allocs/op
```

querier_test.go
```diff
@@ -1691,6 +1692,118 @@ func BenchmarkQuerySeek(b *testing.B) {
...
+	cases := []struct {
+		numBlocks                   int
+		numSeries                   int
+		numSamplesPerSeriesPerBlock int
+		setPattern                  string
+		regexPattern                string
+	}{
+		{
+			numBlocks:                   1,
+			numSeries:                   15,
```
codesome: As this is a benchmark, let's keep this number a little higher (a few thousands). Also a higher `numSamplesPerSeriesPerBlock`.

querier_test.go
```diff
@@ -1691,6 +1692,118 @@ func BenchmarkQuerySeek(b *testing.B) {
...
+			numSamplesPerSeriesPerBlock: 10,
+			setPattern:                  "^(?:1|2|3)$",
+			regexPattern:                "1|2|3",
```
codesome: Instead of checking this, we can check `setPattern` itself for before and after this PR and get the diff. This would give the actual impact of the changes in this PR.

querier_test.go
```diff
@@ -1691,6 +1692,118 @@ func BenchmarkQuerySeek(b *testing.B) {
...
+		)
+		for i := int64(0); i < int64(c.numBlocks); i++ {
+			mint := i * int64(c.numSamplesPerSeriesPerBlock)
+			maxt := mint + int64(c.numSamplesPerSeriesPerBlock) - 1
+			if len(prefilledLabels) == 0 {
+				generatedSeries = make([]Series, c.numSeries)
+				for i := 0; i < c.numSeries; i++ {
+					lbls := make(map[string]string, 10)
+					// The first label pair is {"test", "i"} which is for benchmarking set matcher.
+					lbls["test"] = strconv.Itoa(i)
```
codesome: Can we repeat the same `i` for more than 1 series? If we are having a higher number of series now, we can maybe have this as `i%someNumber` (50-100 is what is in my mind).

querier_test.go
naivewong: This case I set 15 is because I want to check if the case "I'd be interested to know if there's a point where it's better to use the regex rather than the set." exists.

querier_test.go
naivewong: You mean delete `regexPattern`?

querier_test.go
codesome: Makes sense. You can also add benchmark case for higher number of series to know how it will impact real settings.

querier_test.go
codesome: You can keep both, but test it before and after this PR.

codesome: Just for reference: https://godoc.org/golang.org/x/tools/cmd/benchcmp

naivewong:
```
benchmark                                                                   old ns/op     new ns/op     delta
BenchmarkSetMatcher/nSeries=15,pattern="^(?:1|2|3)$"-8                      6512          4461          -31.50%
BenchmarkSetMatcher/nSeries=15,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-8       10420         8136          -21.92%
BenchmarkSetMatcher/nSeries=1000,pattern="^(?:1|2|3)$"-8                    346047        234610        -32.20%
BenchmarkSetMatcher/nSeries=1000,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-8     429347        273327        -36.34%

benchmark                                                                   old allocs     new allocs     delta
BenchmarkSetMatcher/nSeries=15,pattern="^(?:1|2|3)$"-8                      72             59             -18.06%
BenchmarkSetMatcher/nSeries=15,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-8       95             96             +1.05%
BenchmarkSetMatcher/nSeries=1000,pattern="^(?:1|2|3)$"-8                    2290           1330           -41.92%
BenchmarkSetMatcher/nSeries=1000,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-8     2655           1975           -25.61%

benchmark                                                                   old bytes     new bytes     delta
BenchmarkSetMatcher/nSeries=15,pattern="^(?:1|2|3)$"-8                      3657          3425          -6.34%
BenchmarkSetMatcher/nSeries=15,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-8       5478          5445          -0.60%
BenchmarkSetMatcher/nSeries=1000,pattern="^(?:1|2|3)$"-8                    105927        90080         -14.96%
BenchmarkSetMatcher/nSeries=1000,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-8     130824        118979        -9.05%
```

brian-brazil:
What about 100k to 1M series? It's the bigger data sizes where the wins are, what you've tested are already fast.

gouthamve:
Also a case where there is only one value for the label, but 15 options in the set matcher. Maybe that will tell us if the set matcher is always faster or not.

naivewong:
```
benchmark                                                                               old ns/op     new ns/op     delta
BenchmarkSetMatcher/nSeries=1,nBlocks=1,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-8          5937          8446          +42.26%
BenchmarkSetMatcher/nSeries=15,nBlocks=1,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-8         12308         8428          -31.52%
BenchmarkSetMatcher/nSeries=15,nBlocks=1,pattern="^(?:1|2|3)$"-8                        8016          4653          -41.95%
BenchmarkSetMatcher/nSeries=1000,nBlocks=20,pattern="^(?:1|2|3)$"-8                     522961        207214        -60.38%
BenchmarkSetMatcher/nSeries=1000,nBlocks=20,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-8      651767        303631        -53.41%
BenchmarkSetMatcher/nSeries=100000,nBlocks=1,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-8     27351         13267         -51.49%
BenchmarkSetMatcher/nSeries=500000,nBlocks=1,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-8     42813         31066         -27.44%

benchmark                                                                               old allocs     new allocs     delta
BenchmarkSetMatcher/nSeries=1,nBlocks=1,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-8          50             96             +92.00%
BenchmarkSetMatcher/nSeries=15,nBlocks=1,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-8         95             96             +1.05%
BenchmarkSetMatcher/nSeries=15,nBlocks=1,pattern="^(?:1|2|3)$"-8                        72             59             -18.06%
BenchmarkSetMatcher/nSeries=1000,nBlocks=20,pattern="^(?:1|2|3)$"-8                     3290           1330           -59.57%
BenchmarkSetMatcher/nSeries=1000,nBlocks=20,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-8      3655           1975           -45.96%
BenchmarkSetMatcher/nSeries=100000,nBlocks=1,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-8     180            96             -46.67%
BenchmarkSetMatcher/nSeries=500000,nBlocks=1,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-8     180            96             -46.67%

benchmark                                                                               old bytes     new bytes     delta
BenchmarkSetMatcher/nSeries=1,nBlocks=1,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-8          3527          5209          +47.69%
BenchmarkSetMatcher/nSeries=15,nBlocks=1,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-8         5476          5436          -0.73%
BenchmarkSetMatcher/nSeries=15,nBlocks=1,pattern="^(?:1|2|3)$"-8                        3657          3425          -6.34%
BenchmarkSetMatcher/nSeries=1000,nBlocks=20,pattern="^(?:1|2|3)$"-8                     121933        90080         -26.12%
BenchmarkSetMatcher/nSeries=1000,nBlocks=20,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-8      146805        118994        -18.94%
BenchmarkSetMatcher/nSeries=100000,nBlocks=1,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-8     6744          5353          -20.63%
BenchmarkSetMatcher/nSeries=500000,nBlocks=1,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-8     6745          5352          -20.65%
```

gouthamve:
This looks good so far!

Can you also add a `nBlocks=10` version for `BenchmarkSetMatcher/nSeries=500000,nBlocks=1,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-8`?

naivewong:
Unable to show you the results because the previous one was already the limit of my MBP.

krasi-georgiev:
pinged you on irc to give you access to a big machine where you can run these test quickly.

querier_test.go
```diff
@@ -1691,6 +1692,188 @@ func BenchmarkQuerySeek(b *testing.B) {
...
+		)
+		for i := int64(0); i < int64(c.numBlocks); i++ {
+			mint := i * int64(c.numSamplesPerSeriesPerBlock)
+			maxt := mint + int64(c.numSamplesPerSeriesPerBlock) - 1
+			if len(prefilledLabels) == 0 {
+				generatedSeries = make([]Series, c.numSeries)
+				for i := 0; i < c.numSeries; i++ {
+					lbls := make(map[string]string, 10)
+					// The first label pair is {"test", "i%100"} which is for benchmarking set matcher.
+					lbls["test"] = strconv.Itoa(i % 100)
```
brian-brazil: This is limiting cardinality to 100, this is what we want to test at 1M

naivewong:
```
benchmark                                                                                                     old ns/op      new ns/op     delta
BenchmarkSetMatcher/nSeries=1,nBlocks=1,cardinality=100,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-48               7499           11422         +52.31%
BenchmarkSetMatcher/nSeries=15,nBlocks=1,cardinality=100,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-48              15715          12522         -20.32%
BenchmarkSetMatcher/nSeries=15,nBlocks=1,cardinality=100,pattern="^(?:1|2|3)$"-48                             9715           6789          -30.12%
BenchmarkSetMatcher/nSeries=1000,nBlocks=20,cardinality=100,pattern="^(?:1|2|3)$"-48                          731441         369395        -49.50%
BenchmarkSetMatcher/nSeries=1000,nBlocks=20,cardinality=100,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-48           951066         466389        -50.96%
BenchmarkSetMatcher/nSeries=100000,nBlocks=1,cardinality=100000,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-48       24648679       16168         -99.93%
BenchmarkSetMatcher/nSeries=500000,nBlocks=1,cardinality=500000,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-48       127604669      12485         -99.99%
BenchmarkSetMatcher/nSeries=500000,nBlocks=10,cardinality=500000,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-48      1222647955     206051        -99.98%
BenchmarkSetMatcher/nSeries=1000000,nBlocks=1,cardinality=1000000,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-48     260974613      15219         -99.99%

benchmark                                                                                                     old allocs     new allocs     delta
BenchmarkSetMatcher/nSeries=1,nBlocks=1,cardinality=100,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-48               50             96             +92.00%
BenchmarkSetMatcher/nSeries=15,nBlocks=1,cardinality=100,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-48              95             96             +1.05%
BenchmarkSetMatcher/nSeries=15,nBlocks=1,cardinality=100,pattern="^(?:1|2|3)$"-48                             72             59             -18.06%
BenchmarkSetMatcher/nSeries=1000,nBlocks=20,cardinality=100,pattern="^(?:1|2|3)$"-48                          3290           1330           -59.57%
BenchmarkSetMatcher/nSeries=1000,nBlocks=20,cardinality=100,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-48           3655           1975           -45.96%
BenchmarkSetMatcher/nSeries=100000,nBlocks=1,cardinality=100000,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-48       100080         96             -99.90%
BenchmarkSetMatcher/nSeries=500000,nBlocks=1,cardinality=500000,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-48       500080         96             -99.98%
BenchmarkSetMatcher/nSeries=500000,nBlocks=10,cardinality=500000,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-48      5000765        919            -99.98%
BenchmarkSetMatcher/nSeries=1000000,nBlocks=1,cardinality=1000000,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-48     1000080        96             -99.99%

benchmark                                                                                                     old bytes     new bytes     delta
BenchmarkSetMatcher/nSeries=1,nBlocks=1,cardinality=100,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-48               3542          5227          +47.57%
BenchmarkSetMatcher/nSeries=15,nBlocks=1,cardinality=100,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-48              5486          5438          -0.87%
BenchmarkSetMatcher/nSeries=15,nBlocks=1,cardinality=100,pattern="^(?:1|2|3)$"-48                             3663          3430          -6.36%
BenchmarkSetMatcher/nSeries=1000,nBlocks=20,cardinality=100,pattern="^(?:1|2|3)$"-48                          122103        90211         -26.12%
BenchmarkSetMatcher/nSeries=1000,nBlocks=20,cardinality=100,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-48           146976        119136        -18.94%
BenchmarkSetMatcher/nSeries=100000,nBlocks=1,cardinality=100000,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-48       1605359       5353          -99.67%
BenchmarkSetMatcher/nSeries=500000,nBlocks=1,cardinality=500000,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-48       8006583       5352          -99.93%
BenchmarkSetMatcher/nSeries=500000,nBlocks=10,cardinality=500000,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-48      80087445      54573         -99.93%
BenchmarkSetMatcher/nSeries=1000000,nBlocks=1,cardinality=1000000,pattern="^(?:1|2|3|4|5|6|7|8|9|10)$"-48     16007303      5352          -99.97%
```

querier_test.go
```diff
@@ -1691,6 +1692,205 @@ func BenchmarkQuerySeek(b *testing.B) {
...
+		var (
+			blocks          []*Block
+			prefilledLabels []map[string]string
+			generatedSeries []Series
+		)
+		for i := int64(0); i < int64(c.numBlocks); i++ {
+			mint := i * int64(c.numSamplesPerSeriesPerBlock)
+			maxt := mint + int64(c.numSamplesPerSeriesPerBlock) - 1
+			if len(prefilledLabels) == 0 {
+				generatedSeries = make([]Series, c.numSeries)
```
krasi-georgiev: can also look at the "genSeries" helper.

inside the genSeries we use a random string, but don't think we even need this and I already removed is part of [this PR](https://github.com/prometheus/tsdb/pull/467/files#diff-7ae027963734feb4c8722aa99f033363R207) so feel free to refactor it a bit if you can use it here as well.

brian-brazil:
Those results look more like it.

codesome:
Gains at higher cardinality are impressive!

querier.go
```diff
@@ -411,6 +476,16 @@ func inversePostingsForMatcher(ix IndexReader, m labels.Matcher) (index.Postings
 	return index.Merge(rit...), nil
 }
 
+func postingsForSetMatcher(ix IndexReader, name string, matches []string) (index.Postings, error) {
+	var its []index.Postings
+	for _, match := range matches {
+		if it, err := ix.Postings(name, match); err == nil {
```
juliusv: Shouldn't `err` be returned when it's not `nil` instead of skipping/ignoring it?

querier.go
```diff
@@ -266,6 +267,62 @@ func (q *blockQuerier) Close() error {
...
+func init() {
+	for _, b := range []byte(`.+*?()|[]{}^$`) {
+		specialBytes[b%16] |= 1 << (b / 16)
+	}
+}
+
+func findSetMatches(pattern string) []string {
+	// Return empty matches if the wrapper from Prometheus is missing.
+	if len(pattern) < 6 || pattern[:4] != "^(?:" || pattern[len(pattern)-2:] != ")$" {
+		return []string{}
```
juliusv: nit: it's more usual in Go to return `nil` instead of explicitly empty slices (same below).

querier.go
```diff
@@ -266,6 +267,62 @@ func (q *blockQuerier) Close() error {
 	return merr.Err()
 }
 
+// Bitmap used by func special to check whether a character needs to be escaped.
+var specialBytes [16]byte
+
+// special reports whether byte b needs to be escaped.
+func special(b byte) bool {
```
juliusv: In the context of the whole file and package, a more specific function name than `special()` would be good. How about `isRegexMetaCharacter()` and `specialBytes` -> `regexMetaCharacterBytes`?

querier.go
```diff
@@ -266,6 +267,62 @@ func (q *blockQuerier) Close() error {
...
+				if pattern[i] == '|' {
+					sets = append(sets, &strings.Builder{})
+				} else {
+					return []string{}
+				}
+			} else if pattern[i] == '\\' {
+				escaped = true
+			} else {
+				sets[len(sets)-1].WriteByte(pattern[i])
+			}
```
juliusv: (minor personal opinion) IMO case differentiations like this are visually easier to follow using `switch`:

```go
switch {
case special(pattern[i]):
  // ...
case pattern[i] == '\\':
  // ...
default:
  // ...
}
```

brian-brazil:
Just had a thought there that we should have some unittests to cover when one of the values in the regex is the empty string.

querier.go
```diff
@@ -266,6 +267,62 @@ func (q *blockQuerier) Close() error {
...
+	// Return empty matches if the wrapper from Prometheus is missing.
+	if len(pattern) < 6 || pattern[:4] != "^(?:" || pattern[len(pattern)-2:] != ")$" {
+		return nil
+	}
+	escaped := false
+	sets := []*strings.Builder{&strings.Builder{}}
+	for i := 4; i < len(pattern)-2; i++ {
+		if escaped {
+			switch {
+			case isRegexMetaCharacter(pattern[i]):
```
bboreham: Why is this check necessary?
Why not just allow every escaped character? (Or perhaps disallow `\0...` since that’s complicated).

naivewong: @brian-brazil Should I include the cases above?

brian-brazil: Sure, though check what Go does for escaping characters that don't need to be escaped.

bboreham: I'm sorry, I was thinking about the wrong language.
The set of escapes we care about here is documented at https://github.com/google/re2/wiki/Syntax and is painfully complicated.
There are lots of escapes we don't want to accept, like `\g`, `\p`, `\x`, so you do need something like what you have.

naivewong: I have rechecked, the special characters you meant like `\n`, `\a` are actually already included in the `else {}` part of `findSetMatches`. What I detect here are the special characters like `\\.`, `\\+` in regexp, which means after I find `\\`, I determine if the next char is special.

block_test.go
```diff
@@ -193,8 +199,9 @@ func genSeries(totalSeries, labelCount int, mint, maxt int64) []Series {
 	series := make([]Series, totalSeries)
 	for i := 0; i < totalSeries; i++ {
 		lbls := make(map[string]string, labelCount)
-		for len(lbls) < labelCount {
-			lbls[randString()] = randString()
+		lbls[defaultLabelName] = strconv.Itoa(i)
+		for j := 1; len(lbls) < labelCount; j++ {
```
codesome: @krasi-georgiev I have lost touch with the codebase lately, does changing this functionality affect other places where it is used? Here the series are no longer random for every `genSeries` call, and it inclined more towards testing this particular PR.

querier.go
```diff
@@ -266,6 +267,62 @@ func (q *blockQuerier) Close() error {
 	return merr.Err()
 }
 
+// Bitmap used by func isRegexMetaCharacter to check whether a character needs to be escaped.
+var regexMetaCharacterBytes [16]byte
+
+// isRegexMetaCharacter reports whether byte b needs to be escaped.
+func isRegexMetaCharacter(b byte) bool {
+	return b < utf8.RuneSelf && regexMetaCharacterBytes[b%16]&(1<<(b/16)) != 0
```
codesome: Won't make a big difference, but why not.
```suggestion
	return b < utf8.RuneSelf && regexMetaCharacterBytes[b%16]&(1<<(b>>4)) != 0
```

brian-brazil: This code came from the Go source, probably best not to change it.

brian-brazil:
:+1:

block_test.go
krasi-georgiev: It is fine so far we don't need randomness just different label=value variations.
```
lbls[defaultLabelName+strconv.Itoa(j)] = defaultLabelValue + strconv.Itoa(j)
```

PR #605, opened 2019-05-16
Start a new WAL segement on head truncation.

brian-brazil:
This reduces disk space usage to not be a minimum of 3 128MB files
in small setups. This will possibly also help debug wal data issues,
by making things a bit more deterministic.

Signed-off-by: Brian Brazil <brian.brazil@robustperception.io>

### DIFF
--- a/head.go
+++ b/head.go
@@ -549,6 +549,12 @@ func (h *Head) Truncate(mint int64) (err error) {
 	if err != nil {
 		return errors.Wrap(err, "get segment range")
 	}
+	// Start a new segment, so low ingestion volume TSDB don't have more WAL than
+	// needed.
+	err = h.wal.NextSegment()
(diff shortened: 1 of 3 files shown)


krasi-georgiev:
this is a new behavior so not a bad idea to add a test.

 - Create enough blocks to run compaction.
 - Repeat few times and every time check that the wal files count is as expected even 
 when the wal size is below 128MB.

brian-brazil:
Added a test

krasi-georgiev:
>This reduces disk space usage to not be a minimum of 3 128MB files
in small setups. 

we added the custom wal size for such setups.

brian-brazil:
This avoids any special configuration being needed in such setups, and that one particular user had extremely low resources available.

PR #606, opened 2019-05-17
move WAL tailing code from Prometheus to TSDB WAL package

cstyan:
This is probably a bit of a pain to review, but moving the WAL watcher code from Prometheus to TSDB (into the WAL package) requires moving some code from the root tsdb package. Specifically the Watcher uses `RefSeries` and `RefSample`. Moving those requires a bunch of other code to be moved as well.

If anyone knows a nicer way to make the `RefSeries`/`RefSamples` change or has thoughts about the new package organization let me know, none of this is set in stone.

cc @tomwilkie @gouthamve


record/internal.go
```diff
@@ -0,0 +1,371 @@
...
+	"github.com/prometheus/tsdb/chunkenc"
+	"github.com/prometheus/tsdb/chunks"
+	"github.com/prometheus/tsdb/fileutil"
+	"github.com/prometheus/tsdb/labels"
+)
+
+var (
+	// ErrOutOfOrderSample is returned if an appended sample has a
+	// timestamp smaller than the most recent sample.
+	ErrOutOfOrderSample = errors.New("out of order sample")
```
csmarchbanks: These error messages do not seem related to records, but more appenders/series. Sometimes ErrNotFound is used for missing files. Is there a better place for them?

record/internal.go
```diff
@@ -0,0 +1,371 @@
...
+
+	// ErrAmendSample is returned if an appended sample has the same timestamp
+	// as the most recent sample but a different value.
+	ErrAmendSample = errors.New("amending sample")
+)
+
+// The table gets initialized with sync.Once but may still cause a race
+// with any other use of the crc32 package anywhere. Thus we initialize it
+// before.
+var castagnoliTable *crc32.Table
```
csmarchbanks: I don't think this is used.

record/internal.go
```diff
@@ -0,0 +1,371 @@
...
+type RefSample struct {
+	Ref    uint64
+	T      int64
+	V      float64
+	Series *MemSeries
+}
+
+// MemSeries is the in-memory representation of a series. None of its methods
+// are goroutine safe and it is the caller's responsibility to lock it.
+type MemSeries struct {
```
csmarchbanks: This only seems to be used by the root package. Is there a reason it had to move that I am missing?

record/tombstones.go
```diff
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package tsdb
+package record
```
csmarchbanks: Does it make sense for the whole tombstone code to move? Or just the `Stones` type?

record/internal.go
cstyan: it's used by `NewCRC32`, which is used by tombstones in the record package

record/internal.go
cstyan: `ErrNotFound` could live in `fileutil` I think. `ErrOutOfOrderSample` sample is only used by `MemSeries`. Maybe the package `record` should just have a different name?

record/internal.go
cstyan: It's also used by the `RefSample` type just above it in this same package. I thought it made sense to keep them both in the same package.

record/tombstones.go
cstyan: Up for debate. For the WAL tailing code to work I think only `Stone` would need to move. I don't feel strongly about moving all of `tombstones.go` but in general I think the amount of code in the root package is problematic.

record/internal.go
csmarchbanks: Hmmm... How important do you think it is to have a pointer to `MemSeries` on `RefSample`? Having the two tied together really only makes sense in the head, whereas we now use RefSample in a number of places.

record/internal.go
csmarchbanks: A different name could help, not totally sure what the best name is though.

csmarchbanks:
Thanks for all the responses, they helped me get my head around all this! As you said... the amount of code in the root package can be problematic to wrap your head around...

My main concern with the current state of this PR is that the `record` package appears to be random types with no obvious relation. I don't think will help with code being more isolated/easier to wrap your head around.

After thinking about this for a couple hours:
1. I am pretty against `memSeries` being moved away from head.
   * Perhaps have a slice of `*memSeries` next to the slice of `RefSample`s in `headAppender` instead of having `RefSample` to depend on `memSeries`?
1. Should the record code (and `RefSample`/`RefSeries`) live in `wal`? Are records used anywhere else? Otherwise I like them living in their own `record` package.
1. Perhaps tombstones should be in their own package? I don't feel too strongly about this either, just having `Stone` live wherever `RefSample` does would be fine by me too.

I am definitely curious what you and the maintainers think!

codesome:
As I am not much familiar with WAL tailing yet, I am still wrapping my head about what the `record` package should be. But I agree with some points which @csmarchbanks said
* `memSeries` looks closely tied with `Head` and would be good to keep it there.
* If there are only a few things used from `tombstones.go` in the `record`, then duplicate only those constructs and not the entire file. Else `tombstones` as a separate package would make more sense than being in inside `record`. (I support the former)

I will take a better look before the weekend hits.

cstyan:
I've moved the tombstones code to it's own package. Unfortunately keeping `memSeries` in head and have the records package `RefSample` not contain `memSeries` is going to be pretty messy.

cstyan:
cc @csmarchbanks @codesome

csmarchbanks:
I like the tombstone move. I think you could remove the `memSeries` from `RefSample` without it becoming too messy by keeping two slices in sync. What do you think of
[memSeries.patch.txt](https://github.com/prometheus/tsdb/files/3273314/memSeries.patch.txt)? You would have to apply to master, not your branch.

krasi-georgiev:
What is the problem you are trying to solve with this PR?

cstyan:
@krasi-georgiev the code for tailing/watching the WAL currently lives in the remote write package in prometheus/prometheus. We think it makes more sense for it to live in the WAL package, which is in tsdb repo right now. Unfortunately the WAL tailing code relies on a couple of types that live in the root `tsdb` package, which would end up causing some circular dependency issues.

krasi-georgiev:
hm , I see.
In that case need to spend some more time to look into this.

record/internal.go
csmarchbanks: I think `ErrOutOfOrderSample` and `ErrAmendSample` can go back into the base package now correct since they only relate to appenders?

I am still not sure what to do with `ErrNotFound`, it seems to be used both for files not being found and for series or chunks not being found. Neither of which are related to records.

record/internal.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// RefSample is a timestamp/value pair associated with a reference to a series.
+type RefSample struct {
+	Ref uint64
+	T   int64
+	V   float64
+}
+
+func rangeForTimestamp(t int64, width int64) (maxt int64) {
```
csmarchbanks: Are `rangeForTimestamp`, `computeChunkEndTime`, or `RenameFile` used anywhere?

wal/wal_watcher.go
```diff
@@ -0,0 +1,573 @@
...
+}
+
+// NewWALWatcher creates a new WAL watcher for a given WriteTo.
+func NewWALWatcher(logger log.Logger, reg prometheus.Registerer, name string, writer writeTo, walDir string) *WALWatcher {
+	if logger == nil {
+		logger = log.NewNopLogger()
+	}
+	if reg != nil {
+		// We can't use MustRegister because WALWatcher's are recreated on config changes within Prometheus.
+		reg.Register(watcherRecordsRead)
```
csmarchbanks: These seem to be registered both here and in `init`.

Also, should these metrics be scoped to a `WALWatcher` similar to the `LiveReader` change recently done?

wal/checkpoint.go
```diff
@@ -146,14 +147,14 @@ func Checkpoint(w *wal.WAL, from, to int, keep func(id uint64) bool, mint int64)
...
 	var (
-		series  []RefSeries
-		samples []RefSample
-		tstones []Stone
-		dec     RecordDecoder
-		enc     RecordEncoder
+		series  []record.RefSeries
+		samples []record.RefSample
+		tstones []tombstones.Stone
+		dec     record.RecordDecoder
```
csmarchbanks: Now that `RecordDecoder` and `RecordEncoder` are in their own package these stutter a bit. Perhaps rename to `record.Decoder` and `record.Encoder`?

wal/checkpoint.go
```diff
@@ -167,7 +168,7 @@ func Checkpoint(w *wal.WAL, from, to int, keep func(id uint64) bool, mint int64)
 		rec := r.Record()
 
 		switch dec.Type(rec) {
-		case RecordSeries:
+		case record.RecordSeries:
```
csmarchbanks: These all stutter now as well. `record.Series`, `record.Samples`, and `record.Tombstones` would be nicer to read.

cstyan:
@csmarchbanks thanks for the review and for the sync.Pool suggestion https://github.com/prometheus/tsdb/pull/606#issuecomment-500541120

head.go
```diff
@@ -1605,25 +1630,25 @@ func (s *memSeries) cut(mint int64) *memChunk {
...
-		metas = append(metas, chunks.Meta{Chunk: chk.chunk, MinTime: chk.minTime, MaxTime: chk.maxTime})
+		metas = append(metas, chunks.Meta{Chunk: chk.Chunk, MinTime: chk.MinTime, MaxTime: chk.MaxTime})
 	}
 	return metas
 }
 
 // reset re-initialises all the variable in the memSeries except 'lset', 'ref',
-// and 'chunkRange', like how it would appear after 'newMemSeries(...)'.
-func (s *memSeries) reset() {
+// and 'chunkRange', like how it would appear after 'newmemSeries(...)'.
```
codesome: I think this was changed by mistake.

record/internal.go
```diff
@@ -0,0 +1,37 @@
+// Copyright 2017 The Prometheus Authors
```
codesome: The year should be when the file was created.

wal/watcher.go
```diff
@@ -0,0 +1,587 @@
+// Copyright 2018 The Prometheus Authors
```
codesome: Year is when the file was created.

wal/watcher_test.go
```diff
@@ -0,0 +1,504 @@
...
+	dir, err := ioutil.TempDir("", "readCheckpoint")
+	testutil.Ok(t, err)
+	defer os.RemoveAll(dir)
+
+	wdir := path.Join(dir, "wal")
+	err = os.Mkdir(wdir, 0777)
+	testutil.Ok(t, err)
+
+	enc := record.Encoder{}
+	w, err := NewSize(nil, prometheus.DefaultRegisterer, wdir, 128*pageSize, false)
```
csmarchbanks: Would you mind re-copying this file? I changed all the tests to run with compression enabled and disabled as part of the 2.11 release.

cstyan: yeah I need to recopy this and the watcher, quite a few small improvements have been made

head.go
```diff
@@ -916,10 +938,12 @@ func (a *headAppender) Commit() error {
 
 func (a *headAppender) Rollback() error {
 	a.head.metrics.activeAppenders.Dec()
-	for _, s := range a.samples {
-		s.series.Lock()
-		s.series.pendingCommit = false
-		s.series.Unlock()
+	var series *memSeries
```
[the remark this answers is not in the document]
cstyan: I am just a few lines below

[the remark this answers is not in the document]
csmarchbanks: D'oh! I am just blind, sorry.

wal/watcher.go
cstyan: this file was copied from the prometheus repo so I left he original copyright, I can change the year if you want

cstyan:
@csmarchbanks @codesome could use another review when you have some time

wal/watcher.go
```diff
@@ -0,0 +1,580 @@
...
+	return t.Unix()*1000 + int64(t.Nanosecond())/int64(time.Millisecond)
+}
+
+type WriteTo interface {
+	Append([]record.RefSample) bool
+	StoreSeries([]record.RefSeries, int)
+	SeriesReset(int)
+}
+
+type watcherMetrics struct {
```
csmarchbanks: I would prefer if this were `WatcherMetrics` to make it easier to pass around elsewhere.

wal/watcher.go
```diff
@@ -0,0 +1,580 @@
...
+		reg.Register(m.recordDecodeFails)
+		reg.Register(m.samplesSentPreTailing)
+		reg.Register(m.currentSegment)
+	}
+
+	return m
+}
+
+// NewWatcher creates a new WAL watcher for a given WriteTo.
+func NewWatcher(reg prometheus.Registerer, logger log.Logger, name string, writer WriteTo, walDir string) *Watcher {
```
csmarchbanks: This should take a `WatcherMetrics`, not a registerer, otherwise creating two watchers with the same registerer would panic.

cstyan: Both `NewWatcherMetrics` and `NewLiveReaderMetrics` call `reg.Register`, not `reg.MustRegister`. But I guess the real question here is whether it's reasonable for consumers of the WAL Watcher to have to create and pass the metrics struct rather than just passing a Registerer.

csmarchbanks: True, `reg.Register` won't panic, but it also won't register the new metrics, so we would be blind for those queues/after a reload.

I think it is reasonable to have a consumer create and pass the metrics struct, just like `NewLiveReader`.

wal/watcher.go
```diff
@@ -0,0 +1,580 @@
...
+	currentSegmentMetric    prometheus.Gauge
+
+	quit chan struct{}
+	done chan struct{}
+
+	// For testing, stop when we hit this segment.
+	maxSegment int
+}
+
+func NewWatcherMetrics(reg prometheus.Registerer) *watcherMetrics {
```
csmarchbanks: Should these take a consumer and provide an `Unregister` so that if a consumer disappears it can be removed from the metrics?

wal/reader_test.go
```diff
@@ -41,7 +41,7 @@ type reader interface {
 	Offset() int64
 }
 
-type record struct {
+type rec struct {
```
codesome: Why the change?

codesome:
Forgot to mention I haven't reviewed `wal/watcher.go` assuming that it was already reviewed when it was added to Prometheus.

wal/reader_test.go
cstyan: this package is importing `"github.com/prometheus/tsdb/record"`, so we get `record redeclared...` without renaming the struct type here

wal/watcher.go
cstyan: We can't `Unregister` as all the Watcher metrics are under the same metric but just `WithLabels`. Both QueueManager and Watcher `Stop` delete their metrics via `DeleteLabelValues`.

cstyan:
thanks @codesome, the only new addition to `watcher.go` here is how it's metrics are initialized

wal/watcher.go
csmarchbanks: :+1:

compact.go
```diff
@@ -607,7 +608,7 @@ func (c *LeveledCompactor) write(dest string, meta *BlockMeta, blocks ...BlockRe
 	}
 
 	// Create an empty tombstones file.
-	if _, err := writeTombstoneFile(c.logger, tmp, newMemTombstones()); err != nil {
+	if _, err := tombstones.WriteTombstoneFile(c.logger, tmp, tombstones.NewMemTombstones()); err != nil {
```
bwplotka: Are we sure `tombstones.WriteTombstoneFile` needs to be `WriteTombstoneFile` etc? Why not just `tombstones.WriteFile` if we have clear package name?

bwplotka: same for `NewMemTombstones`

compact.go
```diff
@@ -876,15 +877,15 @@ type compactionSeriesSet struct {
...
 
 	l         labels.Labels
 	c         []chunks.Meta
-	intervals Intervals
+	intervals tombstones.Intervals
 	err       error
 }
 
-func newCompactionSeriesSet(i IndexReader, c ChunkReader, t TombstoneReader, p index.Postings) *compactionSeriesSet {
+func newCompactionSeriesSet(i IndexReader, c ChunkReader, t tombstones.TombstoneReader, p index.Postings) *compactionSeriesSet {
```
bwplotka: ditto
```suggestion
func newCompactionSeriesSet(i IndexReader, c ChunkReader, t tombstones.Reader, p index.Postings) *compactionSeriesSet {
```

head.go
```diff
@@ -44,17 +46,17 @@ var (
...
 
-	// ErrAmendSample is returned if an appended sample has the same timestamp
-	// as the most recent sample but a different value.
-	ErrAmendSample = errors.New("amending sample")
-
 	// ErrOutOfBounds is returned if an appended sample is out of the
 	// writable time range.
 	ErrOutOfBounds = errors.New("out of bounds")
 
+	// ErrAmendSample is returned if an appended sample has the same timestamp
```
bwplotka: Why moving around?

head.go
```diff
@@ -789,19 +795,32 @@ func max(a, b int64) int64 {
...
+	return b.([]record.RefSample)
 }
 
-func (h *Head) putAppendBuffer(b []RefSample) {
+func (h *Head) putAppendBuffer(b []record.RefSample) {
 	//lint:ignore SA6002 safe to ignore and actually fixing it has some performance penalty.
 	h.appendPool.Put(b[:0])
 }
 
+func (h *Head) getSeriesBuffer() []*memSeries {
```
bwplotka: do we have to change this while just moving packages around?

tombstones/tombstones.go
```diff
@@ -30,7 +32,7 @@ import (
 	"github.com/prometheus/tsdb/fileutil"
 )
 
-const tombstoneFilename = "tombstones"
+const TombstoneFilename = "tombstones"
```
bwplotka: again, with `tombstones` package name you don't need `tombstones` prefix everywhere, we can simplify (:

tombstones/tombstones.go
```diff
@@ -54,8 +71,8 @@ type TombstoneReader interface {
 	Close() error
 }
 
-func writeTombstoneFile(logger log.Logger, dir string, tr TombstoneReader) (int64, error) {
-	path := filepath.Join(dir, tombstoneFilename)
+func WriteTombstoneFile(logger log.Logger, dir string, tr TombstoneReader) (int64, error) {
```
bwplotka: As it is exposed we need comment e.g for goDoc ): Same everywhere else for public functions we add.

PR #608, opened 2019-05-17
Always create a new clean segment when starting the WAL.

cstyan:
Also changes repair to read a single segment at a time as the offset detection is still problematic.

@brian-brazil 
cc @gouthamve

### DIFF
--- a/head.go
+++ b/head.go
@@ -499,7 +499,7 @@ func (h *Head) Init(minValidTime int64) error {
 		return errors.Wrap(err, "finding WAL segments")
 	}
 
-	// Backfill segments from the last checkpoint onwards.
+	// Backfill segments from the most recent checkpoint onwards.
 	for i := startFrom; i <= last; i++ {
 		s, err := wal.OpenReadSegment(wal.SegmentName(h.wal.Dir(), i))
 		if err != nil {
--- a/head.go
+++ b/head.go
@@ -14,6 +14,7 @@
 package tsdb
 
 import (
+	"fmt"
 	"math"
 	"runtime"
 	"sort"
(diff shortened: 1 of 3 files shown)


head.go
```diff
@@ -492,22 +493,29 @@ func (h *Head) Init(minValidTime int64) error {
...
-	err = h.loadWAL(wal.NewReader(sr))
-	sr.Close() // Close the reader so that if there was an error the repair can remove the corrupted file under Windows.
-	if err == nil {
-		return nil
-	}
-	level.Warn(h.logger).Log("msg", "encountered WAL error, attempting repair", "err", err)
-	h.metrics.walCorruptionsTotal.Inc()
-	if err := h.wal.Repair(err); err != nil {
-		return errors.Wrap(err, "repair corrupted WAL")
+	// Backfill segments from the last checkpoint onwards.
```
brian-brazil: most recent would be clearer than last, I misunderstood this as last segment

wal/wal.go
```diff
@@ -228,29 +228,25 @@ func NewSize(logger log.Logger, reg prometheus.Registerer, dir string, segmentSi
...
 	if err != nil {
 		return nil, errors.Wrap(err, "get segment range")
 	}
-	// Fresh dir, no segments yet.
-	if j == -1 {
-		segment, err := CreateSegment(w.dir, 0)
-		if err != nil {
-			return nil, err
-		}
+	// If some segments already exist create one with a higher index than the last segment.
```
brian-brazil: Wouldn't it still be j + 1 in the j == -1 case?

cstyan: `Segments()` returns -1 for first and last if the WAL dir exists but there are no segment files present. So in that case we won't change `writeSegmentIndex` and well create the segment `0`.

brian-brazil:
You've test failures.

cstyan:
@brian-brazil thanks for the review. Took me longer than I'd have liked to figure out what the cause of the test failures were on Windows.

brian-brazil:
Thanks!

krasi-georgiev:
seems like a rather big change and I would have appreciated to have looked at it before the merge.

head.go
```diff
@@ -492,22 +493,32 @@ func (h *Head) Init(minValidTime int64) error {
...
-	if err == nil {
-		return nil
-	}
-	level.Warn(h.logger).Log("msg", "encountered WAL error, attempting repair", "err", err)
-	h.metrics.walCorruptionsTotal.Inc()
-	if err := h.wal.Repair(err); err != nil {
-		return errors.Wrap(err, "repair corrupted WAL")
+	// Backfill segments from the most recent checkpoint onwards.
+	for i := startFrom; i <= last; i++ {
+		s, err := wal.OpenReadSegment(wal.SegmentName(h.wal.Dir(), i))
```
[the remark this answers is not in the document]
cstyan: @krasi-georgiev The `SegmentsRangeReader` is very problematic imo. The offset tracking causes all sorts of edge case issues. In this PR I'm not using it because we're now always creating a new segment, then doing the corruption check and `Repair`. The way offset tracking works Repair will always think the corruption is in the new segment rather than the previous segment if the corruption is at the very end of the segment.

PR #610, opened 2019-05-25
Fix `make bench`

mrasu:
When running `make bench`, it fails saying
```
>> completed stage=stopStorage duration=502.693441ms
no such flag -log.level
Makefile:27: recipe for target 'bench' failed
make: *** [bench] Error 1
```

Signed-off-by: mrasu <m.rasu.hitsuji@gmail.com>

### DIFF
--- a/cmd/tsdb/main.go
+++ b/cmd/tsdb/main.go
@@ -15,7 +15,6 @@ package main
 
 import (
 	"bufio"
-	"flag"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -106,9 +105,6 @@ func main() {
 			exitWithError(err)
 		}
 		dumpSamples(db, *dumpMinTime, *dumpMaxTime)
-	}
-	if err := flag.CommandLine.Set("log.level", "debug"); err != nil {
-		exitWithError(err)
 	}
 }
 


codesome:
I feel we don't need logging for all the commands as we will print only the desired stuff. I will wait for more comments on this before you would make any more changes.

krasi-georgiev:
yeah logging seems unnecessary here.

@mrasu Is there any specific reason why you want to add logging?

mrasu:
Sorry, I thought I didn't change whether we log or not.

I misunderstood passing `nil` to `tsdb.Open` creates default logger.
But seeing again, it creates `NewNopLogger`.
https://github.com/prometheus/tsdb/blob/master/db.go#L249

I will back to `nil` except `bench`.

But about `bench`, I'm still thinking there was a logger. Is it correct?
https://github.com/prometheus/tsdb/pull/610/files#diff-beaca4df72bb224a9c7b294c84b2b493R132

krasi-georgiev:
not sure why it was removed, but unless someone has a good use case, don't see why would we re-add.

mrasu:
Do you mean below doesn't log anything?
https://github.com/prometheus/tsdb/pull/610/files#diff-beaca4df72bb224a9c7b294c84b2b493L147

Then, I will change this PR to just remove these three lines :smiley: 
https://github.com/prometheus/tsdb/pull/610/files#diff-beaca4df72bb224a9c7b294c84b2b493L110

krasi-georgiev:
yep go ahead. If we actually need a logger somewhere will revisit.

mrasu:
changed!

krasi-georgiev:
Thanks

PR #614, opened 2019-05-29
update copyright headers

krasi-georgiev:

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -1,5 +1,4 @@
-// Copyright 2016 The etcd Authors
-//
+// Copyright 2017 The Prometheus Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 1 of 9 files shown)


gouthamve:
I think we took this code from the etcd codebase. Not sure if changing the headers is the right thing to do here.

krasi-georgiev:
good point, 
maybe @juliusv can comment on that?

juliusv:
(or is keeping things the way they are fine?)

swinslow:
Thanks @juliusv -- existing third-party copyright notices should be retained and should not be removed. 

If additional Prometheus content has been added to the files, then it should be fine to add an additional "Prometheus authors" notice -- but in any event the existing etcd statement should remain there too.

juliusv:
@swinslow Thanks! :) I take it that everything should remain as is then, unless people have added stuff to them.

swinslow:
Yes, exactly  :)  Thanks!

swinslow:
One other thought -- if you're concerned that people in the future might think the "etcd" notice is a mistake and try to change it: you might consider moving files like this into a "third_party/etcd/" directory or something similar. That way it's clearer that they originated from a third party upstream project. Not critical here, since this is also Apache-2.0-licensed code, but just a thought.

krasi-georgiev:
thanks for the info 
lets keep things as they are.

PR #616, opened 2019-05-31
Improvement on postings intersection

naivewong:
Tree-like recursive methods in the current implementation of Intersect() can cause unnecessary Seek().

**An example**
Suppose there are 4 postings [1,2,3,4,5], [3,4, 5], [5,6,7,8], [5, 8].
Calling func Intersect(), p1 and p2 will form an intersectPostings, p3 and p4 will form an intersectPostings. The steps are as follows.
1. Next() for ip1
1.1 Next() for p1 → 1
1.2 Seek(1) for p2 → 3
1.3 Seek(3) for p1 → 3
2. Seek(3) for ip2
2.1 Seek(3) for p3 → 5
2.2 Seek(3) for p4 → 5
3. Seek(5) for ip1
3.1 Seek(5) for p1 → 5
3.2 Seek(5) for p2 → 5
4. Next() for ip1

There are unnecessary Seek(3).
If not use recursive func Intersect(), instead manage all together in an array and Seek() the max value each time. The steps are as follows.
1. Next() for p1, p2, p3, p4 → max=5
2. Seek(5) for p1 → 5
3. Seek(5) for p2 → 5
4. Next() for p1, p2, p3, p4 → finish

**Benchmark**
```
benchmark                       old ns/op     new ns/op     delta
BenchmarkIntersect/case_1-8     199           100           -49.75%
BenchmarkIntersect/case_2-8     195           95.6          -50.97%

benchmark                       old allocs     new allocs     delta
BenchmarkIntersect/case_1-8     4              2              -50.00%
BenchmarkIntersect/case_2-8     4              2              -50.00%

benchmark                       old bytes     new bytes     delta
BenchmarkIntersect/case_1-8     208           96            -53.85%
BenchmarkIntersect/case_2-8     208           96            -53.85%
```

### DIFF
--- a/index/postings.go
+++ b/index/postings.go
@@ -303,6 +303,11 @@ func Intersect(its ...Postings) Postings {
 	if len(its) == 1 {
 		return its[0]
 	}
+	for _, p := range its {
+		if p == EmptyPostings() {
+			return EmptyPostings()
+		}
+	}
 
 	return newIntersectPostings(its...)
 }
@@ -320,21 +325,19 @@ func (it *intersectPostings) At() uint64 {
 	return it.cur
 }
 
-func (it *intersectPostings) doNext(id uint64) bool {
+func (it *intersectPostings) doNext() bool {
+Loop:
 	for {
-		find := true
 		for _, p := range it.arr {
 			if !p.Seek(it.cur) {
 				return false
 			}
 			if p.At() > it.cur {
 				it.cur = p.At()
-				find = false
+				continue Loop
 			}
 		}
-		if find {
-			return true
-		}
+		return true
 	}
 }
 
@@ -347,19 +350,12 @@ func (it *intersectPostings) Next() bool {
 			it.cur = p.At()
 		}
 	}
-	return it.doNext(it.cur)
+	return it.doNext()
 }
 
 func (it *intersectPostings) Seek(id uint64) bool {
-	for _, p := range it.arr {
-		if !p.Seek(id) {
-			return false
-		}
-		if p.At() > it.cur {
-			it.cur = p.At()
-		}
-	}
-	return it.doNext(it.cur)
+	it.cur = id
+	return it.doNext()
 }
 
 func (it *intersectPostings) Err() error {
--- a/index/postings.go
+++ b/index/postings.go
@@ -304,23 +304,16 @@ func Intersect(its ...Postings) Postings {
 		return its[0]
 	}
 
-	l := len(its) / 2
-	a := Intersect(its[:l]...)
-	b := Intersect(its[l:]...)
-
-	if a == EmptyPostings() || b == EmptyPostings() {
-		return EmptyPostings()
-	}
-	return newIntersectPostings(a, b)
+	return newIntersectPostings(its...)
 }
 
 type intersectPostings struct {
-	a, b Postings
-	cur  uint64
+	arr []Postings
+	cur uint64
 }
 
-func newIntersectPostings(a, b Postings) *intersectPostings {
-	return &intersectPostings{a: a, b: b}
+func newIntersectPostings(its ...Postings) *intersectPostings {
+	return &intersectPostings{arr: its}
 }
 
 func (it *intersectPostings) At() uint64 {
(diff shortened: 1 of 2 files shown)


index/postings.go
```diff
@@ -329,42 +327,53 @@ func (it *intersectPostings) At() uint64 {
...
+		for _, p := range it.arr {
+			if !p.Seek(it.cur) {
 				return false
 			}
-			id = it.a.At()
-			if vb != id {
-				continue
+			if p.At() > it.cur {
+				it.cur = p.At()
+				find = false
```
codesome: We can directly continue to the outer loop here and skip few `Seek` calls.

gouthamve: This doesn't change a lot of things according to the benchmarks, but good to have yes.

gouthamve:
After banging my head for a long time, this looks good! I think the recursion causes some overhead in the previous implementation and this new one is actually better in all cases I tested with.

index/postings.go
```diff
@@ -303,68 +303,79 @@ func Intersect(its ...Postings) Postings {
...
 
 func (it *intersectPostings) doNext(id uint64) bool {
+Loop:
 	for {
-		if !it.b.Seek(id) {
-			return false
-		}
-		if vb := it.b.At(); vb != id {
-			if !it.a.Seek(vb) {
+		find := true
```
codesome: This is not needed anymore after the latest changes.

index/postings_test.go
```diff
@@ -244,13 +244,45 @@ func BenchmarkIntersect(t *testing.B) {
...
+	for i := 7500000; i < 12500000; i++ {
+		b = append(b, uint64(i))
 	}
+	for i := 9000000; i < 20000000; i++ {
+		c = append(c, uint64(i))
+	}
+	for i := 10000000; i < 12000000; i++ {
+		d = append(d, uint64(i))
+	}
+	i1 = newListPostings(a...)
```
brian-brazil: Can we add a case with many many matchers?

naivewong:
```
benchmark                       old ns/op     new ns/op     delta
BenchmarkIntersect/case_1-8     210           95.5          -54.52%
BenchmarkIntersect/case_2-8     235           115           -51.06%
BenchmarkIntersect/case_3-8     7617750       385602        -94.94%

benchmark                       old allocs     new allocs     delta
BenchmarkIntersect/case_1-8     4              2              -50.00%
BenchmarkIntersect/case_2-8     4              2              -50.00%
BenchmarkIntersect/case_3-8     99999          1              -100.00%

benchmark                       old bytes     new bytes     delta
BenchmarkIntersect/case_1-8     208           96            -53.85%
BenchmarkIntersect/case_2-8     212           98            -53.77%
BenchmarkIntersect/case_3-8     4799952       32            -100.00%
```

index/postings_test.go
```diff
@@ -221,36 +221,96 @@ func TestMultiIntersect(t *testing.T) {
...
+		bench.ReportAllocs()
+		for i := 0; i < bench.N; i++ {
+			if _, err := ExpandPostings(Intersect(i1, i2, i3, i4)); err != nil {
+				bench.Fatal(err)
+			}
+		}
+	})
+
+	// Many matchers(k >> n).
+	t.Run("case 3", func(bench *testing.B) {
```
brian-brazil: The case names could be more descriptive.

index/postings_test.go
```diff
@@ -221,36 +221,96 @@ func TestMultiIntersect(t *testing.T) {
...
+		var its []Postings
+
+		// 100000 matchers(k=100000).
+		for i := 0; i < 100000; i++ {
+			m := make(map[uint64]struct{})
+			// n=100.
+			for count := 0; count < 100; count++ {
+				m[uint64(rand.Intn(500))] = struct{}{}
```
brian-brazil: This could skip past everything in one go, how about something more pathological like the same ids in each?

krasi-georgiev:
much easyer to understand now, Thanks @naivewong 

I found a way to simplify it a bit more.

```
func (it *intersectPostings) doNext(id uint64) bool {
Loop:
	for {
		for _, p := range it.arr {
			if !p.Seek(id) {
				return false
			}
			id = p.At()
			if p.At() > it.cur {
				it.cur = p.At()
				continue Loop
			}
		}
		return true
	}
}

func (it *intersectPostings) Seek(id uint64) bool {
	return it.doNext(id)
}
```

```
benchmark                              old ns/op     new ns/op     delta
BenchmarkIntersect/ManyPostings-8      18495452      649025        -96.49%
BenchmarkIntersect/LongPostings2-8     422           136           -67.77%
BenchmarkIntersect/LongPostings1-8     341           125           -63.34%

benchmark                              old allocs     new allocs     delta
BenchmarkIntersect/ManyPostings-8      99999          1              -100.00%
BenchmarkIntersect/LongPostings1-8     4              2              -50.00%
BenchmarkIntersect/LongPostings2-8     4              2              -50.00%

benchmark                              old bytes     new bytes     delta
BenchmarkIntersect/ManyPostings-8      4799963       32            -100.00%
BenchmarkIntersect/LongPostings2-8     216           98            -54.63%
BenchmarkIntersect/LongPostings1-8     208           96            -53.85%
```

PR #617, opened 2019-05-31
Limit number of series in Head

codesome:
By default, there is no limit.

We need this in https://github.com/cortexproject/cortex/pull/1427, but this limit might also help in other use cases.

TODO:
- [ ] Handle returning of `nil` series from `getOrCreateWithID(...)`.

/cc @tomwilkie

### DIFF
--- a/db.go
+++ b/db.go
@@ -51,6 +51,7 @@ var DefaultOptions = &Options{
 	BlockRanges:            ExponentialBlockRanges(int64(2*time.Hour)/1e6, 3, 5),
 	NoLockfile:             false,
 	AllowOverlappingBlocks: false,
+	MaxHeadSeries:          -1,
 }
 
 // Options of the DB storage.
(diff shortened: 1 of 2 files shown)


brian-brazil:
I suspect this will panic when this is hit, which is basically the correct action for such a limit - I don't think this makes sense for Prometheus as all we can do without panicing is stop all scraping and rule evaluations which isn't very useful. This also doesn't consider GC.

krasi-georgiev:
@codesome can you explain what is the use case as the Cortex PR is quite large and don't want to read through  all code changes there.

codesome:
@krasi-georgiev Cortex can limit the number of series, so while trying to use prometheus tsdb in cortex, we wanted some similar functionality.

On some more thought, this limit might not be needed in tsdb. I will revisit this later if this has some good use-case for Prometheus.

PR #618, opened 2019-05-31
Don't error out Snapshot for missing tombstones

codesome:
Closes https://github.com/prometheus/prometheus/issues/5606

This ignores any errors while linking tombstones to the snapshot.

### DIFF
--- a/block.go
+++ b/block.go
@@ -603,18 +603,23 @@ func (pb *Block) Snapshot(dir string) error {
 		return errors.Wrap(err, "create snapshot chunk dir")
 	}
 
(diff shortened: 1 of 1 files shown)


krasi-georgiev:
IIRC there should be no case when the tombstone is missing. 
I think at block creation it always creates a tombstone file even if it is empty.

codesome:
Oh, I tried to find where is tombstone file created but was not able to find. Found it now (https://github.com/prometheus/tsdb/blob/882162d5b95ea23b26642192049e9590fc121242/compact.go#L608), my bad. Yes, it should not be missing.

PR #619, opened 2019-06-03
wal: Replace promauto with injected registry

brancz:
All of tsdb consistently uses an injected registry, let's stick with that pattern.

@gouthamve @krasi-georgiev @cstyan

### DIFF
--- a/wal/live_reader.go
+++ b/wal/live_reader.go
@@ -24,19 +24,11 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/client_golang/prometheus/promauto"
-)
-
-var (
-	readerCorruptionErrors = promauto.NewCounterVec(prometheus.CounterOpts{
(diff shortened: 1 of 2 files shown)


brancz:
Agreed.

wal/live_reader.go
```diff
@@ -24,26 +24,29 @@ import (
...
 		logger: logger,
 		rdr:    r,
 
 		// Until we understand how they come about, make readers permissive
 		// to records spanning pages.
 		permissive: true,
 	}
+
+	lr.readerCorruptionErrors = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "prometheus_tsdb_wal_reader_corruption_errors",
```
bwplotka: Probably we don't want to make incompatibility but shouldn't this have `_total` suffix? (:

bwplotka: Also I would remove `prometheus_` prefix and add it by wrapping Registerer instead. More flexible for any package user.

brian-brazil: We've already gone through this, this is the prometheus_tsdb library and metrics should be named accordingly.

PR #622, opened 2019-06-05
wal: Add `_total` suffix as metric is a counter

brancz:
Follow up from https://github.com/prometheus/tsdb/pull/619#discussion_r290019916

@bwplotka @gouthamve @cstyan @krasi-georgiev 

Signed-off-by: Frederic Branczyk <fbranczyk@gmail.com>


brian-brazil:
:+1:

PR #626, opened 2019-06-07
Print label names with highest cumulative label value length during tsdb analyze

csmarchbanks:
Add a print section for the total size of all the label values by each label. We found this instrumental recently for finding the culprit of very high symbol table size.

cc: @rscot231

### DIFF
--- a/cmd/tsdb/main.go
+++ b/cmd/tsdb/main.go
@@ -508,6 +508,30 @@ func analyzeBlock(b *tsdb.Block, limit int) {
 
 	postingInfos = postingInfos[:0]
 	for _, n := range allLabelNames {
+		values, err := ir.LabelValues(n)
+		if err != nil {
+			exitWithError(err)
+		}
+		var cumulativeLength uint64
+
+		for i := 0; i < values.Len(); i++ {
+			value, _ := values.At(i)
+			if err != nil {
+				exitWithError(err)
+			}
+			for _, str := range value {
+				cumulativeLength += uint64(len(str))
+			}
+		}
+
+		postingInfos = append(postingInfos, postingInfo{n, cumulativeLength})
+	}
+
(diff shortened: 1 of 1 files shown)


cmd/tsdb/main.go
```diff
@@ -473,6 +474,7 @@ func analyzeBlock(b *tsdb.Block, limit int) {
 			labelsUncovered[lbl.Name] += uncovered
 			labelpairsUncovered[key] += uncovered
 			labelpairsCount[key]++
+			labelValueLengths[lbl.Name] += uint64(len(lbl.Value))
```
brian-brazil: This is going to over-count, you only want to this once per value. Look at how highest cardinality is done.

csmarchbanks: :+1: Let me know if the new implementation looks better!

cmd/tsdb/main.go
```diff
@@ -506,6 +506,30 @@ func analyzeBlock(b *tsdb.Block, limit int) {
 	fmt.Printf("\nMost common label pairs:\n")
 	printInfo(postingInfos)
 
+	postingInfos = postingInfos[:0]
+	for _, n := range allLabelNames {
+		values, err := ir.LabelValues(n)
+		if err != nil {
+			exitWithError(err)
+		}
+		cumulativeLength := uint64(0)
```
brian-brazil: I'd do `var cumulativeLength unit64`

csmarchbanks: Done.

brian-brazil:
:+1:

csmarchbanks:
@krasi-georgiev Would you be willing to take a look at this when you get a chance?

krasi-georgiev:
no need, it is a rather small change and Brian already approved so merging.

PR #627, opened 2019-06-11
Attempt to reduce memory footprint of compaction

codesome:
What does this PR do
* Removes building of entire new postings using the postings from other blocks.
* Builds ([old series id] -> [new series id]) map for each block, and uses that to write postings 1 by 1 for each label-value pair.

I will do any more cleanup needed and post the benchmark results tomorrow. Hoping for some good news.

**Update**: This PR ended up have much more memory allocation savings in many places than just the above. [This comment](https://github.com/prometheus/tsdb/pull/627#issuecomment-504933179) summarizes it all (hopefully).

**UPDATE 2**: As this PR got very complicated for review, it is broken down into multiple PRs (more to come soon). For now these are the child PRs which originate from this PR
1. https://github.com/prometheus/tsdb/pull/643 - Breakdown generic `writeOffsetTable`
2. https://github.com/prometheus/tsdb/pull/644 - Check error before wrapping
3. https://github.com/prometheus/tsdb/pull/645 - Reuse `keys` in `ReadOffsetTable`
4. https://github.com/prometheus/tsdb/pull/653 - Reuse byte buffer in `WriteChunks` and `writeHash`
5. https://github.com/prometheus/tsdb/pull/654 - Reuse string buffer in `stringTuples.Swap`
6. https://github.com/prometheus/tsdb/pull/656 - `NumSeries()` method for `BlockReader` interface


compact.go
```diff
@@ -914,93 +947,109 @@ func (c *compactionSeriesSet) Err() error {
...
+func (c *compactionSeriesSet) At() (uint64, labels.Labels, []chunks.Meta, Intervals) {
+	return c.p.At(), c.l, c.c, c.intervals
 }
 
 type compactionMerger struct {
-	a, b ChunkSeriesSet
+	sets  []ChunkSeriesSet
+	first bool
+	oks   []bool
+	// TODO(enhancement): If we can know number of series in each block, we can
```
brian-brazil: Are the series ids consecutive though?

codesome:
Results are little disappointing
```
benchmark                                                                                old ns/op       new ns/op       delta
BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     22137952284     37095788828     +67.57%

benchmark                                                                                old allocs     new allocs     delta
BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     17821797       23269082       +30.57%

benchmark                                                                                old bytes      new bytes      delta
BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     2111300240     2192055400     +3.82%
```

Memory profile for this PR: [compact_opt_PR.zip](https://github.com/prometheus/tsdb/files/3279938/compact_opt_PR.zip)

Memory profile for master: [compact_opt_master.zip](https://github.com/prometheus/tsdb/files/3279939/compact_opt_master.zip)

codesome:
I managed to make it better but has a failing test in vertical compaction. Will push after I fix it.
```
benchmark                                                                                old ns/op       new ns/op       delta
BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     22137952284     30677731997     +38.58%

benchmark                                                                                old allocs     new allocs     delta
BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     17821797       20881225       +17.17%

benchmark                                                                                old bytes      new bytes      delta
BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     2111300240     1629291296     -22.83%
```

codesome:
This is the profile after the latest commit (getting rid of `postingMap`). Finally some reduced memory usage, but allocs are still more.
[compact_opt_noPostingMap.zip](https://github.com/prometheus/tsdb/files/3280257/compact_opt_noPostingMap.zip)

compact.go
codesome: I added `NumSeries()` for `BlockReader` and tried to use slice instead of map. Though I pre-allocate using `NumSeries()`, I was taking care if the series ref went beyond the slice and append appropriately. Though this was faster than upstream and B/op was less than upstream, B/op was still more than using map. 
One strange thing that I found out: The series ref went beyond the number of series. I saw series ref going upwards of 4M though number of series are 1M (this could be the reason why B/op was more than using map). @gouthamve this shouldn't happen right?

brian-brazil: That can happen.

codesome: I thought series ref spanned from `0` to `numSeries-1`.

codesome: Not sure if I was using the right term, but postings should span only till `numSeries-1` according to this
https://github.com/prometheus/tsdb/blob/6ab483071a5d4dce43634f6eef2dce67d1f4740f/compact.go#L819

codesome: I confused between id and reference (the offset). I will try id and see if it helps.

codesome:
Minute improvement after the last commit
```
BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     22627970234     24963923703     +10.32%

BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     17837934       20792238       +16.56%

BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     2095948320     1549497480     -26.07%
```

codesome:
This zip container memory and cpu profiles for both master and this PR while running those above benchmarks (the block creation was done independently and `BenchmarkCompaction` was modified to do only compaction) 
[mem_cpu_profiles.zip](https://github.com/prometheus/tsdb/files/3281478/mem_cpu_profiles.zip)

codesome:
While trying to reduce allocs, I was able to get the B/op less than before instead in the last few commits.
```
BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     15926839269     21870626474     +37.32%

BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     17854969       20803998       +16.52%

BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     2111793184     1377519288     -34.77%
```
Still digging more to reduce allocs.

codesome:
Found the culprit for increased allocs: https://github.com/prometheus/tsdb/blob/c1c39ed2e7900fb056445c9818754e3254518a49/compact.go#L870 (this line alone)

codesome:
With the latest commit (re-using the `bigEndianPostings` object instead of creating a new one every time), allocs reduced significantly (With ~17.8M allocs at 0%, +5.11% now compared to +16.x% before).

With that came more `B/op` reduction. CPU usage is little higher as expected.
```
BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     15923268078     21262455380     +33.53%

BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     17853104       18765444       +5.11%

BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     2112769168     1245923288     -41.03%
```

I will be looking into reducing more allocs if possible.

codesome:
Good news in the last 2 commits. Allocs reduced significantly and the diff for allocs is negative now.
```
BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     16858323771     22428340021     +33.04%

BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     17847227       11111037       -37.74%

BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     2112914256     1152807768     -45.44%
```

codesome:
More reduction in allocs in the last few commits!
```
BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     16858323771     20462639317     +21.38%

BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     17847227       10110743       -43.35%

BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     2112914256     1151946488     -45.48%
```

codesome:
Calling it a day on this PR with the following numbers, will continue one last time on it tomorrow.
```
BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     16858323771     19238843123     +14.12%

BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     17847227       5111341        -71.36%

BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     2112914256     1037778768     -50.88%
```
Some of them looks like some hacks, but I guess it's worth having.

codesome:
I am stopping anymore optimization in this PR to reduce efforts in reviews. Final numbers after the last commit:
```
BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     15797027093     18920848674     +19.77%

BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     17855829       4111140        -76.98%

BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     2111896368     1023454832     -51.54%
```
This PR is now ready for review.

chunks/chunks.go
```diff
@@ -320,7 +326,7 @@ func (w *Writer) WriteChunks(chks ...Meta) error {
 		}
 
 		w.crc32.Reset()
-		if err := chk.writeHash(w.crc32); err != nil {
+		if err := chk.writeHash(w.crc32, b[:1]); err != nil {
```
brian-brazil: Passing :1 here is odd, I'd expect b[:0] to be passed.

compact.go
```diff
@@ -788,11 +784,11 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
 				return errors.Wrap(err, "merge overlapping chunks")
 			}
 		}
-		if err := chunkw.WriteChunks(mergedChks...); err != nil {
+		if err := chunkw.WriteChunks(b[:], mergedChks...); err != nil {
```
brian-brazil: You mean b[:0] I suspect.

chunks/chunks.go
```diff
@@ -297,10 +303,10 @@ func (w *Writer) WriteChunks(chks ...Meta) error {
 		}
 	}
 
-	var (
-		b   = [binary.MaxVarintLen32]byte{}
-		seq = uint64(w.seq()) << 32
-	)
+	var seq = uint64(w.seq()) << 32
+	if len(b) < binary.MaxVarintLen32 {
+		b = make([]byte, binary.MaxVarintLen32)
```
brian-brazil: If this is made bigger here, then it doesn't get reused in subsequent calls.

compact.go
```diff
@@ -835,12 +846,54 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
...
+	seriesMap := set.SeriesMap()
+	var bigEndianPos index.Postings = index.NewBigEndianPostings(nil)
+	var listPos = index.NewListPostings(nil).(*index.ListPostings)
+	for _, n := range names {
+		s = s[:0]
+		if n == apkName {
+			s = append(s, apkValue)
+		} else {
+			for v := range values[n] {
+				s = append(s, v)
```
brian-brazil: Why not s = append(s, values[n]...)

compact.go
```diff
@@ -835,12 +846,54 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
...
+
+	if idxw, ok := indexw.(*index.Writer); ok {
+		idxw.HintPostingsWriteCount(numLabelValues)
+	}
 
-	for _, l := range postings.SortedKeys() {
-		if err := indexw.WritePostings(l.Name, l.Value, postings.Get(l.Name, l.Value)); err != nil {
-			return errors.Wrap(err, "write postings")
+	postingBuf := make([]uint64, 0, 1000000)
+	seriesMap := set.SeriesMap()
```
brian-brazil: This is all in the same file, why not access the private field directly?

block.go
```diff
@@ -455,21 +461,29 @@ func (r blockIndexReader) LabelValues(names ...string) (index.StringTuples, erro
 	return st, errors.Wrapf(err, "block: %s", r.b.Meta().ULID)
 }
 
-func (r blockIndexReader) Postings(name, value string) (index.Postings, error) {
-	p, err := r.ir.Postings(name, value)
-	return p, errors.Wrapf(err, "block: %s", r.b.Meta().ULID)
+func (r blockIndexReader) Postings(name, value string, reusePos index.Postings) (index.Postings, error) {
+	p, err := r.ir.Postings(name, value, reusePos)
+	if err != nil {
+		// Checking for error before wrapping saves some allocs.
```
brian-brazil: How does this save allocs? errors.Wrapf returns immediately if err is nil.

index/index.go
```diff
@@ -893,7 +928,8 @@ func (r *Reader) Series(id uint64, lbls *labels.Labels, chks *[]chunks.Meta) err
 }
 
 // Postings returns a postings list for the given label pair.
-func (r *Reader) Postings(name, value string) (Postings, error) {
+// 'reusePos' is used if it is a 'bigEndianPostings'.
```
brian-brazil: This feels like a layering violation

compact.go
```diff
@@ -835,12 +846,54 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
...
+			for v := range values[n] {
+				s = append(s, v)
+			}
+			sort.Strings(s)
+		}
+
+		for _, v := range s {
+			postingBuf = postingBuf[:0]
+			for i, ir := range indexReaders {
+				bigEndianPos, err = ir.Postings(n, v, bigEndianPos)
```
brian-brazil: pos reads as position, rather than postings

compact.go
```diff
@@ -835,12 +846,54 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
...
+				if err != nil {
+					return errors.Wrap(err, "read postings")
+				}
+				for bigEndianPos.Next() {
+					if newVal, ok := seriesMap[i][bigEndianPos.At()]; ok {
+						// idx is the index at which newVal exists or index at which we need to insert.
```
brian-brazil: Insertion sort is n^2. How about putting it all into a map, and then sorting the keys? That'd be nlogn.

compact.go
```diff
@@ -914,93 +967,109 @@ func (c *compactionSeriesSet) Err() error {
...
 	}
-	a, _, _ := c.a.At()
-	b, _, _ := c.b.At()
-	return labels.Compare(a, b)
+	return c, nil
 }
 
 func (c *compactionMerger) Next() bool {
-	if !c.aok && !c.bok || c.Err() != nil {
+	if c.first {
```
brian-brazil: Why not do this in the constructor?

compact.go
codesome: I meant `[:]` as I was using `len` inside the function. Will change it to `[:0]` and use `cap` instead.

compact.go
codesome: `values[n]` is a `map[string]struct{}`.

compact.go
brian-brazil: Don't use cap, rather use append() to add in the value that's always there.

block.go
codesome: The extra function call results in few allocs. Coupled that with 100s of thousands of calls, the savings in allocs is pretty significant (Same for `Series` function below).

I got that insight from the profile (`-alloc_objects` flag), and % allocs saved was significant in the benchmark.

index/index.go
codesome: Did not understand what is being violated.

compact.go
codesome: It was like that before. But that seemed to cause a lot of allocs because I cannot re-use the map. With improvements in other areas for allocs, I will try it again.

codesome: Now I remember why I removed - it was a significant part of memory usage. I tried it again and memory savings become half. Also for some reason time also goes up (both reusing maps by deleting keys and not reusing map).

codesome: Maybe the reduced allocs/memory covers up the use of maps and heap/quick sort.

block.go
brian-brazil: Function calls don't cause allocs.

index/index.go
brian-brazil: You're altering a generic API merely for a micro-optimisation in one of its implementations.

block.go
codesome: Yes, function calls don't. But some part in it apparent caused allocs. I will post the screenshots of profiles here in some time.

index/index.go
codesome: This saved like 10% of allocs, so I would not count it as micro-opt. But I agree that altering this generic API is not clean. I will see if I can do it in a better way, else might as well leave it as allocs got reduced in other places to cover it up.

block.go
codesome: Before = direct Wrapf, Now = checking error before.

## `Postings(...)`
**Before:**
Screenshot from 2019-06-20 15-14-34
Screenshot from 2019-06-20 15-14-56

**Now:**
Can see that `ir.Postings(...)` causes no allocs. Could not get profile of `Postings(...)` in block.go as there were no allocs.
Screenshot from 2019-06-20 15-11-29

## `Series(...)`
**Before:**
Screenshot from 2019-06-20 15-14-41
(see `set.Next()`)
Screenshot from 2019-06-20 15-15-02

**Now:**
Could not get profile of `Series(...)` in block.go as there were no allocs.
Screenshot from 2019-06-20 15-12-21

index/index.go
codesome: Initially I was thinking to have `Reset(...)` method for `Postings` interface, but could not decide what would be the type of the argument. Now I see the argument can be `interface{}` and particular `Postings` interface implementations will typecast to needed types and reset and return `true` if reset is successful. This way we will use the generic `Reset(...) bool` function without going to specifics.

If `Reset(...)` returned `false`, it means that we need to create another object.

Does this look sane in any way?

brian-brazil: That sounds okay, but I'm not sure why you need the bool return.

codesome: Some of the postings have `[]byte` as base (example `bigEndianPostings`) and some have `[]uint64` (example `ListPostings`). So if I pass `[]uint64` to `bigEndianPostings.Reset`, I am not sure if the `Reset` would make sense, so in that case I would return false.

codesome: I have added that method in the `Postings` interface.

index/index.go
```diff
@@ -1025,11 +1064,21 @@ type Decoder struct {
 }
 
 // Postings returns a postings list for b and its number of elements.
-func (dec *Decoder) Postings(b []byte) (int, Postings, error) {
+func (dec *Decoder) Postings(b []byte, reusePos Postings) (int, Postings, error) {
 	d := encoding.Decbuf{B: b}
 	n := d.Be32int()
-	l := d.Get()
-	return n, newBigEndianPostings(l), d.Err()
+	if bep, ok := reusePos.(*bigEndianPostings); ok {
```
codesome: I have this special case even though I am calling `reusePos.Reset(...)` later because `l := d.Get()` needs an alloc as we are passing it as a pointer. We can skip that alloc if `reusePos` is `bigEndianPostings`.

This looks like a hack but doesn't take away the generic interface flow.

compact.go
```diff
@@ -784,7 +784,7 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
 				return errors.Wrap(err, "merge overlapping chunks")
 			}
 		}
-		if err := chunkw.WriteChunks(b[:], mergedChks...); err != nil {
+		if b, err = chunkw.WriteChunks(b[:0], mergedChks...); err != nil {
```
bwplotka: This is tricky, easy to make mistake and add `:=` and start shadowing `b`. I might think we don't need to return b (see comments above) but if we do separate `if err != nil` might be better

block.go
bwplotka: Wow it's quite interesting, but makes sense (:

bwplotka: Also it's more clear / explcit so :+1: 

I would delete the comment though. I think it is  general knowledge that should be ensure by some linter and documented. Otherwise you need to put this comment on every `Wrapf` with reminder to avoid `Wrapf` on nil error. That does not scale well (:

block.go
```diff
@@ -75,7 +75,9 @@ type IndexReader interface {
 	// The Postings here contain the offsets to the series inside the index.
 	// Found IDs are not strictly required to point to a valid Series, e.g. during
 	// background garbage collections.
-	Postings(name, value string) (index.Postings, error)
+	// 'reusePos' is the Postings object that can be re-used instead of allocating
+	// a new object. Re-use of this object need not be guaranteed.
+	Postings(name, value string, reusePos index.Postings) (index.Postings, error)
```
bwplotka: `reusePostings` or `reusePosts` to be consistent and avoiding `Pos` which can be confused with `position`?

block.go
```diff
@@ -109,9 +111,10 @@ type StringTuples interface {
 type ChunkWriter interface {
 	// WriteChunks writes several chunks. The Chunk field of the ChunkMetas
 	// must be populated.
+	// b is the byte buffer that can be used by WriteChunks.
 	// After returning successfully, the Ref fields in the ChunkMetas
 	// are set and can be used to retrieve the chunks from the written data.
-	WriteChunks(chunks ...chunks.Meta) error
+	WriteChunks(b []byte, chunks ...chunks.Meta) ([]byte, error)
```
bwplotka: Can we comment the returned buffered somehow? I guess it is for better reuse (we reuse full slice not only array).

Also I don't think we need this if we would ensure slice is greater than 5 binary.MaxVarintLen32 here, right? This is the only place we grow the slice.

By look of usage b is already of size b = make([]byte, binary.MaxVarintLen32) in all cases?

block.go
bwplotka: Could you elaborate on ` Re-use of this object need not be guaranteed.` How this helps / warns caller? I am missing the meaning of this.

chunks/chunks.go
```diff
@@ -55,8 +55,9 @@ type Meta struct {
 }
 
 // writeHash writes the chunk encoding and raw data into the provided hash.
-func (cm *Meta) writeHash(h hash.Hash) error {
-	if _, err := h.Write([]byte{byte(cm.Chunk.Encoding())}); err != nil {
+func (cm *Meta) writeHash(h hash.Hash, buf []byte) error {
```
bwplotka: worth to add a comment about `buf`? Also wonder if it's not worth to hide this buffer in `cm`? especially that it is private function?

compact.go
```diff
@@ -816,17 +812,32 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
...
+		// No postings to write, exit early.
+		return nil
+	}
+
+	var (
+		maxNumValues   int
+		numLabelValues int
+		names          = make([]string, 0, len(values)+1)
+	)
+	for _, v := range values {
```
bwplotka: nice, wonder if we can move that to separate function

compact.go
```diff
@@ -835,12 +846,55 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
...
+		for _, v := range s {
+			postingBuf = postingBuf[:0]
+			for i, ir := range indexReaders {
+				bigEndianPost, err = ir.Postings(n, v, bigEndianPost)
+				if err != nil {
+					return errors.Wrap(err, "read postings")
+				}
+				for bigEndianPost.Next() {
+					if newVal, ok := set.seriesMap[i][bigEndianPost.At()]; ok {
+						// idx is the index at which newVal exists or index at which we need to insert.
```
bwplotka: We could reverse the case here and handle `!ok` first to reduce indent?

index/index.go
bwplotka: If the ONLY gain for any caller of this all `Postings` implementation is if the `Postings` is  `(*bigEndianPostings)` this makes the interface quite complex. Maybe we should expect `*BigEndianPostings` then explicitly? This makes `Postings` quite complext to use.

block.go
codesome: As these 2 functions are called a lot of time, it should be fine changing only at these 2 places. Also keeping this comment here would prevent it from reverting back in future as it looks natural to revert back.

index/index.go
codesome: This would mean assuming the inner details of `IndexReader` interface. It could also become harder to refactor if a specific posting is mentioned in the function but the index uses multiple postings (Example if something like #629 is merged). Currently, it would mean modifying this function rather than modifying function signatures if such changes are added in the future.

But I agree that it might become harder to maintain this. I am still trying to figure out any better ways to do it (don't want to let it go as gains are significant).

compact.go
codesome: As `WriteChunks` is an interface, I would not make any assumptions of how much bytes does it allocate (though we can see it easily). So design-wise returning should be fine, which would either be the same slice or a bigger slice.

block.go
codesome: > reusePostings or reusePosts to be consistent and avoiding Pos which can be confused with position?

Ah yes, I will change it. Thanks.

> Could you elaborate on Re-use of this object need not be guaranteed. How this helps / warns caller? I am missing the meaning of this.

I just meant to say that the implementation need not re-use this object, and it is fine to allocate a new object and return it instead. Maybe that sentence is not required at all as I say "_can_ be re-used" already.

block.go
codesome: I will add a comment about it.

And [this comment](https://github.com/prometheus/tsdb/pull/627#discussion_r296443943) for the rest of the part.

chunks/chunks.go
codesome: Will add a comment.

> Also wonder if it's not worth to hide this buffer in cm?

https://github.com/prometheus/tsdb/blob/eb79899bfc3c2edb40ad1ba3af00056d1ecd735e/chunks/chunks.go#L305-L330
That would mean we will duplicate/share those bytes among all the chunk metas (`writeHash` is being called in a loop on different `Meta`). So no, it's not worth.

compact.go
codesome: Done

codesome: Not sure which part you wanted to move to a separate function, I just moved the entire code of writing postings into a separate function as that was getting big.

compact.go
codesome: Done

codesome:
> I think I would split this PR at some point, it's hard to review a bit, in terms of size (:

Yeah I stopped anymore work because it was getting big. I will document what all changed as a comment here (would not like to create another PR). That should make reviewing easier I hope.

compact.go
codesome: Just checked that you cannot have `:=` here, as `b` would be declared and never used. Do you still prefer to be in a separate line?

codesome:
This PR started with implementing `(old series id) -> (new series id)` map and using that to write postings for the new block.
Following are changes that are related to the above, and also independent optimizations.

1. `ChunkSeriesSet.At()` now returns ref of the posting too. This becomes the (old series id) and helps us build the map. 9811a141eeda89031b33634b2af6c4543116294e
2. `(old series id) -> (new series id)` map is calculated inside `compactionMerger` itself during `Next()` calls inside compact.go.
3. `BlockReader` now has `NumSeries()` function. It helps to preallocate some memory for the above map.
4. `index.Writer` now has `HintPostingsWriteCount` method, which pre allocates the size of `index.Writer.postings` map.
5. `IndexReader.Postings` now takes a `Postings` object which can be re-used instead of creating a new object. For this PR, we re-use the `BigEndianPostings` and reset the byte buffer of the postings.
6. Following the above, `Postings` interface got a new method called `Reset(interface{})`, which can be used to reset the data of the `Postings` as if it's newly created (only implemented for `BigEndianPostings` and `ListPostings` for now). 53ccf99acc26ec24ba26039455e35c0f3a07fece
7. `stringTuples` in `index/index.go` uses a common buffer in `stringTuples.Swap` method. This saves a good number of allocs when sorting.
8. `WriteChunks` in `chunks/chunks.go` now takes a byte slice which can be reused instead of allocating a new one. f9dfe070524541010567abf4c197284c14210ec2
9. The above byte slice passed to `WriteChunks` is again passed down to `writeHash` for re-use.
10. Re-use `keys` slice in `ReadOffsetTable` function in `index/index.go`. This reduces allocs for `OpenBlock` and not the compaction itself. (Should this be in another PR?) f9dfe070524541010567abf4c197284c14210ec2
11. In `blockIndexReader.Series` and `blockIndexReader.Postings` in `block.go`, error is checked before wrapping to reduce allocs. 4ff0ed8047e0f5a5144f8d0e4bb6de8cc8dcaa28 387705d11d2409b720e4157c0642b92933310b6d
12. Broke down `writeOffsetTable` into `writeLabelIndexesOffsetTable` and `writePostingsOffsetTable`. With this there was not need to allocate a new slice for the `hashEntry` object for writing postings and we could have only `name` and `value`. This also saves a good amount of allocs. e812fbaeb3401274e4f865ff25d1e2fee23c944f.
13. `populateBlock` was broken down into `populateBlock` and `writePostings`, and writing of postings was moved into `writePostings`.

The final changes can be little different at places than what is in individual commits.

@bwplotka I hope this helps to review if you still wish to. I might have missed some small details, but this is it at a high-level view.

index/index.go
```diff
@@ -447,12 +447,12 @@ func (w *Writer) WriteLabelIndex(names []string, values []string) error {
 	return errors.Wrap(err, "write label index")
 }
 
-// writeOffsetTable writes a sequence of readable hash entries.
-func (w *Writer) writeOffsetTable(entries []hashEntry) error {
```
cstyan: ohhhhh this would have been alloc'ing the slice header even though the slice values are pass by reference, nice find

head.go
```diff
@@ -1025,6 +1030,8 @@ func (h *Head) gc() {
 	h.metrics.series.Sub(float64(seriesRemoved))
 	h.metrics.chunksRemoved.Add(float64(chunksRemoved))
 	h.metrics.chunks.Sub(float64(chunksRemoved))
+	// Ref: https://golang.org/pkg/sync/atomic/#AddUint64
```
cstyan: :+1:

index/index.go
```diff
@@ -781,9 +815,10 @@ func ReadOffsetTable(bs ByteSlice, off uint64, f func([]string, uint64) error) e
 	d := encoding.NewDecbufAt(bs, int(off), castagnoliTable)
 	cnt := d.Be32()
 
+	keys := make([]string, 0, 2)
```
cstyan: any particular reason for 2?

chunks/chunks.go
```diff
@@ -293,14 +295,14 @@ func (w *Writer) WriteChunks(chks ...Meta) error {
...
+			return b, err
 		}
 	}
 
-	var (
-		b   = [binary.MaxVarintLen32]byte{}
-		seq = uint64(w.seq()) << 32
-	)
+	var seq = uint64(w.seq()) << 32
+	for len(b) < binary.MaxVarintLen32 {
```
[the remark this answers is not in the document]
codesome: I will add a comment. It is just a way of assuring that the size of `b` is at least `MaxVarintLen32`, without making any assumptions of the passed byte slice.

chunks/chunks.go
```diff
@@ -293,14 +295,14 @@ func (w *Writer) WriteChunks(chks ...Meta) error {
...
-			return err
+			return b, err
 		}
 	}
 
-	var (
-		b   = [binary.MaxVarintLen32]byte{}
-		seq = uint64(w.seq()) << 32
-	)
+	var seq = uint64(w.seq()) << 32
```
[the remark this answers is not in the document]
codesome: `seq` sounds like sequence.

```go
func (w *Writer) seq() int {
	return len(w.files) - 1
}
```

And it seems to be the index of the last chunk file.

index/postings.go
```diff
@@ -268,17 +268,23 @@ type Postings interface {
 
 	// Err returns the last error of the iterator.
 	Err() error
+
+	// Reset resets the postings using the given data.
+	// It is similar to creating a new Postings using that data.
+	// If Reset() was successful, it returns true.
+	Reset(data interface{}) bool
```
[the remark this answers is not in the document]
codesome: After another look, I think `Reset` is not required in that interface, because all the places where I use, I am first dereferencing and then using it. I added another commit which removes this methods and keeps it only for `ListPostings` and `BigEndianPostings`.

> `Clear(...)` maybe?

I remember seeing `Reset` in a lot of other places for similar use (start the object fresh with new data). `Clear` sounds more like clearing the object without setting the new data.

index/index.go
codesome: Yup

index/index.go
codesome: As per the current use case, not more than 2 keys are required
https://github.com/prometheus/tsdb/blob/c7af8aa9f3752aa39a6602f8143e44ee41485ecd/index/index.go#L717-L740

compact.go
codesome: Done

index/index.go
cstyan: should we be checking the value of `d.Uvarint` to make sure it's 2 then? because we append `keyCount` worth of keys to `keys`

index/postings.go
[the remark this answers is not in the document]
cstyan: > Clear sounds more like clearing the object without setting the new data.

Reset suggests the same thing to me. What about `Overwrite`?

chunks/chunks.go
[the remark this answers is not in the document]
cstyan: hmmm :neutral_face:

index/postings.go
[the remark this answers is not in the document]
codesome: I would like to keep it `Reset` to be consistent.

index/index.go
codesome: As per the current design, the keyCount doesn't exceed 2.

Postings table has keyCount of 2 (and I think it will remain 2).
https://github.com/prometheus/tsdb/blob/c7af8aa9f3752aa39a6602f8143e44ee41485ecd/index/index.go#L471

And Label Index table has keyCount of 1
https://github.com/prometheus/tsdb/blob/c7af8aa9f3752aa39a6602f8143e44ee41485ecd/index/index.go#L451
https://github.com/prometheus/tsdb/blob/c7af8aa9f3752aa39a6602f8143e44ee41485ecd/compact.go#L852

During my GSoC last year, there was a discussion about having label indices for composite label names, but it was decided that we don't need it though the design still allows it. So I don't expect it to exceed 1 anytime soon.

codesome:
I am planning to move some of the stuff here to new PRs. Expect some PRs soon.

index/index.go
cstyan: I just feel like it's worth codifying (or at least leaving a comment) somewhere about this. It wasn't even clear to me what the keys were initially.

cstyan:
:+1: thanks for breaking this up

codesome:
This last commit https://github.com/prometheus/tsdb/pull/627/commits/f8ddc0330af88d108a8cabd109cea63ba0f3672f shed off about 2-3s of compaction time from a total of 15-17s before. The change was the avoid repeated binary search on the `postingBuf` and replace with a single linear scan.

Now it compares like this with the current master.
```
BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     13061950871     15038790382     +15.13%

BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     9149767        7111218        -22.28%

BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     1980380560     1031404000     -47.92%
```

This benchmark results are on top of recently merged optimizations, and does not relate directly to numbers in this comment https://github.com/prometheus/tsdb/pull/627#issuecomment-503453181 (they were w.r.t. the master without any optimizations).

compact.go
```diff
@@ -844,31 +839,107 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
...
+	if idxw, ok := indexw.(*index.Writer); ok {
+		idxw.HintPostingsWriteCount(numLabelValues)
+	}
 
-	for _, l := range postings.SortedKeys() {
-		if err := indexw.WritePostings(l.Name, l.Value, postings.Get(l.Name, l.Value)); err != nil {
-			return errors.Wrap(err, "write postings")
+	postingBuf := make([]uint64, 0, 1000000)
+	var bigEndianPost index.Postings = index.NewBigEndianPostings(nil)
+	var listPost = index.NewListPostings(nil).(*index.ListPostings)
```
gouthamve: Can we make `NewListPostings` return a `ListPostings`?

compact.go
```diff
@@ -844,31 +839,107 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
...
+
+	if idxw, ok := indexw.(*index.Writer); ok {
+		idxw.HintPostingsWriteCount(numLabelValues)
+	}
 
-	for _, l := range postings.SortedKeys() {
-		if err := indexw.WritePostings(l.Name, l.Value, postings.Get(l.Name, l.Value)); err != nil {
-			return errors.Wrap(err, "write postings")
+	postingBuf := make([]uint64, 0, 1000000)
+	var bigEndianPost index.Postings = index.NewBigEndianPostings(nil)
```
gouthamve: Can we make `NewBigEndianPostings` return `index.Postings`?

compact.go
```diff
@@ -844,31 +839,107 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
...
+			postingBuf = postingBuf[:0]
+			for i, ir := range indexReaders {
+				bigEndianPost, err = ir.Postings(n, v, bigEndianPost)
+				if err != nil {
+					return errors.Wrap(err, "read postings")
+				}
+				sMap := seriesMap[i]
+				idx, lastIdx := -1, -1
+				for bigEndianPost.Next() {
+					newVal, ok := sMap[bigEndianPost.At()]
```
gouthamve: Oof, complexity :) I wonder if just appending everything to `postingBuf` and sorting at the end is simpler without too much change in perf.

gouthamve: Or even better, new type called `remappedPostings` which takes the `sMap` and `postings`, and gives `Postings` which returns remapped values. And use those and `mergedPostings` to create the new list.

compact.go
codesome: This is contrary to what you said for `ListPostings`. I will make `NewListPostings` returning `ListPostings` and keep this as it is for consistency.

gouthamve: Well it's useless to return an internal type when calling an exported function. Any caller must know it's an index.Postings and cast it to that. Exported functions should return exported types :)

compact.go
codesome: > I wonder if just appending everything to postingBuf and sorting at the end is simpler without too much change in perf.

This would need deduplication. Also, it would mean we are storing more in the memory, upto `n` times more where `n` is the number of blocks being compacted. 

> Or even better, new type called remappedPostings which takes the sMap and postings, and gives Postings which returns remapped values. And use those and mergedPostings to create the new list.

This could make it more readable. My idea is that `remappedPostings` would take postings one by one and add it in the same way as now, and it would take `seriesMap` and not `sMap` (from the code). If we want to re-use the `mergedPostings`, it would again mean keeping postings from all the blocks together. Instead, a modified `mergedPostings` which maps the postings on the fly.

compact.go
codesome: > Exported functions should return exported types

Now, this makes sense to me.

head.go
```diff
@@ -1349,6 +1361,7 @@ func (h *Head) getOrCreateWithID(id, hash uint64, lset labels.Labels) (*memSerie
 
 	h.metrics.series.Inc()
```
gouthamve: We're tracking this through `NumSeries()`. Can we just drop the gauge and use gaugeFunc()

codesome: Will do. Also, I have another PR for `NumSeries()` https://github.com/prometheus/tsdb/pull/656 if that would help in reviewing. Do you want to be in a separate PR or inside this PR itself?

compact.go
```diff
@@ -844,31 +839,83 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
...
+
+	if idxw, ok := indexw.(*index.Writer); ok {
+		idxw.HintPostingsWriteCount(numLabelValues)
+	}
+
+	remapPostings := index.NewRemappedPostings(seriesMap, 1e6)
+	var bigEndianPost index.Postings = index.NewBigEndianPostings(nil)
+	for _, n := range names {
+		labelValuesBuf = labelValuesBuf[:0]
+		if n == apkName {
```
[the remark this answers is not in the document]
codesome: `apkName` is added separately to `names` here https://github.com/prometheus/tsdb/blob/4293e37c572ae2385e61dc3de3af1f7cd59bc3a9/compact.go#L883. So its values is not in `values` map below. Hence the special case.

I think adding it in `values` should not matter much, I will do that and remove the special case.

compact.go
```diff
@@ -844,31 +839,83 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
...
+			sort.Strings(labelValuesBuf)
+		}
 
-	for _, l := range postings.SortedKeys() {
-		if err := indexw.WritePostings(l.Name, l.Value, postings.Get(l.Name, l.Value)); err != nil {
-			return errors.Wrap(err, "write postings")
+		for _, v := range labelValuesBuf {
+			remapPostings.ClearPostings()
+			for i, ir := range indexReaders {
+				bigEndianPost, err = ir.Postings(n, v, bigEndianPost)
```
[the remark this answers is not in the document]
codesome: Right. I have changed the name and skipped initializing it to `BigEndianPostings`.

index/postings.go
```diff
@@ -686,6 +691,75 @@ func (it *bigEndianPostings) Seek(x uint64) bool {
...
 }
+
+func (it *BigEndianPostings) Reset(list []byte) {
+	it.cur = 0
+	it.list = list
+}
+
+// RemappedPostings is useful in remapping set of postings
+// from given set of maps.
+type RemappedPostings struct {
```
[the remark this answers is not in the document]
codesome: Moved.

codesome:
While running prombench with this PR, compactions for head block was failing with the following error.
```
level=error ts=2019-07-16T09:58:13.238Z caller=db.go:377 component=tsdb msg=\"compaction failed\" err=\"persist head block: write compaction: write postings: 0xc36a049540 series for reference 17138146 not found\"\n
```

I am not able to replicate it yet. Compaction from head is working fine in synthetic test.

codesome:
Able to re-create by just running prometheus in local machine! Will dig in more.

codesome:
The previous way of merging remapped postings did not scale well. I was keeping a sorted list of postings. And adding the postings in that list while keeping it sorted. This was slow. During prombench test, compaction of on-disk blocks was stuck at this place and allocations also shot up (it had ~13-14M series in total from all the blocks being compacted).

Now I am gathering the postings list from one index reader at a time and merging the sorted list in the old fashion way. I tested this on the indices that I had downloaded from prombench, and they finally don't get stuck here.

Below are the updated (synthetic) benchmark results.

2 blocks, same as all the benchmarks in this PR thread.
```
benchmark                                                                                  old ns/op       new ns/op       delta
BenchmarkCompaction2/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     12804627758     14663944235     +14.52%

benchmark                                                                                  old allocs     new allocs     delta
BenchmarkCompaction2/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     6149618        4111065        -33.15%

benchmark                                                                                  old bytes      new bytes      delta
BenchmarkCompaction2/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     1973256560     1042666800     -47.16%
```

This is for 3 blocks. An indication that with more blocks the time will get more worse.:
```
benchmark                                                                                  old ns/op       new ns/op       delta
BenchmarkCompaction2/type=normal,blocks=3,series=1000000,samplesPerSeriesPerBlock=51-8     15223759663     19132151137     +25.67%

benchmark                                                                                  old allocs     new allocs     delta
BenchmarkCompaction2/type=normal,blocks=3,series=1000000,samplesPerSeriesPerBlock=51-8     7135520        5096165        -28.58%

benchmark                                                                                  old bytes      new bytes      delta
BenchmarkCompaction2/type=normal,blocks=3,series=1000000,samplesPerSeriesPerBlock=51-8     2137628704     1276175792     -40.30%
```

I will test this on prombench now.

codesome:
Closing it for now. Analysis of this PR vs master is in this doc https://docs.google.com/document/d/1ZzZ8sslkA5LPshr9gqz-MmRpPmyY19jFDz-AKrkwpVw/edit?usp=sharing.

bwplotka:
Thanks for that work guys, I think we learnt a lot from this despite negative outcome of this particular experiment!

PR #628, opened 2019-06-12
Make objectives of Summaries explicit

beorn7:
With the next release of client_golang, Summaries will not have
objectives by default. To not lose the objectives we have right now,
explicitly state the current default objectives.

This strictly doesn't change anything with the current
metrics. However, similar to what's being discussed in
https://github.com/prometheus/alertmanager/pull/1921, some of these
summaries probably only receive observations every couple of hours or
so. In those cases, the status-quo of the quantiles doesn't make
sense. (They expire after 10m.) Either the MaxAge has to be increased
a lot for the affected summaries, or the Objectives should be removed
altogether. Please advise.

Signed-off-by: beorn7 <beorn@grafana.com>

### DIFF
--- a/head.go
+++ b/head.go
@@ -141,8 +141,9 @@ func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
 		Help: "Total number of chunks removed in the head",
 	})
 	m.gcDuration = prometheus.NewSummary(prometheus.SummaryOpts{
-		Name: "prometheus_tsdb_head_gc_duration_seconds",
-		Help: "Runtime of garbage collection in the head block.",
+		Name:       "prometheus_tsdb_head_gc_duration_seconds",
+		Help:       "Runtime of garbage collection in the head block.",
+		Objectives: map[float64]float64{},
 	})
 	m.maxTime = prometheus.NewGaugeFunc(prometheus.GaugeOpts{
 		Name: "prometheus_tsdb_head_max_time",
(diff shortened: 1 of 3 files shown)


brian-brazil:
wal/wal.go is the only one that might currently be useful, if you're ingesting at least ~35k/s.

krasi-georgiev:
I am not familiar with these so feel free to merge as you see fit based on some real use cases.
maybe @SuperQ 	can also advise here.

beorn7:
In doubt, I would remove the quantiles. I just didn't want to remove any quantile that people actually would like to see.

gouthamve:
Yep, `wal/wal.go` can be kept but the others should also be recorded but just a gauge that represents the last value.

beorn7:
Ok, so not even a summary without objectives?

brian-brazil:
I think a summary makes sense, as if e.g. I want to see how much time fsync took altogether I need counters.

beorn7:
I added the summary → gauge conversion as a separate commit.

beorn7:
Perhaps we should have both?
@gouthamve what do you think?

brian-brazil:
In normal circumstances, you can calculate that gauge from the summary by the usual rate+divide (I use that trick in some places on the benchmark dashboard).

beorn7:
But then you have to know the usual observation frequency, don't you?

brian-brazil:
You need to know the typical event frequency, and ensure your range is smaller than that. This only really works for things that happen hourly or so - which these are.

beorn7:
Wait a second. The fsync summaries were not changed after all.
They are the summaries where even the objectives make sense because they are updated more frequently.

Question is the two summaries that I actually changed into gauges.
- prometheus_tsdb_head_gc_duration_seconds
- prometheus_tsdb_wal_truncate_duration_seconds

Do we want to keep those as objective-less summaries, or change into a gauge, or have both?

brian-brazil:
I think those two should be objective-less summaries.

gouthamve:
Yeah, thinking about it more. We can always deduce the last duration from the jump in the graph too. Having sum + count works. I agree with Brian.

beorn7:
Thanks, everybody. Will switch to objective-less summaries for prometheus_tsdb_head_gc_duration_seconds and prometheus_tsdb_wal_truncate_duration_seconds and then merge.

beorn7:
Updated the commit description as well.

beorn7:
Will merge on green.

brian-brazil:
:+1:

PR #629, opened 2019-06-12
postings compression exploration

naivewong:
1. Store the first posting, then store the delta to that (not the delta to the previous element, the delta to the first number only). Find the minimum number of bits required to store the deltas. The bitpack using those many bits. Also be able to do binary search here.
2. Use blocks of postings, like 4KB block, store deltas with the previous number. Binary search to the right block and then iterate.

### DIFF
--- a/index/postings.go
+++ b/index/postings.go
@@ -16,11 +16,13 @@ package index
 import (
 	"container/heap"
 	"encoding/binary"
+	"math/bits"
 	"runtime"
 	"sort"
 	"strings"
 	"sync"
 
+	"github.com/prometheus/tsdb/encoding"
 	"github.com/prometheus/tsdb/labels"
 )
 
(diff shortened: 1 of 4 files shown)


naivewong:
all benchmarks: https://docs.google.com/document/d/1lSPRHaL8P-bILSogJNtZ_X5aVeC9opFTHvjwagQxgCY/edit#heading=h.fhoqupzha3ie

index/postings.go
```diff
@@ -689,3 +691,258 @@ func (it *bigEndianPostings) Seek(x uint64) bool {
...
+}
+
+func (it *deltaBlockPostings) Seek(x uint64) bool {
+	if it.cur >= x {
+		return true
+	}
+
+	startOff := it.offset / (deltaBlockSize * 8) * deltaBlockSize
+	num := len(it.bs.bstream) / deltaBlockSize - it.offset / (deltaBlockSize * 8)
+	// Do binary search between current position and end.
```
brian-brazil: Try checking if the next block is too big before doing the search.

naivewong:
- Fix a bug in Seek() and the benchmark results look much better.
- Add another idea of block postings, using delta to the base not the previous. And we can binary search inside the block also.
- Benchmark results updated.

PR #631, opened 2019-06-13
fix two benchmarks

naivewong:
- BenchmarkCompaction will always OOM even in Prometheus test machine
- BenchmarkCreateSeries has a high probability to fail

### DIFF
--- a/head_test.go
+++ b/head_test.go
@@ -34,7 +34,11 @@ import (
 )
 
 func BenchmarkCreateSeries(b *testing.B) {
-	lbls, err := labels.ReadLabels(filepath.Join("testdata", "20kseries.json"), b.N)
+	n := b.N
+	if n > 20000 {
+		n = 20000
+	}
+	lbls, err := labels.ReadLabels(filepath.Join("testdata", "20kseries.json"), n)
 	testutil.Ok(b, err)
 
 	h, err := NewHead(nil, nil, nil, 10000)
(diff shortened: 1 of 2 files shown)


krasi-georgiev:
most likely because of the `genSeries` , can you check if we can optimize it or if not will merge this change.

head_test.go
```diff
@@ -34,7 +34,11 @@ import (
 )
 
 func BenchmarkCreateSeries(b *testing.B) {
-	lbls, err := labels.ReadLabels(filepath.Join("testdata", "20kseries.json"), b.N)
+	n := b.N
+	if n > 20000 {
+		n = 20000
+	}
+	lbls, err := labels.ReadLabels(filepath.Join("testdata", "20kseries.json"), n)
```
krasi-georgiev: can we use `genSeries` here?

codesome:
@krasi-georgiev yes, while benchmarking compaction recently myself I found that creation of blocks takes a lot of memory. I plan to optimize it in the future.

head_test.go
codesome: We may need some options for `genSeries` in that case. For example `20kseries.json` has changing label names for series (label names do repeat), but `genSeries` will have the same label names for all the series.

krasi-georgiev:
@codesome is it ok to merge this change or shall @naivewong try to optimize genSeries instead?

codesome:
@krasi-georgiev It can be merged now. We can visit that later.

PR #633, opened 2019-06-14
move the wal repair logic in db.Open

krasi-georgiev:
This is to allow opening a wal in a read only mode without triggering a
repair.

This is as part of working on a PR to add opening a db in a read only mode.

https://github.com/prometheus/tsdb/pull/588

cc @codesome 	trying to brake the huge PR into few smaller ones.

### DIFF
--- a/wal/wal.go
+++ b/wal/wal.go
@@ -166,13 +166,12 @@ type WAL struct {
 	actorc      chan func()
 	closed      bool // To allow calling Close() more than once without blocking.
 
-	fsyncDuration       prometheus.Summary
-	pageFlushes         prometheus.Counter
-	pageCompletions     prometheus.Counter
-	truncateFail        prometheus.Counter
-	truncateTotal       prometheus.Counter
-	currentSegment      prometheus.Gauge
-	walCorruptionsTotal prometheus.Counter
+	fsyncDuration   prometheus.Summary
+	pageFlushes     prometheus.Counter
+	pageCompletions prometheus.Counter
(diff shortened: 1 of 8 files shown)


wal/wal.go
```diff
@@ -289,6 +294,7 @@ func (w *WAL) Repair(origErr error) error {
 	if !ok {
 		return errors.Wrap(origErr, "cannot handle error")
 	}
+	w.walCorruptionsTotal.Inc()
```
codesome: Is it safe to assume that `Repair` is always called after corruption? (Should not regress in future as this is a separate package)

krasi-georgiev: sorry I didn't understand what you mean. 
I think the metric is to notify for corruption, regardless if it has been repaired or not.

krasi-georgiev:
> Maybe you can also mention the accidental bugfix in the CHANGELOG.

I think there was no actual bug after all as wal.Repair would delete all wal files after the corrupted one so the reader wouldn't be able to read these anyway.

I think I even checked that one of the test covers such case and it wasn't failing.

wal/wal.go
codesome: Yes, so it should not happen in the future that we encounter a corruption but don't call `Repair` due to some reasons. In that case, corruption is not counted. Would it be better to increment the counter right when we encounter corruption?

krasi-georgiev: yeah I see you point. I moved it.

PR #634, opened 2019-06-14
refactor to set the block size when created.

krasi-georgiev:
Another change required for #588

cc @codesome


tombstones.go
```diff
@@ -119,6 +123,8 @@ func writeTombstoneFile(logger log.Logger, dir string, tr TombstoneReader) error
 		return err
 	}
 	f = nil
+	meta.Stats.NumBytes += size
```
codesome: This can result in higher `NumBytes` than it is actually on disk. 

Consider that `sizeof(stone1)` was already in `NumBytes`, and with another `Delete()`, `stone2` was added to the set of tombstones. In that case the tombstone file is written fresh, so the writes contain `stone1` and `stone2`. So incrementing like this would count `sizeof(stone1) + sizeof(stone1) + sizeof(stone2)`.

index/index.go
```diff
@@ -225,6 +227,7 @@ func (w *Writer) write(bufs ...[]byte) error {
 		if err != nil {
 			return err
 		}
+		w.size += int64(n)
```
codesome: Doesn't `w.pos` give the same thing?

krasi-georgiev: correct will update.

tombstones.go
krasi-georgiev: correct, will refactor.

krasi-georgiev:
@codesome 	thanks for the review. I refactored the code and added a proper test that checks the block sizes before and after a delete.

block.go
```diff
@@ -113,6 +119,10 @@ type ChunkWriter interface {
 	// are set and can be used to retrieve the chunks from the written data.
 	WriteChunks(chunks ...chunks.Meta) error
 
+	// Size returns the total size in bytes written by the writer.
+	// It also adds any bytes that would be written when calling the close method.
```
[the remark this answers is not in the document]
krasi-georgiev: Why would one need this extra details?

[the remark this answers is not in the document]
bwplotka: just `any bytes` - maybe moving to `close` would make sense then?

tombstones.go
bwplotka: :+1:

PR #636, opened 2019-06-19
fileutil.Replace - remove destination only when a directory.

krasi-georgiev:
In cases where a rename fails the `fileutil.Replace` would delete the
source files/folder.

There is no easy way to make directory renaming atomic, but for files
`os.Rename` is atomic and replaced the destination file so there is no
need to remove the destination file explicitly.

this was caught while working on the FS errors injections on rename in https://github.com/prometheus/tsdb/pull/583 with `block.Delete` . Before this change a failure during a rename would cause deleting the source meta and tombstone files.

### DIFF
--- a/fileutil/fileutil.go
+++ b/fileutil/fileutil.go
@@ -128,9 +128,20 @@ func Rename(from, to string) error {
 // Replace moves a file or directory to a new location and deletes any previous data.
 // It is not atomic.
 func Replace(from, to string) error {
-	if err := os.RemoveAll(to); err != nil {
-		return err
+	// Remove destionation only if it is a dir.
+	// Otherwise os.Rename replaces the destination file and is atomic.
+	{
+		f, err := os.Stat(to)
+		if err != nil {
+			return err
+		}
+		if f.IsDir() {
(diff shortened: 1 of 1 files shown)


fileutil/fileutil.go
```diff
@@ -128,9 +128,20 @@ func Rename(from, to string) error {
 // Replace moves a file or directory to a new location and deletes any previous data.
 // It is not atomic.
 func Replace(from, to string) error {
-	if err := os.RemoveAll(to); err != nil {
-		return err
+	// Remove destionation only if it is a dir.
```
bwplotka:
```suggestion
	// Remove destination only if it is a dir, os.Rename will fail on that case.
```

krasi-georgiev:
> I think it's fine, wonder how it behaves on different OS systems than Linux though.

yes probably broken for windows, but will wait for some Windows user to contribute the windows changes.

fileutil/fileutil.go
krasi-georgiev: I updated the comment to be more clear.

PR #637, opened 2019-06-21
refactor block size calculation

krasi-georgiev:
It now includes the size of the meta file itself as well for a more
correct block size.
It fixes a bug where the size didn't change when calling block.Delete.
Adds a dedicated test to ensure correct block sizes.
To allow opening the db in a read only mode it ensures the meta file is
not rewritten when the block size hasn't change .

Signed-off-by: Krasi Georgiev <8903888+krasi-georgiev@users.noreply.github.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,7 @@
 ## master / unreleased
  - [FEATURE] Provide option to compress WAL records using Snappy. [#609](https://github.com/prometheus/tsdb/pull/609)
  - [BUGFIX] Re-calculate block size when calling `block.Delete`.
-    - as part of this the `BlockStats` now includes size information for chunks,index,tombstone,meta separately.
-
+ - [CHANGE] `BlockStats` no longer holds size information and this is kept in memory.
 
 ## 0.8.0
  - [BUGFIX] Calling `Close` more than once on a querier returns an error instead of a panic.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## master / unreleased
  - [FEATURE] Provide option to compress WAL records using Snappy. [#609](https://github.com/prometheus/tsdb/pull/609)
  - [BUGFIX] Re-calculate block size when calling `block.Delete`.
- - [CHANGE] `BlockStats` no longer holds size information and this is kept in memory.
+- [CHANGE] The meta file `BlockStats` no longer holds size information. This is now dynamically calculated and kept in memory. It also includes the meta file size which was not included before.
 
 ## 0.8.0
  - [BUGFIX] Calling `Close` more than once on a querier returns an error instead of a panic.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## master / unreleased
  - [FEATURE] Provide option to compress WAL records using Snappy. [#609](https://github.com/prometheus/tsdb/pull/609)
+ - [BUGFIX] Re-calculate block size when calling `block.Delete`.
+    - as part of this the `BlockStats` now includes size information for chunks,index,tombstone,meta separately.
+
 
 ## 0.8.0
  - [BUGFIX] Calling `Close` more than once on a querier returns an error instead of a panic.
--- a/block.go
+++ b/block.go
@@ -429,8 +429,11 @@ func (pb *Block) GetSymbolTableSize() uint64 {
 func (pb *Block) setCompactionFailed() error {
 	pb.meta.Compaction.Failed = true
 	n, err := writeMetaFile(pb.logger, pb.dir, &pb.meta)
+	if err != nil {
+		return err
+	}
 	pb.numBytesMeta = n
-	return err
+	return nil
 }
 
 type blockIndexReader struct {
--- a/block.go
+++ b/block.go
@@ -240,8 +240,7 @@ func writeMetaFile(logger log.Logger, dir string, meta *BlockMeta) error {
 
 	// The meta file is within a block so
 	// its size needs to be calcualted and set upfront.
-	err := setMetaFileSize(meta)
-	if err != nil {
+	if err := setMetaFileSize(meta); err != nil {
 		return err
 	}
 
@@ -265,8 +264,7 @@ func writeMetaFile(logger log.Logger, dir string, meta *BlockMeta) error {
 	}
 
 	var merr tsdb_errors.MultiError
-	_, err = f.Write(jsonMeta)
-	if err != nil {
+	if _, err = f.Write(jsonMeta); err != nil {
 		merr.Add(err)
 		merr.Add(f.Close())
 		return merr.Err()
@@ -275,6 +273,7 @@ func writeMetaFile(logger log.Logger, dir string, meta *BlockMeta) error {
 	// Force the kernel to persist the file on disk to avoid data loss if the host crashes.
 	if err := f.Sync(); err != nil {
 		merr.Add(err)
+		merr.Add(f.Close())
 		return merr.Err()
 	}
 	if err := f.Close(); err != nil {
@@ -285,7 +284,6 @@ func writeMetaFile(logger log.Logger, dir string, meta *BlockMeta) error {
 
 // setMetaFileSize calculates and set the meta file size.
 func setMetaFileSize(meta *BlockMeta) error {
-
 	jsonMeta, err := json.MarshalIndent(meta, "", "\t")
 	if err != nil {
 		return err
--- a/block.go
+++ b/block.go
@@ -20,6 +20,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strconv"
 	"sync"
 
 	"github.com/go-kit/kit/log"
@@ -179,11 +180,14 @@ type BlockMeta struct {
 
 // BlockStats contains stats about contents of a block.
 type BlockStats struct {
-	NumSamples    uint64 `json:"numSamples,omitempty"`
-	NumSeries     uint64 `json:"numSeries,omitempty"`
-	NumChunks     uint64 `json:"numChunks,omitempty"`
-	NumTombstones uint64 `json:"numTombstones,omitempty"`
-	NumBytes      int64  `json:"numBytes,omitempty"`
+	NumSamples        uint64 `json:"numSamples,omitempty"`
+	NumSeries         uint64 `json:"numSeries,omitempty"`
+	NumChunks         uint64 `json:"numChunks,omitempty"`
+	NumTombstones     uint64 `json:"numTombstones,omitempty"`
+	NumBytesChunks    int64  `json:"numBytesChunks"`
+	NumBytesIndex     int64  `json:"numBytesIndex"`
+	NumBytesMeta      int64  `json:"numBytesMeta"`
+	NumBytesTombstone int64  `json:"numBytesTombstone"`
 }
 
 // BlockDesc describes a block by ULID and time range.
@@ -234,6 +238,13 @@ func readMetaFile(dir string) (*BlockMeta, error) {
 func writeMetaFile(logger log.Logger, dir string, meta *BlockMeta) error {
 	meta.Version = 1
 
+	// The meta file is within a block so
+	// its size needs to be calcualted and set upfront.
+	err := setMetaFileSize(meta)
+	if err != nil {
+		return err
+	}
+
 	// Make any changes to the file appear atomic.
 	path := filepath.Join(dir, metaFilename)
 	tmp := path + ".tmp"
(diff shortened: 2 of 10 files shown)


block.go
```diff
@@ -267,6 +283,20 @@ func writeMetaFile(logger log.Logger, dir string, meta *BlockMeta) error {
 	return fileutil.Replace(tmp, path)
 }
 
+// setMetaFileSize calculates and set the meta file size.
+func setMetaFileSize(meta *BlockMeta) error {
+
```
codesome:
```suggestion
```

tombstones.go
```diff
@@ -94,32 +96,34 @@ func writeTombstoneFile(logger log.Logger, dir string, tr TombstoneReader) error
...
+			n, err = mw.Write(buf.Get())
 			if err != nil {
 				return err
 			}
+			size += n
 		}
 		return nil
 	}); err != nil {
-		return fmt.Errorf("error writing tombstones: %v", err)
+		return int64(n), fmt.Errorf("error writing tombstones: %v", err)
```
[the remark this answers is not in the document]
krasi-georgiev: done

bwplotka:
Taking look now (:

krasi-georgiev:
I don't see any problem in that as when opening the db it will re-write the meta to the new version because of this check
https://github.com/prometheus/tsdb/pull/637/files#diff-4c74d030dafa9036e73f63926059db29R363

block.go
```diff
@@ -179,11 +180,14 @@ type BlockMeta struct {
...
-	NumSamples    uint64 `json:"numSamples,omitempty"`
-	NumSeries     uint64 `json:"numSeries,omitempty"`
-	NumChunks     uint64 `json:"numChunks,omitempty"`
-	NumTombstones uint64 `json:"numTombstones,omitempty"`
-	NumBytes      int64  `json:"numBytes,omitempty"`
+	NumSamples        uint64 `json:"numSamples,omitempty"`
+	NumSeries         uint64 `json:"numSeries,omitempty"`
+	NumChunks         uint64 `json:"numChunks,omitempty"`
+	NumTombstones     uint64 `json:"numTombstones,omitempty"`
+	NumBytesChunks    int64  `json:"numBytesChunks"`
```
bwplotka: They are lacking `omitempty` intentionally? To have known bytes even if the value is 0 right? Wonder if we really want to break consistency here.

Also also why we are breaking consistency in types? I would vote for `uint64` as we don't expect negative values, right?

block.go
```diff
@@ -325,13 +353,26 @@ func OpenBlock(logger log.Logger, dir string, pool chunkenc.Pool) (pb *Block, er
...
-	// that would be the logical place for a block size to be calculated.
-	bs := blockSize(cr, ir, tsr)
-	meta.Stats.NumBytes = bs
-	err = writeMetaFile(logger, dir, meta)
-	if err != nil {
-		level.Warn(logger).Log("msg", "couldn't write the meta file for the block size", "block", dir, "err", err)
+	// Calculate the block size.
+	// This is done at read and not when creating a block to avoid blocks return zero size.
+	// For example opening a block created by an older library version wouldn't include the size information.
+	{
```
bwplotka: Why this ? Can we put `if` here? No need for this separation IMO.

block.go
```diff
@@ -325,13 +353,26 @@ func OpenBlock(logger log.Logger, dir string, pool chunkenc.Pool) (pb *Block, er
...
+		// Rewrite the meta only when the sizes has changed.
+		// This check is to allow opening the db in read only mode without
+		// any meta file rewrites when the blocks size hasn't changed.
+		if meta.Stats.NumBytesChunks != cr.Size() ||
```
bwplotka: Do we need this complex check? Can we just check `meta.Stats.NumBytesChunk == 0`? Or `meta.Stats.NumBytes != 0` (if we would keep this field) or `meta.Version == 1` (if we would change meta version)

block.go
```diff
@@ -325,13 +353,26 @@ func OpenBlock(logger log.Logger, dir string, pool chunkenc.Pool) (pb *Block, er
...
+			meta.Stats.NumBytesIndex != ir.Size() ||
+			meta.Stats.NumBytesTombstone != tsr.Size() {
+
+			meta.Stats.NumBytesChunks = cr.Size()
+			meta.Stats.NumBytesIndex = ir.Size()
+			meta.Stats.NumBytesTombstone = tsr.Size()
+
+			err = writeMetaFile(logger, dir, meta)
```
bwplotka: `if err := ... ; err != nil` pattern available

krasi-georgiev:
just tested it and works as expected, all meta files are rewritten to included the correct block size information for each piece.

block.go
bwplotka: Other question. Can we remove all `NumBytes` things at all? Reasons in review summary.

block.go
bwplotka: Also can we stop rewritting meta at all? We already use the `.Size` anyway.

block.go
```diff
@@ -267,6 +282,19 @@ func writeMetaFile(logger log.Logger, dir string, meta *BlockMeta) error {
 	return fileutil.Replace(tmp, path)
 }
 
+// setMetaFileSize calculates and set the meta file size.
+func setMetaFileSize(meta *BlockMeta) error {
```
bwplotka: We can complete remove this if we would always dynamically compute the sizes.

brian-brazil:
By chance today's blog post uses numBytes. Populating it correctly on compaction seems like it would make more sense.

codesome:
> So my question is: Do we really need to bake num of bytes in meta.json even though we calculate precise number in OpenBlock anyway? (even just to check if we have to rewritte)!

This makes me think that we don't need any size info in the meta at all for size based retention.

bwplotka:
> This makes me think that we don't need any size info in the meta at all for size based retention.

Indeed. However if we already have docs for it I wonder what's the use case we described for `NumBytes` there. Can you link the post @brian-brazil here?

krasi-georgiev:
> By chance today's blog post uses numBytes. Populating it correctly on compaction seems like it would make more sense.

I already considered this option(writing the meta size info at compaction when creating a block), but there would be blocks that would never be touched by compaction(e.g. blocks that are too big already).

> This makes me think that we don't need any size info in the meta at all for size based retention.

I considered this option as well, but using the meta simplifies the implementation quite a bit.



btw  I think everyone should focus on block.Size() rather than the actual meta format.

bwplotka:
Yes but the meta json format takes the most lines of this PR and the most questions.

How keeping this size in meta helps in implementation? we do .Size of everything on every open block anyway in this PR (:

Looking on article: https://www.robustperception.io/how-much-disk-space-do-prometheus-blocks-use

It seems like use case is to tell block sizes. I would say log line with calculated block size on compaction would be enough for this use case.

I would vote for removing any num size from stars and just calculating on open block.

krasi-georgiev:
> I would vote for removing any num size from stars and just calculating on open block.

That would probably work as well, but I thought that having it in the actual meta would simplify debugging or some other use cases where people would want to know the block size just by looking at the block meta.

codesome:
> having it in the actual meta would simplify debugging or some other use cases where people would want to know the block size just by looking at the block meta.

I thought of this, but then a simple `du` command would also be enough.

krasi-georgiev:
> I thought of this, but then a simple du command would also be enough.

yep, ok I will refactor to NOT use the meta file and lets see how it looks than.

bwplotka:
Agree with @codesome - it's really easy to see this using linux tooling / object storage meta or whatever else. If any, we can discuss it in another PRs, right? (:

krasi-georgiev:
new version ready.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## master / unreleased
  - [FEATURE] Provide option to compress WAL records using Snappy. [#609](https://github.com/prometheus/tsdb/pull/609)
+ - [BUGFIX] Re-calculate block size when calling `block.Delete`.
+ - [CHANGE] `BlockStats` no longer holds size information and this is kept in memory.
```
bwplotka:
```suggestion
 - [CHANGE] `BlockStats` no longer holds (not precise) size information. This is now dynamically calculate while opening a block in memory.
```

block.go
```diff
@@ -285,6 +284,11 @@ type Block struct {
 	tombstones TombstoneReader
 
 	logger log.Logger
+
+	numBytesChunks    int64
```
bwplotka: `uint64` and separate struct maybe for those? Not a blocker.

block.go
```diff
@@ -437,7 +428,9 @@ func (pb *Block) GetSymbolTableSize() uint64 {
 
 func (pb *Block) setCompactionFailed() error {
 	pb.meta.Compaction.Failed = true
-	return writeMetaFile(pb.logger, pb.dir, &pb.meta)
+	n, err := writeMetaFile(pb.logger, pb.dir, &pb.meta)
+	pb.numBytesMeta = n
```
bwplotka: Slight inconsistency with below code. We update size only on no error, here we do it no matter what. Can we do it on no error here as well? (: this will avoid potential confusion.

compact.go
```diff
@@ -420,12 +420,14 @@ func (c *LeveledCompactor) Compact(dest string, dirs []string, open []*Block) (u
...
 				b.meta.Compaction.Deletable = true
-				if err = writeMetaFile(c.logger, b.dir, &b.meta); err != nil {
+				n, err := writeMetaFile(c.logger, b.dir, &b.meta)
+				if err != nil {
 					level.Error(c.logger).Log(
 						"msg", "Failed to write 'Deletable' to meta file after compaction",
 						"ulid", b.meta.ULID,
 					)
 				}
+				b.numBytesMeta = n
```
bwplotka: Why we set numBytes to new block, but we ignore the other sizes later on?

block.go
```diff
@@ -561,10 +554,17 @@ Outer:
 	pb.tombstones = stones
 	pb.meta.Stats.NumTombstones = pb.tombstones.Total()
 
-	if err := writeTombstoneFile(pb.logger, pb.dir, pb.tombstones); err != nil {
+	n, err := writeTombstoneFile(pb.logger, pb.dir, pb.tombstones)
```
bwplotka: I think we might be able to avoid calculating size on write for both meta and tombstone if we just read both at the end. But this might be even more confusing, so not sure. Happy with this for now.

CHANGELOG.md
krasi-georgiev: the size is recalculated NOT just when opening a block, but also when calling block.Delete so I prefer to keep this a bit more generic.

bwplotka: Fair. Can we have then first fix `(not precise)` and then `This is now dynamically calculated and kept in memory.`

block.go
krasi-georgiev: updated

block.go
krasi-georgiev: yeah I think it will be confusing as well. will refactor if needed at some point.

compact.go
krasi-georgiev: actually this is setting the numBytesMeta for the existing blocks. Existing blocks are already open so need to set the size information when it changes.

Later on it is creating new blocks and these would have their size calculated when they are open at 'db.reload()'.

CHANGELOG.md
krasi-georgiev: what do you mean by `(not precise)'? If that is the missing meta size that it can be something like:
```
- [CHANGE] `BlockStats` no longer holds size information. This is now dynamically calculated and kept in memory. It also includes the meta file size which was not included before.
```

block.go
krasi-georgiev: all Size,Write interfaces I have seen so far use int or int64 so I prefer to keep it as is.
Also for `db.opts.MaxBytes` we also use `int64` where `-1` means that the size retention is disabled.

CHANGELOG.md
bwplotka: perfect!

block.go
bwplotka: Ok, fine by me then

bwplotka: Thanks for explaining

block.go
```diff
@@ -437,7 +428,13 @@ func (pb *Block) GetSymbolTableSize() uint64 {
 
 func (pb *Block) setCompactionFailed() error {
 	pb.meta.Compaction.Failed = true
-	return writeMetaFile(pb.logger, pb.dir, &pb.meta)
+	n, err := writeMetaFile(pb.logger, pb.dir, &pb.meta)
+	if err != nil {
+		pb.numBytesMeta = 0
```
bwplotka: Not really -> I would just not set anything. (:

compact.go
bwplotka: Good point, LGTM.

krasi-georgiev:
@brian-brazil This one is needed to unblock the db read only PR so will merge with the current implementation, but feel free to try and refactor if you have a different idea.
Once we have the read only db working and the tests for it it should be easier to refactor and make sure it doesn't brake the read only db mode.

PR #639, opened 2019-06-24
compact: Rewrite dynamic chunks on compaction.

bwplotka:
Fixes: https://github.com/prometheus/prometheus/issues/5105

This wild change with what I understand with recent changes on memSeries vs compaction vs snapshotting (: Let me know if that makes sense. 

cc @brian-brazil 

### DIFF
--- a/compact.go
+++ b/compact.go
@@ -757,9 +757,14 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta, 
 		}
 
 		for i, chk := range chks {
-			if chk.MinTime < meta.MinTime || chk.MaxTime > meta.MaxTime {
+			if chk.MinTime < meta.MinTime {
 				return errors.Errorf("found chunk with minTime: %d maxTime: %d outside of compacted minTime: %d maxTime: %d",
 					chk.MinTime, chk.MaxTime, meta.MinTime, meta.MaxTime)
+			}
+
+			// Block range is max exclusive, chunk min and max are inclusive, so chk.max == meta.max is wrong as well.
+			if chk.MaxTime >= meta.MaxTime {
+				dranges = append(dranges, Interval{Mint: meta.MaxTime, Maxt: math.MaxInt64})
 			}
 
 			if len(dranges) > 0 {
@@ -767,6 +772,7 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta, 
 				if !chk.OverlapsClosedInterval(dranges[0].Mint, dranges[len(dranges)-1].Maxt) {
 					continue
 				}
+
 				newChunk := chunkenc.NewXORChunk()
 				app, err := newChunk.Appender()
 				if err != nil {
--- a/db.go
+++ b/db.go
@@ -940,11 +940,31 @@ func (db *DB) Snapshot(dir string, withHead bool) error {
 			return errors.Wrapf(err, "error snapshotting block: %s", b.Dir())
 		}
 	}
+
 	if !withHead {
 		return nil
 	}
-	_, err := db.compactor.Write(dir, db.head, db.head.MinTime(), db.head.MaxTime(), nil)
-	return errors.Wrap(err, "snapshot head block")
+
+	mint := db.head.MinTime()
+	maxt := db.head.MaxTime()
+
+	// Wrap head into a range that bounds all reads to it.
+	// Otherwise we can have race between appends and head.Chunks() that takes all it has.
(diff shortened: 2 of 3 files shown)


db.go
```diff
@@ -940,11 +940,31 @@ func (db *DB) Snapshot(dir string, withHead bool) error {
...
 		return nil
 	}
-	_, err := db.compactor.Write(dir, db.head, db.head.MinTime(), db.head.MaxTime(), nil)
-	return errors.Wrap(err, "snapshot head block")
+
+	mint := db.head.MinTime()
+	maxt := db.head.MaxTime()
+
+	// Wrap head into a range that bounds all reads to it.
+	// Otherwise we can have race between appends and head.Chunks() that takes all it has.
```
brian-brazil: This doesn't eliminate that race, as a single chunk can be extended beyond the max.

bwplotka: Yes  yes - it's leftover comment, sorry. I found it just more explicit to wrap it with range - it will at least not fetch chunks that starts after maxTime of the moment of invoking `Write`. The actual race is solved on `populateBlock` with injecting extra `drange`

compact.go
```diff
@@ -757,16 +757,22 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
...
 
 		for i, chk := range chks {
-			if chk.MinTime < meta.MinTime || chk.MaxTime > meta.MaxTime {
+			if chk.MinTime < meta.MinTime {
 				return errors.Errorf("found chunk with minTime: %d maxTime: %d outside of compacted minTime: %d maxTime: %d",
 					chk.MinTime, chk.MaxTime, meta.MinTime, meta.MaxTime)
 			}
 
+			// Block range is max exclusive, chunk min and max are inclusive, so chk.max == meta.max is wrong as well.
+			if chk.MaxTime >= meta.MaxTime {
```
brian-brazil: This loses us the sanity check for on-disk blocks.

A comment as to that this is only for the head block would also be good.

brian-brazil:
Thinking a bit, we'll still want the unit64 idea for the head chunk that you suggested elsewhere. Otherwise the head chunk may be in range when we're pulling chunks, but an out of range sample could be added before we write it out.

krasi-georgiev:
I can imagine the following scenario:

`lset, chks, dranges := set.At()`
returns chunks that are `chk.MaxTime < meta.MaxTime` so they are not added any delete intervals and are not rewritten.
After this check an active appender writes to those chunks.
Than `chunkw.WriteChunks(mergedChks...)` calls the `chunk.Bytes()` method and it will still return corrupt data or samples that are outside the requested range.

That is why I think that the safest way is to ignore chunks that are still being appended to. Like in https://github.com/prometheus/tsdb/pull/641

brian-brazil:
That could lead to ignoring different lengths of time for different series, which is not a useful snapshot.

compact.go
krasi-georgiev: > This loses us the sanity check for on-disk blocks.

what do you mean by this?

bwplotka: @brian-brazil means that we want to rewrite chunks only for head chunks. 

For any other chunks this sanity check would be nice to have as it's an error case in any other case than head chunk.

bwplotka:
> Thinking a bit, we'll still want the unit64 idea for the head chunk that you suggested elsewhere. Otherwise the head chunk may be in range when we're pulling chunks, but an out of range sample could be added before we write it 

Yup you are right @brian-brazil also it is exactly the same thing @krasi-georgiev explains in `I think there is still a race`.

Ok so:
* let's add MaxTime = unlimited for open head chunks
* let's rewrite chunks dynamically on `MaxtTime > meta.MaxTime`

This drop sanity check for other blocks then head, but at least we are safe on dynamic, head chunks.

I am now double checking if we can rewrite *only* on ` Maxt Time == math.MaxInt64` I intiially thought it can is not enough, looking now.

brian-brazil:
> I am now double checking if we can rewrite only on Maxt Time == math.MaxInt64 I intiially thought it is not really enough, looking now.

That should cover it.

krasi-georgiev:
I did merge both PRs in mine so feel free to borrow the code from my PR or if you prefer can continue the work there.

bwplotka:
OK, seems like decision is:

* let's add MaxTime = max for open head chunks
* let's rewrite chunks dynamically on `MaxtTime == meta.MaxTime`

@krasi-georgiev is taking this over, and will adjust his PR, closing this one. (: Thanks @krasi-georgiev !

Issue #518, opened 2019-02-06
Snapshots with head block are unsafe


gouthamve:
Initial review of https://github.com/prometheus/tsdb/pull/514/ made me realise how snapshots can produce chunks of arbitrary bytes if head block is included.

When we write out the headblock here: https://github.com/prometheus/tsdb/blob/master/db.go#L883, we use the `Chunks` inside head here: https://github.com/prometheus/tsdb/blob/master/compact.go#L655

The chunks returned are [`safeChunk`](https://github.com/prometheus/tsdb/blob/master/head.go#L1020-L1024)s that are safe for iteration but not for [writing out](https://github.com/prometheus/tsdb/blob/master/compact.go#L689) as the bytes returned would be changing as the chunks are being appended to.

This means that the current headblock snapshots could be completely broken and might even have "wrong" data. 

I don't have a good solution for this now, but should we disable snapshots with `Head` for the time being? @krasi-georgiev @bwplotka

gouthamve:
When we call snapshot, just cutting a new chunk for every series would fix this imo. That is the most performant way I can think of.

Another way would be re-write open chunks on Snapshot which would mean millions of chunks to be re-written, just not sure how bad it will be allocation and CPU wise. Needs to be benchmarked.

bwplotka:
Yup, wonder what would be performance of that rewrite of open chunks.

bwplotka:
But disabling sounds bad as well, wonder how it worked before chunk boundary check? I guess it worked fine with race right? (:

gouthamve:
I might be being too pedantic, but I would think it "never" worked. The snapshots could literally be persisting a wrong sample and we would never know it.

bwplotka:
Agree.

I really think iterating over open chunks won't be that bad. And we don't need super fast snapshots withHead (:

bboreham:
Is this similar to the problem I found at https://github.com/cortexproject/cortex/pull/1221#issuecomment-462499977 - marshalling doesn't save enough info to be able to add to the chunk afterwards ?

(PS: when creating a Github link to code, you can press the `y` key to get a specific commit rather than `master`, so that people who follow the link later will know which line you meant)

krasi-georgiev:
sounds different, 
this one is getting the raw bytes (Chunk.Bytes()) from the last chunk while being appended to.

PR #641, opened 2019-06-27
Re-encode chunks that are still being appended to when snapshoting.

krasi-georgiev:
fixes: https://github.com/prometheus/tsdb/issues/518
Chunks that are still being appended to might return corrupted data when
calling `chunk.Bytes()` that is why incomplete chunks should never be
persisted. This is relevant mainly when snapshotting with the head
included.

The current snapshot test is sufficient to ensure that incomplete chunks(being appended to) are ignored.

Had this idea while reviewing https://github.com/prometheus/tsdb/pull/639

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## master / unreleased
  - [FEATURE] Provide option to compress WAL records using Snappy. [#609](https://github.com/prometheus/tsdb/pull/609)
  - [BUGFIX] Re-calculate block size when calling `block.Delete`.
- - [BUGFIX] Re-encode all chunks at compaction that are open(being appended to). This avoids writing out corrupt data. It happens when snapshotting with the head included.
+ - [BUGFIX] Re-encode all head chunks at compaction that are open (being appended to) or outside the Maxt block range. This avoids writing out corrupt data. It happens when snapshotting with the head included.
  - [CHANGE] The meta file `BlockStats` no longer holds size information. This is now dynamically calculated and kept in memory. It also includes the meta file size which was not included before.
 
 ## 0.8.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 ## master / unreleased
  - [FEATURE] Provide option to compress WAL records using Snappy. [#609](https://github.com/prometheus/tsdb/pull/609)
  - [BUGFIX] Re-calculate block size when calling `block.Delete`.
-- [CHANGE] The meta file `BlockStats` no longer holds size information. This is now dynamically calculated and kept in memory. It also includes the meta file size which was not included before.
+ - [BUGFIX] Re-encode all chunks at compaction that are open(being appended to). This avoids writing out corrupt data. It happens when snapshotting with the head included.
+ - [CHANGE] The meta file `BlockStats` no longer holds size information. This is now dynamically calculated and kept in memory. It also includes the meta file size which was not included before.
 
 ## 0.8.0
  - [BUGFIX] Calling `Close` more than once on a querier returns an error instead of a panic.
--- a/compact.go
+++ b/compact.go
@@ -757,12 +757,13 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta, 
 		}
 
 		for i, chk := range chks {
-			// Re-encode chunks that are still open(being written to).
-			// The chunk.Bytes() method is not safe for open chunks hence the re-encoding.
-			// This happens only when snapshotting the head block.
-			if chk.MaxTime == math.MaxInt64 {
+			// Re-encode head chunks that are still open(being written to) or
+			// outside the compacted MaxTime range.
+			// The chunk.Bytes() method is not safe for these chunks hence the re-encoding.
+			// This happens when snapshotting the head block.
+			if _, isHeadChunk := chk.Chunk.(*safeChunk); isHeadChunk && (chk.MaxTime > meta.MaxTime || chk.MaxTime == math.MaxInt64) {
 				dranges = append(dranges, Interval{Mint: meta.MaxTime, Maxt: math.MaxInt64})
-			} else if chk.MinTime < meta.MinTime || chk.MaxTime > meta.MaxTime {
+			} else if chk.MinTime < meta.MinTime || chk.MaxTime > meta.MaxTime { // Sanity check for disk blocks.
 				return errors.Errorf("found chunk with minTime: %d maxTime: %d outside of compacted minTime: %d maxTime: %d",
 					chk.MinTime, chk.MaxTime, meta.MinTime, meta.MaxTime)
 			}
@@ -779,12 +780,23 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta, 
 				}
 
 				it := &deletedIterator{it: chk.Chunk.Iterator(), intervals: dranges}
+
+				var (
+					min = int64(math.MinInt64)
+					t   int64
+					v   float64
+				)
 				for it.Next() {
-					ts, v := it.At()
-					app.Append(ts, v)
+					t, v = it.At()
+					app.Append(t, v)
+					if min == math.MinInt64 {
+						min = t
+					}
 				}
 
 				chks[i].Chunk = newChunk
+				chks[i].MaxTime = t
+				chks[i].MinTime = min
 			}
 		}
 
--- a/compact.go
+++ b/compact.go
@@ -757,15 +757,22 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta, 
 		}
 
 		for i, chk := range chks {
-			// Re-encode head chunks that are still open(being written to) or
+			// Re-encode head chunks that are still open (being appended to) or
 			// outside the compacted MaxTime range.
-			// The chunk.Bytes() method is not safe for these chunks hence the re-encoding.
+			// The chunk.Bytes() method is not safe for open chunks hence the re-encoding.
 			// This happens when snapshotting the head block.
-			if _, isHeadChunk := chk.Chunk.(*safeChunk); isHeadChunk && (chk.MaxTime > meta.MaxTime || chk.MaxTime == math.MaxInt64) {
+			//
+			// Block time range is half-open: [meta.MinTime, meta.MaxTime) and
+			// chunks are closed hence the chk.MaxTime >= meta.MaxTime check.
+			if _, isHeadChunk := chk.Chunk.(*safeChunk); isHeadChunk && chk.MaxTime >= meta.MaxTime {
 				dranges = append(dranges, Interval{Mint: meta.MaxTime, Maxt: math.MaxInt64})
 			} else if chk.MinTime < meta.MinTime || chk.MaxTime > meta.MaxTime { // Sanity check for disk blocks.
 				return errors.Errorf("found chunk with minTime: %d maxTime: %d outside of compacted minTime: %d maxTime: %d",
 					chk.MinTime, chk.MaxTime, meta.MinTime, meta.MaxTime)
+			} else if chk.MaxTime == meta.MaxTime { // This shouldn't happen as well, but will brake many users so keeping it as a warning only.
+				level.Warn(c.logger).Log("msg", "found chunk at max time boundary",
+					"chunk minTime", chk.MinTime, "chunk maxTime", chk.MaxTime,
+					"compacted minTime", meta.MinTime, "compacted mmaxTime", meta.MaxTime)
 			}
 
 			if len(dranges) > 0 {
@@ -782,21 +789,19 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta, 
 				it := &deletedIterator{it: chk.Chunk.Iterator(), intervals: dranges}
 
 				var (
-					min = int64(math.MinInt64)
-					t   int64
-					v   float64
+					t int64
+					v float64
 				)
 				for it.Next() {
 					t, v = it.At()
 					app.Append(t, v)
-					if min == math.MinInt64 {
-						min = t
-					}
+				}
+				if err := it.Err(); err != nil {
+					return errors.Wrap(err, "iterate chunk while re-encoding")
 				}
 
 				chks[i].Chunk = newChunk
 				chks[i].MaxTime = t
-				chks[i].MinTime = min
 			}
 		}
 
--- a/compact.go
+++ b/compact.go
@@ -764,15 +764,17 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta, 
 			//
 			// Block time range is half-open: [meta.MinTime, meta.MaxTime) and
 			// chunks are closed hence the chk.MaxTime >= meta.MaxTime check.
+			//
+			// TODO think how to avoid the typecasting to verify when it is head block.
 			if _, isHeadChunk := chk.Chunk.(*safeChunk); isHeadChunk && chk.MaxTime >= meta.MaxTime {
 				dranges = append(dranges, Interval{Mint: meta.MaxTime, Maxt: math.MaxInt64})
-			} else if chk.MinTime < meta.MinTime || chk.MaxTime > meta.MaxTime { // Sanity check for disk blocks.
+
+			} else
+			// Sanity check for disk blocks.
+			// chk.MaxTime == meta.MaxTime shouldn't happen as well, but will brake many users so not checking for that.
+			if chk.MinTime < meta.MinTime || chk.MaxTime > meta.MaxTime {
 				return errors.Errorf("found chunk with minTime: %d maxTime: %d outside of compacted minTime: %d maxTime: %d",
 					chk.MinTime, chk.MaxTime, meta.MinTime, meta.MaxTime)
-			} else if chk.MaxTime == meta.MaxTime { // This shouldn't happen as well, but will brake many users so keeping it as a warning only.
-				level.Warn(c.logger).Log("msg", "found chunk at max time boundary",
-					"chunk minTime", chk.MinTime, "chunk maxTime", chk.MaxTime,
-					"compacted minTime", meta.MinTime, "compacted mmaxTime", meta.MaxTime)
 			}
 
 			if len(dranges) > 0 {
--- a/compact.go
+++ b/compact.go
@@ -757,7 +757,12 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta, 
 		}
 
 		for i, chk := range chks {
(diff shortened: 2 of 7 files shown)


gouthamve:
But this is an incomplete snapshot then? We can basically not persist 2hrs of ingested data with this.

krasi-georgiev:
why 2h? chunk have a maximum of 120 samples.

https://github.com/prometheus/tsdb/blob/ba7f8025a9101663519d34020c283ed8f5ca79ff/head.go#L1695

brian-brazil:
It's not safe to do this, you need a clean break in the data. If a user wanted this they can skip the head block.

krasi-georgiev:
Why not safe? this is skipping only few of the last samples not the entire head.

I looked at the code quite a bit and otherwise `chunk.Bytes()` could always return corrupt data.

brian-brazil:
> Why not safe? this is skipping only few of the last samples not the entire head.

It's randomly skipping different amounts of data for different series.

krasi-georgiev:
I guess we can run some additional check to ensure all chunks are cut at the same time.
So i understand the problem with different timings Can you give an example of a query that will cause issues?

Do you have any other idea to solve that race?

bwplotka:
Yes. Reencode those moving chunks: https://github.com/prometheus/tsdb/pull/639

krasi-georgiev:
Did you see my comment in your pr? I think it still has a race there.

krasi-georgiev:
Maybe we can combine both prs. Which is
To Reincode all incomplete chunks(head chunks)

bwplotka:
I am not saying my PR is perfect. Even CI is saying it's not but my main
focus is on remote read TBH so joining forces would be nice. I can push my
changes to some tsdb Branch for easier collaboration @Krasi (:

codesome:
I had a look at this and #639 , and for some reason I find re-encoding the last chunk safely and using it seems to be a better option than ignoring the last chunk. Reasons would be (1) What @brian-brazil [said](https://github.com/prometheus/tsdb/pull/641#issuecomment-506321857) (2) Snapshot as much data as possible.

#639 still needs to address the race as described by @krasi-georgiev  [here](https://github.com/prometheus/tsdb/pull/639#issuecomment-506300327).

krasi-georgiev:
I have just merged both PRs

bwplotka:
Ok, l might be biased but I think having `MaxTime == max` is bit better way of saying if it's full. This is because it's quite natural and does not requre change in the interface. What do you think about closing this to avoid confusion and trying this approach on https://github.com/prometheus/tsdb/pull/639 ? Or we can start TSDB branch a new PR for better collab.

krasi-georgiev:
ready for another review

CHANGELOG.md
```diff
@@ -1,7 +1,8 @@
 ## master / unreleased
  - [FEATURE] Provide option to compress WAL records using Snappy. [#609](https://github.com/prometheus/tsdb/pull/609)
  - [BUGFIX] Re-calculate block size when calling `block.Delete`.
-- [CHANGE] The meta file `BlockStats` no longer holds size information. This is now dynamically calculated and kept in memory. It also includes the meta file size which was not included before.
+ - [BUGFIX] Re-encode all chunks at compaction that are open(being appended to). This avoids writing out corrupt data. It happens when snapshotting with the head included.
```
bwplotka:
```suggestion
 - [BUGFIX] Re-encode all chunks at compaction that are open (being appended to). This avoids writing out corrupted data when snapshotting with the head included.
```

compact.go
```diff
@@ -757,7 +757,12 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
 		}
 
 		for i, chk := range chks {
-			if chk.MinTime < meta.MinTime || chk.MaxTime > meta.MaxTime {
+			// Re-encode chunks that are still open(being written to).
+			// The chunk.Bytes() method is not safe for open chunks hence the re-encoding.
+			// This happens only when snapshotting the head block.
+			if chk.MaxTime == math.MaxInt64 {
```
bwplotka: Don't we want to check `chk.MinTime < meta.MinTime ` still ? for sanity?

compact.go
```diff
@@ -757,7 +757,12 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
 		}
 
 		for i, chk := range chks {
-			if chk.MinTime < meta.MinTime || chk.MaxTime > meta.MaxTime {
+			// Re-encode chunks that are still open(being written to).
```
bwplotka:
```suggestion
			// Re-encode chunks that are still open (being written to).
```

head.go
```diff
@@ -1298,9 +1298,15 @@ func (h *headIndexReader) Series(ref uint64, lbls *labels.Labels, chks *[]chunks
 		if !c.OverlapsClosedInterval(h.mint, h.maxt) {
 			continue
 		}
+		// Set the head chunks as open(being appended to).
```
bwplotka:
```suggestion
		// Set the head chunks as open (being appended to).
```

head.go
```diff
@@ -1298,9 +1298,15 @@ func (h *headIndexReader) Series(ref uint64, lbls *labels.Labels, chks *[]chunks
...
+		maxTime := c.maxTime
```
bwplotka: Just to check `s.Lock` locks append  /cut to this open chunk here right?

If not, we have a race.

bwplotka:
Another idea would be to pass somehow else that it is a head block and treat those differently in terms of maxTime being outside of desired meta.MaxTime

brian-brazil:
> Another idea would be to pass somehow else that it is a head block (both open and closed ones) and treat those differently in terms of chunk maxTime being outside of desired meta.MaxTime

That's probably easiest at this stage.

krasi-georgiev:
> Another idea would be to pass somehow else that it is a head block (both open and closed ones) and treat those differently in terms of chunk maxTime being outside of desired meta.MaxTime

could you explain a bit more about this idea?

bwplotka:
No worries. Essentially we are now marking `open chunk` in certain way and we rewrite only them. This part makes sense. But to fix edge case we need something more. I propose to let's say:
 
1. leave this marker we have now for `open` blocks. Rewrite those 100%.
2. Add some another way to tell if the chunk is part of head block. Another meta field just for purpose of populate is rather no go (will be hard as it is part of populateBlock and IndexReader interface..), but maybe literally some flag to `compute.write` ? 

Now we will have:
* check if head, open block. If yes rewrite.
* else check if it is outside of meta.Max.
  * if the chunk is part of head and is closed: rewrite
  * anything else: fail.

krasi-georgiev:
hm still a bit unclear, but IIUC you mean check if a chunk belongs to the head and rewrite when it is open or outside `meta.Max` ? for all other chunks just fail when outside `meta.Maxt`

bwplotka:
Exactly (:

compact.go
```diff
@@ -757,7 +757,13 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
 		}
 
 		for i, chk := range chks {
-			if chk.MinTime < meta.MinTime || chk.MaxTime > meta.MaxTime {
+			// Re-encode head chunks that are still open(being written to) or
+			// outside the compacted MaxTime range.
+			// The chunk.Bytes() method is not safe for these chunks hence the re-encoding.
```
bwplotka: It's not safe only for open blocks TBH

compact.go
```diff
@@ -757,7 +757,13 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
...
+			// This happens when snapshotting the head block.
+			if _, isHeadChunk := chk.Chunk.(*safeChunk); isHeadChunk && (chk.MaxTime > meta.MaxTime || chk.MaxTime == math.MaxInt64) {
```
bwplotka: I think you don't need == math.MaxInt because MaxInt will be definitely bigger than `meta.MaxtTime` (:
```suggestion
			if _, isHeadChunk := chk.Chunk.(*safeChunk); isHeadChunk && (chk.MaxTime > meta.MaxTime) {
```

compact.go
```diff
@@ -774,12 +780,23 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
...
+					t, v = it.At()
+					app.Append(t, v)
+					if min == math.MinInt64 {
+						min = t
+					}
 				}
 
 				chks[i].Chunk = newChunk
+				chks[i].MaxTime = t
+				chks[i].MinTime = min
```
bwplotka: Why we need this? This chunk has right minTime for sure, right?

compact.go
```diff
@@ -774,12 +780,23 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
 				}
 
 				it := &deletedIterator{it: chk.Chunk.Iterator(), intervals: dranges}
+
+				var (
+					min = int64(math.MinInt64)
+					t   int64
+					v   float64
+				)
 				for it.Next() {
```
bwplotka: Can we check `it.Err` just to follow good patterns (even if current implementaiton does return error)

compact.go
```diff
@@ -757,7 +757,13 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
 		}
 
 		for i, chk := range chks {
-			if chk.MinTime < meta.MinTime || chk.MaxTime > meta.MaxTime {
+			// Re-encode head chunks that are still open (being appended to) or
+			// outside the compacted MaxTime range.
+			// The chunk.Bytes() method is not safe for open chunks hence the re-encoding.
+			// This happens when snapshotting the head block.
+			if _, isHeadChunk := chk.Chunk.(*safeChunk); isHeadChunk && chk.MaxTime > meta.MaxTime {
```
bwplotka: Don't we want `>=` here? And same for sanity check? We don't expect meta.MaxTime sample timestamp in chunk, right?

krasi-georgiev: updated. Thanks

compact.go
```diff
@@ -757,7 +757,16 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
 		}
 
 		for i, chk := range chks {
-			if chk.MinTime < meta.MinTime || chk.MaxTime > meta.MaxTime {
+			// Re-encode head chunks that are still open (being appended to) or
+			// outside the compacted MaxTime range.
+			// The chunk.Bytes() method is not safe for open chunks hence the re-encoding.
+			// This happens when snapshotting the head block.
+			//
+			// Bock time range is half-open: [meta.MinTime, meta.MaxTime) and
```
bwplotka:
```suggestion
			// Block time range is half-open: [meta.MinTime, meta.MaxTime) and
```

compact.go
```diff
@@ -757,7 +757,16 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
...
+			// chunks are closed hence the chk.MaxTime >= meta.MaxTime not check.
+			if _, isHeadChunk := chk.Chunk.(*safeChunk); isHeadChunk && chk.MaxTime >= meta.MaxTime {
+				dranges = append(dranges, Interval{Mint: meta.MaxTime, Maxt: math.MaxInt64})
+			} else if chk.MinTime < meta.MinTime || chk.MaxTime >= meta.MaxTime { // Sanity check for disk blocks.
```
bwplotka: ` >= ` Wonder if this will not break compaction for old users who might have such chunk for some reaons :thinking:

bwplotka: shouldn't but maybe warning would be nice instead here? (sorry it was my idea to put it)

bwplotka: Especially that there might not much harm in leaving just `>`

head.go
krasi-georgiev: yes, I just double checked and the lock is fine

compact.go
krasi-georgiev: updated to be only a warning.

codesome: Why are we not returning error here? Wouldn't this mean we would allow chunks outside the range?

brian-brazil: We should still have an error here

compact.go
```diff
@@ -774,12 +784,21 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
 				}
 
 				it := &deletedIterator{it: chk.Chunk.Iterator(), intervals: dranges}
+
+				var (
```
brian-brazil: Why are you doing this? The Go compiler is smart enough to handle this itself

compact.go
```diff
@@ -757,9 +757,19 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
...
-				return errors.Errorf("found chunk with minTime: %d maxTime: %d outside of compacted minTime: %d maxTime: %d",
-					chk.MinTime, chk.MaxTime, meta.MinTime, meta.MaxTime)
+			// Re-encode head chunks that are still open (being appended to) or
+			// outside the compacted MaxTime range.
+			// The chunk.Bytes() method is not safe for open chunks hence the re-encoding.
+			// This happens when snapshotting the head block.
+			//
+			// Block time range is half-open: [meta.MinTime, meta.MaxTime) and
+			// chunks are closed hence the chk.MaxTime >= meta.MaxTime check.
+			if _, isHeadChunk := chk.Chunk.(*safeChunk); isHeadChunk && chk.MaxTime >= meta.MaxTime {
```
brian-brazil: Checking for safeChunk is a bit magic, I'd explicitly pass in a bool to the function.

codesome: +1, also the bool would avoid checking every chunk for a head chunk.

bwplotka: Can agree. but I would rather have separate method for `populateFromHead` instead of bool

compact.go
bwplotka: > shouldn't but maybe warning would be nice instead here? (sorry it was my idea to put it)

I meant warning for `ch.MaxTime == meta.MaxTime` and error for the rest... sorry for confusion - should be more explicit here ):

bwplotka: We should definitely not add regression and not skip the sanity error to what we had previously

krasi-georgiev: but why the hard fail? The next compaction would remove these extra samples from the chunks, no?

compact.go
krasi-georgiev: because we need it to set the chunk meta maxt later
`chks[i].MaxTime = t`

compact.go
codesome: This would lead to a mismatch in what the meta says and what is actually in the block if we allow chunks outside the meta range.

compact.go
krasi-georgiev: I am thinking the same as `populateBlock` is already getting too messy and too long.

bwplotka: so we are waiting on this right?

Also `safeChunk` is not that bad I guess with given comment IMO - same as `http Handler` and casting `http.Flusher` to flush.

krasi-georgiev: yeah I can spend a bit more time, but so far can't see a good clean way to separate `populateBlock` and `populateHead` without duplicating a lot of code.

bwplotka:
Today is release day and this PR is ready with only one nit: `safeChunk` casting to tell if it's chunk block or not. 

I would add `TODO` and otherwise merge it and do TSDB update on Prometheus side. Any attempt of refactoring `populate` will probably take *days* to do and review, so I would fix this nit later.

Thoughts? @brancz @brian-brazil @codesome @krasi-georgiev

compact.go
bwplotka: https://github.com/prometheus/tsdb/pull/641#issuecomment-507986776

codesome:
I ran a quick benchmark which I had ready, and here are the results. And this doesn't have head in compaction.
```
BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     12470779450     12815518004     +2.76%

BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     17818580       24843658       +39.43%

BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     2112751344     2578539664     +22.05%
```

codesome:
Increase in allocs (and also B/op) are alarming for a small change in `populateBlock`

codesome:
And confirmed that it is in `populateBlock`, I am investigating it.

compact.go
```diff
@@ -757,9 +757,22 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
...
+				dranges = append(dranges, Interval{Mint: meta.MaxTime, Maxt: math.MaxInt64})
+			} else if chk.MinTime < meta.MinTime || chk.MaxTime > meta.MaxTime { // Sanity check for disk blocks.
 				return errors.Errorf("found chunk with minTime: %d maxTime: %d outside of compacted minTime: %d maxTime: %d",
 					chk.MinTime, chk.MaxTime, meta.MinTime, meta.MaxTime)
+			} else if chk.MaxTime == meta.MaxTime { // This shouldn't happen as well, but will brake many users so keeping it as a warning only.
+				level.Warn(c.logger).Log("msg", "found chunk at max time boundary",
```
codesome: And the culprit for increase allocs is this warning. I am not sure why we need warning here, we used to allow it before. When compacting 2 blocks, it is possible to have chunk MaxTime matching the meta.MaxTime. And given how many chunks there might be in a block, this will lead to 10s of thousands (or millions?) of warnings per compaction.

bwplotka: I mean it would be probably enough to just check for warnings in unit test (:

bwplotka:
Great check @codesome thanks, let's remove this warning then @krasi-georgiev

compact.go
```diff
@@ -757,6 +757,21 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, meta *BlockMeta,
...
+			// Block time range is half-open: [meta.MinTime, meta.MaxTime) and
+			// chunks are closed hence the chk.MaxTime >= meta.MaxTime check.
+			//
+			// TODO think how to avoid the typecasting to verify when it is head block.
+			if _, isHeadChunk := chk.Chunk.(*safeChunk); isHeadChunk && chk.MaxTime >= meta.MaxTime {
+				dranges = append(dranges, Interval{Mint: meta.MaxTime, Maxt: math.MaxInt64})
+
+			} else
+			// Sanity check for disk blocks.
+			// chk.MaxTime == meta.MaxTime shouldn't happen as well, but will brake many users so not checking for that.
```
krasi-georgiev: before anyone complains about the comment formatting, I tried all variations and this one is with the most intuitive indentation so far for the  `else if` block

brian-brazil:
:+1:

compact.go
Wsyzxxn: image

populateBlock use long time when thanos compact

krasi-georgiev: It is normal that `populateBlock` takes long time.
Can you be more specific what are showing with this screenshot?

PR #643, opened 2019-07-02
Breakdown generic writeOffsetTable

codesome:
This is a broken down piece of https://github.com/prometheus/tsdb/pull/627

With this, I am avoiding allocating slice headers for postings by having the name and value as separate fields. This saves quite some allocs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,6 @@
 ## Master / unreleased
 
  - [FEATURE] `chunckenc.Chunk.Iterator` method now takes a `chunckenc.Iterator` interface as an argument for reuse.
- - [ENHANCEMENT] `hashEntry` is broken down into `labelIndexHashEntry` and `postingsHashEntry` for optimizing allocs.
-   - `writeOffsetTable` was broken down into `writeLabelIndexesOffsetTable` and `writePostingsOffsetTable` in the process of breaking down `hashEntry`.
 
 ## 0.9.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## Master / unreleased
 
  - [FEATURE] `chunckenc.Chunk.Iterator` method now takes a `chunckenc.Iterator` interface as an argument for reuse.
+ - [ENHANCEMENT] `hashEntry` is broken down into `labelIndexHashEntry` and `postingsHashEntry` for optimizing allocs.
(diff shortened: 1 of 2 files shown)


cstyan:
> This saves quite some allocs.

Any sense of how much? 10%? 20%?

codesome:
I don't have numbers for individual optimizations. All combined numbers is in this comment https://github.com/prometheus/tsdb/pull/627#issuecomment-503453181. I will try to get number for each optimizations.

codesome:
So the gains are mostly allocs.
```
BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     12824719385     12761476707     -0.49%

BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     15845398       14839951       -6.35%

BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     2074569456     2074402112     -0.01%
```

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 ## Master / unreleased
 
  - [FEATURE] `chunckenc.Chunk.Iterator` method now takes a `chunckenc.Iterator` interface as an argument for reuse.
+ - [ENHANCEMENT] `hashEntry` is broken down into `labelIndexHashEntry` and `postingsHashEntry` for optimizing allocs.
```
gouthamve: Need not be mentioned

index/index.go
```diff
@@ -494,8 +513,9 @@ func (w *Writer) WritePostings(name, value string, it Postings) error {
 		return err
 	}
 
-	w.postings = append(w.postings, hashEntry{
-		keys:   []string{name, value},
+	w.postings = append(w.postings, postingsHashEntry{
```
gouthamve: I'm just surprised that this small change reduces allocs so much.

codesome:
> I am not sure the minor improvement in allocs is worth this change.

I would still like to have it. Multiple small improvements will add up to a big gain. And the absolute gain from this PR will be bigger in different scenarios (more number of label-values and series than the benchmark).

index/index.go
bwplotka: Quite interesting indeed, but keep in mind that slice header is like 16 bytes overall, so in some cases might be bigger than `name` or `key` (:

PR #645, opened 2019-07-02
Re-use 'keys' in ReadOffsetTable

codesome:

This saves a whole bunch of allocs when opening a block.

- [ ] Add comment for why the len of `keys` is set to 2

### DIFF
--- a/index/index.go
+++ b/index/index.go
@@ -781,9 +781,10 @@ func ReadOffsetTable(bs ByteSlice, off uint64, f func([]string, uint64) error) e
 	d := encoding.NewDecbufAt(bs, int(off), castagnoliTable)
 	cnt := d.Be32()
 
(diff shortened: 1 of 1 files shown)


index/index.go
```diff
@@ -781,9 +781,10 @@ func ReadOffsetTable(bs ByteSlice, off uint64, f func([]string, uint64) error) e
 	d := encoding.NewDecbufAt(bs, int(off), castagnoliTable)
 	cnt := d.Be32()
 
+	keys := make([]string, 0, 2)
```
codesome: https://github.com/prometheus/tsdb/pull/627#discussion_r299344697 tells why I have set the size to 2.

bwplotka:
I love PR splitting - so much easier to review :heart: @tomwilkie  :trollface:

PR #649, opened 2019-07-03
wal: Inject LiveReader metrics rather than registry

### LINKED ISSUE #646: Cannot run more than one wal.LiveReader with metrics
https://github.com/prometheus/tsdb/pull/619 changed the LiveReader to take a prometheus registerer instead of always using the default registerer. However, now if you try to instantiate more than one LiveReader with the same registerer it will panic.

You can pass nil as the registerer to avoid the panic but that means that metric will no longer show up.

cc: @brancz since I think you are doing the Prometheus release soon. This was breaking Prometheus when I tried to update https://github.com/prometheus/prometheus/pull/5592 in preparation for 2.11.

brancz:
LiveReaders are instantiated `number of remote write queues * segments`
times, which would cause double registration of the metrics. Instead
this should be orchestrated by the layer above, instantiating the live
reader.

Fixes #646 

@bwplotka @cstyan @krasi-georgiev @csmarchbanks

### DIFF
--- a/wal/live_reader.go
+++ b/wal/live_reader.go
@@ -27,24 +27,38 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+// liveReaderMetrics holds all metrics exposed by the LiveReader.
+type liveReaderMetrics struct {
+	readerCorruptionErrors *prometheus.CounterVec
+}
+
+// LiveReaderMetrics instatiates, registers and returns metrics to be injected
+// at LiveReader instantiation.
+func NewLiveReaderMetrics(reg prometheus.Registerer) *liveReaderMetrics {
+	m := &liveReaderMetrics{
+		readerCorruptionErrors: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "prometheus_tsdb_wal_reader_corruption_errors_total",
+			Help: "Errors encountered when reading the WAL.",
+		}, []string{"error"}),
+	}
+
+	if reg != nil {
+		reg.Register(m.readerCorruptionErrors)
+	}
+
+	return m
+}
+
 // NewLiveReader returns a new live reader.
-func NewLiveReader(logger log.Logger, reg prometheus.Registerer, r io.Reader) *LiveReader {
+func NewLiveReader(logger log.Logger, metrics *liveReaderMetrics, r io.Reader) *LiveReader {
 	lr := &LiveReader{
-		logger: logger,
-		rdr:    r,
+		logger:  logger,
+		rdr:     r,
+		metrics: metrics,
 
 		// Until we understand how they come about, make readers permissive
 		// to records spanning pages.
 		permissive: true,
-	}
-
-	lr.readerCorruptionErrors = prometheus.NewCounterVec(prometheus.CounterOpts{
-		Name: "prometheus_tsdb_wal_reader_corruption_errors_total",
-		Help: "Errors encountered when reading the WAL.",
-	}, []string{"error"})
-
-	if reg != nil {
-		reg.MustRegister(lr.readerCorruptionErrors)
 	}
 
 	return lr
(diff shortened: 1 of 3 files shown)


wal/live_reader.go
```diff
@@ -27,26 +27,31 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+type LiveReaderMetrics struct {
+	ReaderCorruptionErrors *prometheus.CounterVec
+}
+
+func NewLiveReaderMetrics() *LiveReaderMetrics {
```
csmarchbanks: Should this take a registry and register all metrics as well? It would be easy for a caller to miss registering one in the future.

wal/live_reader.go
```diff
@@ -27,26 +27,31 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+type LiveReaderMetrics struct {
```
bwplotka: Some comment would be nice I guess

wal/live_reader.go
bwplotka: Yes I thought that is the pattern - to pass registry here in constructor.

wal/live_reader.go
```diff
@@ -27,26 +27,31 @@ import (
...
+	ReaderCorruptionErrors *prometheus.CounterVec
```
brian-brazil: This should probably be private.

wal/live_reader.go
```diff
@@ -27,26 +27,31 @@ import (
...
+		ReaderCorruptionErrors: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "prometheus_tsdb_wal_reader_corruption_errors_total",
+			Help: "Errors encountered when reading the WAL.",
+		}, []string{"error"}),
+	}
+}
+
 // NewLiveReader returns a new live reader.
-func NewLiveReader(logger log.Logger, reg prometheus.Registerer, r io.Reader) *LiveReader {
+func NewLiveReader(logger log.Logger, metrics *LiveReaderMetrics, r io.Reader) *LiveReader {
```
bwplotka: For one metric I would just pass one metric, but up to you (:

wal/live_reader.go
bwplotka: Also it's only one metric.. YAGNI and pass just metric? (:

wal/live_reader.go
csmarchbanks: Personally I like the struct pattern a bit more in order to make it easy to add future metrics and not change the interface.

brancz: This typically grows, I usually like to put this infrastructure in place right away, but no strong opinion.

bwplotka: Not entirely sold, but happy to proceed as we are in hurry. (:

brancz: I don't think we should skip, just because we're in a rush. Worst case we delay the release, that's fine. Maybe @brian-brazil can give his opinion.

bwplotka:
Can you this item to CHANGELOG?

wal/live_reader.go
brian-brazil: Right now I'd focus on getting the release out, I'd generally go with struct though.

wal/live_reader.go
```diff
@@ -27,26 +27,40 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+// LiveReaderMetrics holds all metrics exposed by the LiveReader.
+type LiveReaderMetrics struct {
+	readerCorruptionErrors *prometheus.CounterVec
+}
+
+// LiveReaderMetrics instatiates, register and return metrics to be injected at
```
csmarchbanks: Nit: I think "registers and returns" would be grammatically correct.

wal/live_reader.go
```diff
@@ -27,26 +27,40 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+// liveReaderMetrics holds all metrics exposed by the LiveReader.
+type liveReaderMetrics struct {
```
csmarchbanks: Doesn't the struct need to be public, and the metrics private? So that the struct can be passed around in Prometheus?

bwplotka: It does not need to be so all good.

However go style guide says it's antipattern (:

PR #653, opened 2019-07-05
Reuse byte buffer in WriteChunks and writeHash

codesome:

- [x] CHANGELOG entry

### DIFF
--- a/chunks/chunks.go
+++ b/chunks/chunks.go
@@ -248,8 +248,8 @@ func MergeChunks(a, b chunkenc.Chunk) (*chunkenc.XORChunk, error) {
 	if err != nil {
 		return nil, err
 	}
-	ait := a.Iterator()
-	bit := b.Iterator()
+	ait := a.Iterator(nil)
+	bit := b.Iterator(nil)
 	aok, bok := ait.Next(), bit.Next()
 	for aok && bok {
 		at, av := ait.At()
--- a/chunks/chunks.go
+++ b/chunks/chunks.go
@@ -59,7 +59,7 @@ type Meta struct {
 // writeHash writes the chunk encoding and raw data into the provided hash.
 // 'buf' byte slice is used to avoid allocating new byte slice. Assumes that len(buf) is 0.
 func (cm *Meta) writeHash(h hash.Hash, buf []byte) error {
-	buf = append(buf, byte(cm.Chunk.Encoding()))
+	buf = append(buf[:0], byte(cm.Chunk.Encoding()))
 	if _, err := h.Write(buf[:1]); err != nil {
 		return err
 	}
@@ -99,6 +99,7 @@ type Writer struct {
 	wbuf    *bufio.Writer
 	n       int64
 	crc32   hash.Hash
+	buf     [binary.MaxVarintLen32]byte
 
 	segmentSize int64
 }
@@ -284,7 +285,7 @@ func MergeChunks(a, b chunkenc.Chunk) (*chunkenc.XORChunk, error) {
 	return newChunk, nil
 }
 
-func (w *Writer) WriteChunks(b []byte, chks ...Meta) ([]byte, error) {
+func (w *Writer) WriteChunks(chks ...Meta) error {
 	// Calculate maximum space we need and cut a new segment in case
 	// we don't fit into the current one.
 	maxLen := int64(binary.MaxVarintLen32) // The number of chunks.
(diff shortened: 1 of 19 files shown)


codesome:
Here are the benchmark results (ns/op diff fluctuates, it's not always negative)
```
BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     13072562459     13007510801     -0.50%

BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     15860945       12854710       -18.95%

BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     2075024560     2063998112     -0.53%
```

chunks/chunks.go
```diff
@@ -57,8 +57,10 @@ type Meta struct {
 }
 
 // writeHash writes the chunk encoding and raw data into the provided hash.
-func (cm *Meta) writeHash(h hash.Hash) error {
-	if _, err := h.Write([]byte{byte(cm.Chunk.Encoding())}); err != nil {
+// 'buf' byte slice is used to avoid allocating new byte slice. Assumes that len(buf) is 0.
+func (cm *Meta) writeHash(h hash.Hash, buf []byte) error {
+	buf = append(buf, byte(cm.Chunk.Encoding()))
```
csmarchbanks: Instead of assuming the buffer always has zero length, you could explicitly use `buf[:0]`.

codesome: As it was internal function, I just assumed it will be passed with 0 len. I will change it to `buf[:0]`.

codesome:
> Rather than changing the interface, would it make sense to add the buffer to the Writer struct, and then initialize it to the correct size in NewWriter?

That is a good idea. I will make the change.

chunks/chunks.go
```diff
@@ -57,8 +57,10 @@ type Meta struct {
 }
 
 // writeHash writes the chunk encoding and raw data into the provided hash.
-func (cm *Meta) writeHash(h hash.Hash) error {
-	if _, err := h.Write([]byte{byte(cm.Chunk.Encoding())}); err != nil {
+// 'buf' byte slice is used to avoid allocating new byte slice. Assumes that len(buf) is 0.
+func (cm *Meta) writeHash(h hash.Hash, buf []byte) error {
+	buf = append(buf[:0], byte(cm.Chunk.Encoding()))
```
bwplotka: Hm.. are we trying to save here one byte of allocation per ops? (: Also not sure if it really assumes that `len(buf)` is 0 right?

codesome: I forgot the remove the old comment (removed now), we don't assume `len(buf)` to be `0` now.

> Hm.. are we trying to save here one byte of allocation per ops

Nope. Slice header = 2 bytes (?), 1 byte for the data. And this is called per series per chunk. So it would be millions of allocs at a large scale. This is the benchmark difference just for the `writeHash` (the base contains optimization in `WriteChunks`).
```
BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     14825915       12829175       -13.47%
```

Apparently, in my benchmark, about 60-67% of allocs savings is from `writeHash` and remaining from re-using buffer in `WriteChunks`.

csmarchbanks: Wouldn't new slice headers be created when passing the slice around, so the only saving is the single byte? That said, I can see how this being in an inner loop could save a lot of gc.

codesome: > Wouldn't new slice headers be created when passing the slice around

Interesting, I didn't know/think about it!

bwplotka: > Nope. Slice header = 2 bytes (?), 

https://golang.org/pkg/reflect/#SliceHeader

I can see 3 integers so it's definitely not 2 bytes -> rather 12 if those as those probably are int64 (:

codesome: Shows that I was in a hurry :sweat_smile:. I considered 1 byte per variable without much thought!

The savings in allocs with this is good anyway, so WDYT?

chunks/chunks.go
```diff
@@ -97,6 +98,7 @@ type Writer struct {
 	wbuf    *bufio.Writer
 	n       int64
 	crc32   hash.Hash
+	buf     [binary.MaxVarintLen32]byte
```
csmarchbanks: A thought here, by moving this array to the chunk, it will now always be allocated to the heap, rather than having a chance of being allocated to the stack when it was in the function correct? Could that make things actually less efficient overall?

Not sure if it matters, but I am a bit curious to have a brief discussion.

chunks/chunks.go
bwplotka: Yup :+1:

chunks/chunks.go
codesome: `chunks.Writer` is created only once per compaction, so I would not worry much about it (5-byte slice) being allocated either in heap or stack (would it make a big difference?). This looked cleaner hence kept it like that.

chunks/chunks.go
```diff
@@ -57,8 +57,9 @@ type Meta struct {
 }
 
 // writeHash writes the chunk encoding and raw data into the provided hash.
-func (cm *Meta) writeHash(h hash.Hash) error {
-	if _, err := h.Write([]byte{byte(cm.Chunk.Encoding())}); err != nil {
+func (cm *Meta) writeHash(h hash.Hash, buf []byte) error {
+	buf = append(buf[:0], byte(cm.Chunk.Encoding()))
```
gouthamve: Can we be consistent here and do `buf[0] = byte(...)` as we do here: https://github.com/prometheus/tsdb/pull/653/files#diff-ba6ef14901e90aea742f3fd2909d7f07R315

chunks/chunks.go
```diff
@@ -299,33 +301,30 @@ func (w *Writer) WriteChunks(chks ...Meta) error {
...
-		seq = uint64(w.seq()) << 32
-	)
+	var seq = uint64(w.seq()) << 32
 	for i := range chks {
 		chk := &chks[i]
 
 		chk.Ref = seq | uint64(w.n)
 
-		n := binary.PutUvarint(b[:], uint64(len(chk.Chunk.Bytes())))
+		n := binary.PutUvarint(w.buf[:], uint64(len(chk.Chunk.Bytes())))
```
gouthamve: Curious what the difference b/w passing `w.buf` and `w.buf[:]` is...

codesome: `w.buf` is an array, but `PutUvarint` expects a slice. `w.buf[:]` would convert an array into a slice.

chunks/chunks.go
codesome: As `writeHash` is not a part of `chunks.Writer`, I would not make any size assumptions of the passed buffer. It would be safer to append. The one you pointed is a `chunks.Writer` method and is assured of it's size. (But if you think it is fine as it is all internal, I can make the change)

codesome:
> but these changes honestly look like micro-optimisations to me

~19% savings in allocs, would not call it micro-opt :)
All these optimizations add up to 76-77% saved allocs during compaction. https://github.com/prometheus/tsdb/pull/627#issuecomment-503453181

chunks/chunks.go
bwplotka: Nice!

bwplotka:
:+1:

krasi-georgiev:
@codesome I would also prefer to run a prombench test with all these before merging to justify this additional complexity.

codesome:
@krasi-georgiev given that I have merged multiple optimizations, do you want me to run prombench for each of them separately?

krasi-georgiev:
I would guess each one would be a bit of a headache so maybe all at once.
However you prefer, I don't mind.
Main point is to decide whether it is worth this extra mental baggage.

codesome:
I will start prombench with all together on Monday as I wouldn't be able to track on the weekends.

PR #654, opened 2019-07-05
Reuse string buffer in stringTuples.Swap

codesome:

### DIFF
--- a/index/index.go
+++ b/index/index.go
@@ -951,25 +951,28 @@ func (r *Reader) LabelNames() ([]string, error) {
 type stringTuples struct {
 	length  int      // tuple length
 	entries []string // flattened tuple entries
+	swapBuf []string
 }
 
 func NewStringTuples(entries []string, length int) (*stringTuples, error) {
 	if len(entries)%length != 0 {
 		return nil, errors.Wrap(encoding.ErrInvalidSize, "string tuple list")
 	}
-	return &stringTuples{entries: entries, length: length}, nil
+	return &stringTuples{
+		entries: entries,
+		length:  length,
+		swapBuf: make([]string, length),
+	}, nil
 }
 
 func (t *stringTuples) Len() int                   { return len(t.entries) / t.length }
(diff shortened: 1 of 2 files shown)


codesome:
Here are the benchmark results
```
BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     12925384892     12786361556     -1.08%

BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     15836497       10150083       -35.91%

BenchmarkCompaction/type=normal,blocks=2,series=1000000,samplesPerSeriesPerBlock=51-8     2080160880     1988797488     -4.39%
```

index/index.go
```diff
@@ -951,25 +951,28 @@ func (r *Reader) LabelNames() ([]string, error) {
...
 
 func NewStringTuples(entries []string, length int) (*stringTuples, error) {
 	if len(entries)%length != 0 {
 		return nil, errors.Wrap(encoding.ErrInvalidSize, "string tuple list")
 	}
-	return &stringTuples{entries: entries, length: length}, nil
+	return &stringTuples{
+		entries: entries,
+		length:  length,
+		swapBuf: make([]string, length),
```
csmarchbanks: Would it be worthwhile to lazily initialize the swap buffer? How often are string tuples created and `Swap` never called?

codesome: During compaction it would always call `Swap` (via the `Sort` method) https://github.com/prometheus/tsdb/blob/d230c67aa180850b80ae49e07079f55df1da0502/index/index.go#L412-L416

But the one returned by head (used in `querier.go`) does not call sort `Sort` on it. https://github.com/prometheus/tsdb/blob/d230c67aa180850b80ae49e07079f55df1da0502/head.go#L1230

I will make it lazy. I was just worried about lot of redundant checks in the `Swap`, but I guess it won't hurt perf.

PR #659, opened 2019-07-11
[WIP] Add Chunks method to Series interface.

tomwilkie:

This is to enable Cortex to use TSDB, and so the remote_read API doesn't need to re-encode chunks.

/cc @bwplotka.

TODO
- CHANGELOG
- Tests

### DIFF
--- a/querier.go
+++ b/querier.go
@@ -54,6 +54,9 @@ type Series interface {
 
 	// Iterator returns a new iterator of the data of the series.
 	Iterator() SeriesIterator
+
+	// Chunks returns a copy of the compressed chunks that make up this series.
+	Chunks() []chunks.Meta
 }
 
 // querier aggregates querying results from time blocks within
@@ -876,6 +879,10 @@ func (s *chunkSeries) Iterator() SeriesIterator {
 	return newChunkSeriesIterator(s.chunks, s.intervals, s.mint, s.maxt)
 }
 
+func (s *chunkSeries) Chunks() []chunks.Meta {
+	return s.chunks
+}
+
 // SeriesIterator iterates over the data of a time series.
 type SeriesIterator interface {
 	// Seek advances the iterator forward to the given timestamp.
@@ -902,6 +909,14 @@ func (s *chainedSeries) Labels() labels.Labels {
 
 func (s *chainedSeries) Iterator() SeriesIterator {
 	return newChainedSeriesIterator(s.series...)
+}
+
+func (s *chainedSeries) Chunks() []chunks.Meta {
+	var chunks []chunks.Meta
+	for _, s := range s.series {
+		chunks = append(s.Chunks())
+	}
+	return chunks
 }
 
 // chainedSeriesIterator implements a series iterater over a list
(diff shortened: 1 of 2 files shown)


querier.go
```diff
@@ -904,6 +911,14 @@ func (s *chainedSeries) Iterator() SeriesIterator {
 	return newChainedSeriesIterator(s.series...)
 }
 
+func (s *chainedSeries) Chunks() []chunks.Meta {
+	var chunks []chunks.Meta
```
codesome: Size of the slice can be preallocated to `len(s.series)`

querier.go
```diff
@@ -961,6 +976,14 @@ func (it *chainedSeriesIterator) Err() error {
 	return it.cur.Err()
 }
 
+func (s *verticalChainedSeries) Chunks() []chunks.Meta {
+	var chunks []chunks.Meta
```
codesome: Size of the slice can be preallocated to `len(s.series)`

querier.go
```diff
@@ -54,6 +54,9 @@ type Series interface {
 
 	// Iterator returns a new iterator of the data of the series.
```
bwplotka: Worth to mention that we iterate here over raw samples? `SeriesIterator` does not tell that.

querier.go
```diff
@@ -876,6 +879,10 @@ func (s *chunkSeries) Iterator() SeriesIterator {
 	return newChunkSeriesIterator(s.chunks, s.intervals, s.mint, s.maxt)
 }
 
+func (s *chunkSeries) Chunks() []chunks.Meta {
+	return s.chunks
```
bwplotka: Is that a copy of chunks as comment suggests? I think it is exactly the same underlying array passed around.

querier.go
```diff
@@ -54,6 +54,9 @@ type Series interface {
 
 	// Iterator returns a new iterator of the data of the series.
 	Iterator() SeriesIterator
+
+	// Chunks returns a copy of the compressed chunks that make up this series.
+	Chunks() []chunks.Meta
```
bwplotka: Similar to the comment above, `Chunks` is breaking the consistency here a bit, becaue we have `Iterator` above so maybe `ChunksIterator` would be more appropriate? and have `ChunkIterator` iterator that iterates over chunks.

I know iterator interface is sometimes bit controversial (e.g very hard to read and debug), but especially for remote read use case it might be actually more valuable to have it in `iterator`. This is to so potentially we don't need to allocate the whole slice or chunks within `Series interface implementation` once we compose the frame. For cortex use case, I am not sure as you just mentioned: `This is to enable Cortex to use TSDB`  What do you think?

> of the compressed chunks 

Does it mean we guarantee ALL implementations to return chunks with `chunks.Meta.Bytes` and not only `Ref`? I don't think we can guarantee that so we might want to add `or ref to chunk in chunk file`

bwplotka: Proposed iterators here: https://github.com/prometheus/tsdb/pull/665 (: Let me know if this makes sense.

bwplotka:
Proposed alternative: https://github.com/prometheus/tsdb/pull/665

querier.go
brian-brazil: I think the iterator approach is better.

tomwilkie:
Closing for the alternative.

PR #660, opened 2019-07-12
enhancement: flush the page more precisely

YaoZengzeng:
1. don't flush the page if it's last fragment of the record

2. if it's last record of the bacth, flush the page after it written into the page

### DIFF
--- a/wal/wal.go
+++ b/wal/wal.go
@@ -615,17 +615,21 @@ func (w *WAL) log(rec []byte, final bool) error {
 		copy(buf[recordHeaderSize:], part)
 		p.alloc += len(part) + recordHeaderSize
 
-		// By definition when a record is split it means its size is bigger than
-		// the page boundary so the current page would be full and needs to be flushed.
-		// On contrary if we wrote a full record, we can fit more records of the batch
-		// into the page before flushing it.
-		if final || typ != recFull || w.page.full() {
-			if err := w.flushPage(false); err != nil {
+		if w.page.full() {
+			if err := w.flushPage(true); err != nil {
 				return err
 			}
 		}
 		rec = rec[l:]
 	}
+
+	// If it's the final record of the batch and the page is not empty, flush it.
+	if final && w.page.alloc > 0 {
+		if err := w.flushPage(false); err != nil {
+			return err
+		}
+	}
+
 	return nil
 }
 


YaoZengzeng:
@codesome @krasi-georgiev 

PTAL? :joy:

krasi-georgiev:
started working on a Thanos PR so will get back when I am done there which might be about a week or more.

YaoZengzeng:
@krasi-georgiev Updated.

krasi-georgiev:
ping @gouthamve 	, @codesome 	@bwplotka

wal/wal.go
```diff
@@ -617,15 +617,23 @@ func (w *WAL) log(rec []byte, final bool) error {
 
 		// By definition when a record is split it means its size is bigger than
 		// the page boundary so the current page would be full and needs to be flushed.
-		// On contrary if we wrote a full record, we can fit more records of the batch
-		// into the page before flushing it.
-		if final || typ != recFull || w.page.full() {
+		// On contrary if we wrote a full record or last fragment of a record, we can
+		// fit more records of the batch into the page before flushing it.
+		if (typ != recFull && typ != recLast) || w.page.full() {
```
csmarchbanks: I think this would be clearer to read as
```
if type == recFirst || type == recMiddle || w.page.full() {
...
}
```
then you can get rid of the entire "On contrary" sentence.

wal/wal.go
```diff
@@ -617,15 +617,23 @@ func (w *WAL) log(rec []byte, final bool) error {
...
 			if err := w.flushPage(false); err != nil {
 				return err
 			}
 		}
 		rec = rec[l:]
 	}
+
+	// If it's the final record of the batch, flush the page.
+	if final {
```
csmarchbanks: If the last record made the page full it will have already flushed from the logic above and then we will flush an empty page. That doesn't seem ideal.

wal/wal.go
YaoZengzeng: @csmarchbanks Thanks for your comments, updated!

wal/wal.go
YaoZengzeng: Fixed as well.

wal/wal.go
```diff
@@ -617,15 +621,21 @@ func (w *WAL) log(rec []byte, final bool) error {
 
 		// By definition when a record is split it means its size is bigger than
 		// the page boundary so the current page would be full and needs to be flushed.
-		// On contrary if we wrote a full record, we can fit more records of the batch
-		// into the page before flushing it.
-		if final || typ != recFull || w.page.full() {
+		if typ == recFirst || typ == recMiddle || w.page.full() {
```
codesome: Looking at the above comment, I was thinking whether checking just `w.page.full()` would be enough here.

YaoZengzeng: @codesome Totally agree with you, updated.

wal/wal.go
```diff
@@ -615,17 +619,21 @@ func (w *WAL) log(rec []byte, final bool) error {
 		copy(buf[recordHeaderSize:], part)
 		p.alloc += len(part) + recordHeaderSize
 
-		// By definition when a record is split it means its size is bigger than
-		// the page boundary so the current page would be full and needs to be flushed.
-		// On contrary if we wrote a full record, we can fit more records of the batch
-		// into the page before flushing it.
-		if final || typ != recFull || w.page.full() {
-			if err := w.flushPage(false); err != nil {
+		if w.page.full() {
```
krasi-georgiev: hm , this doesn't look right, There can be a case where the page is not full, but the next record won't fit so needs to e flushed.

wal/wal.go
```diff
@@ -615,17 +619,21 @@ func (w *WAL) log(rec []byte, final bool) error {
...
+			if err := w.flushPage(true); err != nil {
 				return err
 			}
 		}
 		rec = rec[l:]
 	}
+
+	// If it's the final record of the batch and the page is not empty, flush it.
+	if final && !w.page.empty() {
```
krasi-georgiev: don't see a need for an extra func if it is called only in a single place.

wal/wal.go
codesome: `typ == recFirst || typ == recMiddle` means that the page is full, so a single check should be enough, isn't it?

Looking at the code I think it is ensured that at least the header can be put for the next record. I know putting only header sounds inefficient, but that was the case before too I believe on some occassions if `typ` was `recLast`.

wal/wal.go
csmarchbanks: There are a few checks in this file like:
```
if w.page.alloc > 0 {
```
Should that logic all be inlined into `flushPage` such that it is a noop if there is no data to flush?

YaoZengzeng: Updated, if it is necessary, I'll inline it in another PR

wal/wal.go
YaoZengzeng: @krasi-georgiev @codesome @csmarchbanks 

If typ == recFirst || type == recMiddle, then w.page.full() is engouh actually.

But if typ is recFull or recLast and the remaining space after writing this record is < recordHeaderSize. Then for next record, `l = min(len(rec), (pageSize-p.alloc)-recordHeaderSize)` will < 0 and `part=rec[:l]` will result panic.

So atfer write a full (or part of ) the record and w.page.alloc <= recordHeaderSize, then this page should be flushed.

krasi-georgiev:
failing tests

wal/wal.go
csmarchbanks: How does the panic happen?
* If there is not enough space to write a header, then `w.page.full()` will be true and the page will flush. 
* If there is exactly enough space to write a header then `part=rec[:0]`, which would wastefully write an empty part, but not panic.

If `w.page.full()` is not enough for `recFull`, wouldn't there be panics in the master code?

YaoZengzeng: @csmarchbanks My mistake, I understand page.full() literally :joy: 

As @csmarchbanks said, the code in this PR has same effect as the master code.

If the code in this PR is wrong, then the master code is wrong as well.

codesome: Nope, it does not panic. The explanation is precisely what @csmarchbanks said.

To make it less wasteful should we also check for some X bytes to be available in addition to headers before writing the record part? Not sure what that X has to be.

csmarchbanks: Making it less wasteful would be nice, but I would say it should probably be a separate PR.

YaoZengzeng:
@krasi-georgiev @csmarchbanks @codesome Any new comments?

krasi-georgiev:
LGTM

@codesome 	feel free to merge if you don't see any problems.

PR #661, opened 2019-07-13
labels: support for removing duplicate labels.

johncming:
It is possible to insert duplicate labels using the `FromStrings` function

https://github.com/prometheus/tsdb/blob/3bc1ea3d7584d67882fd4019d5bd892905966eb2/labels/labels.go#L148-L162

This pr is modified to improve the function based on 'FromMap'

### DIFF
--- a/labels/labels.go
+++ b/labels/labels.go
@@ -150,15 +150,15 @@ func FromStrings(ss ...string) Labels {
 	if len(ss)%2 != 0 {
 		panic("invalid number of strings")
 	}
-	var res Labels
-	for i := 0; i < len(ss); i += 2 {
-		if ss[i+1] != "" {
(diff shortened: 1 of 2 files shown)


brian-brazil:
Which piece of code is using it this way?

johncming:
> Which piece of code is using it this way?

 To keep the interface consistent.
1. `FromStrings`
2. `FromMap`

`FromMap` imply uniqueness, but `FromStrings` not.

brian-brazil:
It'd be a programming error if it wasn't unique, and silently throwing away data doesn't sound like the best of ideas.

johncming:
> It'd be a programming error if it wasn't unique, and silently throwing away data doesn't sound like the best of ideas.

PR #665, opened 2019-07-20
Add ChunksIterator method to Series interface.

bwplotka:
This is alternative to the https://github.com/prometheus/tsdb/pull/659 I feel like iterators is the way to go here as we want to iterate & stream chunks instead of buffering all.

Both PRs are for same use case: Allowing `Querier` to return raw chunks if asked.  This will help massively on https://github.com/prometheus/prometheus/pull/5701

This iterators implementation is easy here but will be more important & tricky for `head`  in Prometheus.


querier.go
```diff
@@ -1210,3 +1233,67 @@ type errSeriesSet struct {
...
+func (c *chainedChunkIterator) Next() bool {
+	if c.i == 0 {
+		if len(c.chain) == 0 {
+			return false
+		}
+		c.i++
+	}
+
+	for {
+		if c.chain[c.i-1].Next() {
```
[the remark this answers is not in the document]
bwplotka: Thanks, that simplified a lot :+1:

bwplotka:
Addressed comments @codesome  PTAL

querier.go
```diff
@@ -904,6 +911,14 @@ func (s *chainedSeries) Iterator() SeriesIterator {
 	return newChainedSeriesIterator(s.series...)
 }
 
+func (s *chainedSeries) ChunkIterator() ChunkIterator {
+	ch := &chainedChunkIterator{chain: make([]ChunkIterator, 0, len(s.series))}
+	for _, s := range s.series {
+		ch.chain = append(ch.chain, s.ChunkIterator())
```
brian-brazil: What are the ordering guarantees here? I'd expect this to all be in order, reencoding if needed.

bwplotka: Good point, we need to follow on what `Iterator` has I guess.

bwplotka: Ok I looked it through and to me `chainedSeriesIterator` or even `mergedSeriesSet` does not do any sorting on given series. IMO we should not do it here either in that case. I added comments that we expect sorted and non overlapping chunks.

brian-brazil: verticalChainedSeries does though.

bwplotka: Good point, missed that. :man_facepalming:  

What if we will return error (not implemented) for now for `verticalChainedSeries`? It is used by vertical compactor and we can add this implementation only when we will be starting using `ChunkIterations` for that (if any).

My point is to iterate with this faster, so we can gain benefits in remote read streaming (:

brian-brazil: I thought that using that was automatic if overlapping blocks were present? If so that'd break the remote read endpoint.

bwplotka: It is in indeed: https://github.com/prometheus/tsdb/blob/d5b3f0704379a9eaca33b711aa0097f001817fc2/db.go#L1185

Ok, let's not be lazy and try to do overlaping chunk iterators properly then :thinking:

bwplotka: I can see current policy for vertical merging is merging both series and fill the samples from both sides. This breaks all the intervals we used to have if the timestamps are slightly off for series a vs b. Is it intentional? If the timestamps are the same, we just choose arbitrary ones (b).

https://github.com/prometheus/tsdb/blob/d097d3fbf4c87463d3300f897182b269eefd5c30/querier.go#L1061 

In the same how would you treat overlapping chunks in such vertical case?

I think the outcome would be to have non-overlapping flow of chunks for the `ChunkIterator` so we need to reencode two chunks into max 120 samples new chunks probably reusing `verticalMergeSeriesIterator`... let me know if my thinking is wrong.

@codesome @brian-brazil

bwplotka: Side question:

We habe in `chunks`: 
```
// Iterator is a simple iterator that can only get the next value.
type Iterator interface {
	At() (int64, float64)
	Err() error
	Next() bool
}

```
vs
```
// SeriesIterator iterates over the data of a time series.
	// If there's no value exactly at t, it advances to the first value
	// after t.
	Seek(t int64) bool
	// At returns the current timestamp/value pair.
	At() (t int64, v float64)
	// Next advances the iterator by one.
	Next() bool
	// Err returns the current error.
	Err() error
}
```

Should I aim for having chunk iterators implement `Seek` or just convert chunk iterator to the one with `nopSeek` (returning err not implemented). ?

brian-brazil: It wouldn't hurt to have a very simple seek implementation, it doesn't need to be efficient.

bwplotka:
Added bit yolo implementation of `verticalMergeChunkIterator` - need to finish edge cases and add tests, but let me know your thoughts @brian-brazil @codesome

querier.go
```diff
@@ -1078,6 +1075,141 @@ func (it *verticalMergeSeriesIterator) Err() error {
 	return it.b.Err()
 }
 
+type noSeekSeriesIterator struct {
+	chunkenc.Iterator
+	err error
+}
+
+func (it *noSeekSeriesIterator) Seek(t int64) bool {
+	it.err = errors.New("not implemented: Seek method invoked for noSeekSeriesIterator")
```
codesome: I would rather prefer having an inefficient implementation as @brian-brazil said instead of an error here.

querier.go
```diff
@@ -1078,6 +1075,141 @@ func (it *verticalMergeSeriesIterator) Err() error {
...
+			b: s[1].ChunkIterator(),
+		}
+	}
+	return &verticalMergeChunkIterator{
+		a: s[0].ChunkIterator(),
+		b: newVerticalMergeChunkIterator(s[1:]...),
+	}
+}
+
+func (it *verticalMergeChunkIterator) Next() bool {
```
codesome: Error check before doing any operations.

querier.go
bwplotka: fair

querier.go
```diff
@@ -1078,6 +1075,141 @@ func (it *verticalMergeSeriesIterator) Err() error {
...
+	}
+	seriesIter := &verticalMergeSeriesIterator{
+		a: &noSeekSeriesIterator{Iterator: aCurMeta.Chunk.Iterator(it.aReuseIter)},
+		b: &noSeekSeriesIterator{Iterator: bCurMeta.Chunk.Iterator(it.bReuseIter)},
+	}
+
+	mint := int64(math.MaxInt64)
+	maxt := int64(0)
+
+	// TODO: This can end up being up to 240 samples per chunk, so we need to have a case to split to two.
```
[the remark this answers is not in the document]
bwplotka: Happy to fix this

querier.go
bwplotka: we are missing that in other places then as well

codesome: Not sure what are those other places, but here if one of `it.a` or `it.b` errors out, we will still go ahead with the other chunk iterator and return true.

querier.go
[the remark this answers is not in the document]
bwplotka: BTW I think https://github.com/prometheus/tsdb/blob/d5b3f0704379a9eaca33b711aa0097f001817fc2/chunks/chunks.go#L208-L240 can just reuse vertical iterators IMO.

bwplotka:
@brian-brazil very true. But we need to split chunks from time to time as mentioned here: https://github.com/prometheus/tsdb/issues/670 so I would reencode in place here.

Changed the code to reuse more stuff around, let me know how that looks.

chunkenc/chunk.go
```diff
@@ -58,9 +58,16 @@ type Appender interface {
 
 // Iterator is a simple iterator that can only get the next value.
```
brian-brazil: This is no longer true

querier.go
```diff
@@ -54,6 +55,9 @@ type Series interface {
 
 	// Iterator returns a new iterator of the data of the series.
 	Iterator() SeriesIterator
+
+	// ChunkIterator returns a new iterator for the non-overlapping chunks of the series.
```
brian-brazil: This reads as though it discards overlapping chunks.

chunkenc/chunk.go
bwplotka: yup

bwplotka: Wonder if we can totally remove `SeriesIterator` everywhere then :thinking:

bwplotka: and we have even third SeriesIterator in `tsdbutil`!

CHANGELOG.md
```diff
@@ -1,11 +1,14 @@
...
 
+- [CHANGE] `chunks.MergeOverlappingChunks` moved to `tsdb.MergeOverlappingChunks`
+
 ## 0.10.0
 
  - [FEATURE] Added `DBReadOnly` to allow opening a database in read only mode.
     - `DBReadOnly.Blocks()` exposes a slice of `BlockReader`s.
     - `BlockReader` interface - removed MinTime/MaxTime methods and now exposes the full block meta via `Meta()`.
  - [FEATURE] `chunckenc.Chunk.Iterator` method now takes a `chunckenc.Iterator` interface as an argument for reuse.
+ - [CHANGE] `Series` interface allows return chunk iterator that allows iterating over encoded chunks.
```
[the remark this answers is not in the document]
bwplotka: failed rebase

bwplotka:
Thanks @krasi-georgiev for review. Let me add few unit tests to cover all iterators I added (:

querier.go
```diff
@@ -1175,6 +1320,19 @@ func (it *deletedIterator) At() (int64, float64) {
 	return it.it.At()
 }
 
+func (it *deletedIterator) Seek(t int64) bool {
+	if atT, _ := it.At(); t >= atT {
+		return false
+	}
+
+	for it.Next() {
```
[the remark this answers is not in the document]
bwplotka: Awesome! I will add decisions there to Seek interface comment and add tests for this.

querier.go
bwplotka: Seek implementation added.

querier.go
```diff
@@ -1055,6 +1065,141 @@ func (it *verticalMergeSeriesIterator) Err() error {
...
+		return s[0].ChunkIterator()
+	} else if len(s) == 2 {
+		return &verticalMergeChunkIterator{
+			a: s[0].ChunkIterator(),
+			b: s[1].ChunkIterator(),
+		}
+	}
+	return &verticalMergeChunkIterator{
+		a: s[0].ChunkIterator(),
+		b: newVerticalMergeChunkIterator(s[1:]...),
```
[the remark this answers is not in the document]
bwplotka: The recursive approach is sometimes hard to follow indeed. Let's postpone cleaning this though, as we do this everywhere.

querier.go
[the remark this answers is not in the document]
bwplotka: Leaving that to next PR.

querier.go
```diff
@@ -872,22 +876,12 @@ func (s *chunkSeries) Labels() labels.Labels {
...
-	// after t.
-	Seek(t int64) bool
-	// At returns the current timestamp/value pair.
-	At() (t int64, v float64)
-	// Next advances the iterator by one.
-	Next() bool
-	// Err returns the current error.
-	Err() error
+func (s *chunkSeries) ChunkIterator() ChunkIterator {
+	return &chunkIterator{chunks: s.chunks}
```
bwplotka: `newChunkSeriesIterator` accepts mint and maxt. Should we do the same here? @brian-brazil @krasi-georgiev ? I think  we should.

However not sure what mint maxt means here in detail. Should we be strict and allow mint and maxt in terms of samples inside chunks as well?

By the looks of this: https://github.com/prometheus/tsdb/blob/78e9e360581153d73b1c69d31fd6e2d89e060a82/querier.go#L796 we just check if they overlap with required min or maxt here.

bwplotka: Then, we pass to `newChunkSeriesIterator with mint maxt so I am pretty sure we need to be strict here as well in `ChunkIterator`. E.g when we query block [0, 20)` with querier restricted to `[10, 40]` we need to return chunk `[10,19]` even if the block has `[0, 19]` on this `ChunkIterator` level, I assume.

querier.go
```diff
@@ -60,6 +60,19 @@ type Series interface {
 	ChunkIterator() ChunkIterator
 }
 
+// ChunkIterator iterates over the chunk of a time series.
+type ChunkIterator interface {
```
bwplotka: Moved to be near `Series` interface.

querier_test.go
```diff
@@ -1514,11 +1463,13 @@ func (it *listSeriesIterator) Seek(t int64) bool {
 	if it.idx == -1 {
 		it.idx = 0
 	}
+
 	// Do binary search between current position and end.
-	it.idx = sort.Search(len(it.list)-it.idx, func(i int) bool {
```
bwplotka: Problem here was that `sort.Search` on not found gives the index to put the value in. This makes sense but seems like we forgot that we operate on shorter `len(it.list)-it.idx` list (:

bwplotka:
With new changes this PR also fixes https://github.com/prometheus/tsdb/issues/328

querier.go
```diff
@@ -1243,38 +1260,51 @@ func (it *chunkSeriesIterator) resetCurIterator() {
 	it.cur = it.bufDelIter
 }
 
-func (it *chunkSeriesIterator) Seek(t int64) (ok bool) {
-	if t > it.maxt {
+func (it *chunkSeriesIterator) Seek(t int64) bool {
+	if it.Err() != nil || t > it.maxt || it.i > len(it.chunks)-1 {
```
krasi-georgiev: why is this extra check needed? 
Isn't this covered below the loop?

querier.go
```diff
@@ -1175,6 +1350,19 @@ func (it *deletedIterator) At() (int64, float64) {
 	return it.it.At()
 }
 
+func (it *deletedIterator) Seek(t int64) bool {
+	if atT, _ := it.At(); t >= atT {
+		return false
```
krasi-georgiev: this doesn't look right.
I think even my suggestion was wrong.

querier.go
```diff
@@ -1210,3 +1398,85 @@ type errSeriesSet struct {
...
+	chunks []chunks.Meta // series in time order
+	i      int
+}
+
+func (c *chunkIterator) Seek(t int64) bool {
+	if c.i >= len(c.chunks) {
+		return false
+	}
+
+	for c.Next() {
```
krasi-georgiev: Another call to Seek with the same timestamp is not noop here.

Also I don't see any tests for Seek for the chunk iterator.

querier_test.go
```diff
@@ -658,28 +665,52 @@ func TestBaseChunkSeries(t *testing.T) {
 
 // TODO: Remove after simpleSeries is merged
```
krasi-georgiev: :smile: TODO from 2y ago

querier.go
krasi-georgiev: > newChunkSeriesIterator accepts mint and maxt. Should we do the same here? @brian-brazil @krasi-georgiev ? I think we should.

I think it is fine to start with with the current interface if it covers the streaming use case. Don't see a problem to refactor when/if needed.

querier.go
```diff
@@ -971,21 +973,25 @@ func (s *verticalChainedSeries) Labels() labels.Labels {
 	return s.series[0].Labels()
 }
 
-func (s *verticalChainedSeries) Iterator() SeriesIterator {
+func (s *verticalChainedSeries) Iterator() chunkenc.Iterator {
 	return newVerticalMergeSeriesIterator(s.series...)
 }
 
-// verticalMergeSeriesIterator implements a series iterater over a list
+func (s *verticalChainedSeries) ChunkIterator() ChunkIterator {
```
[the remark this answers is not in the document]
krasi-georgiev: Is there a test that covers this?

querier.go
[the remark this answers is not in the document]
krasi-georgiev: I did run a debug session and the final struct is really bizarre, sort of deeply nested. 

But yeah a conversation for another PR.

querier_test.go
bwplotka: Can we kill it ? I have no context

chunkenc/chunk_test.go
```diff
@@ -67,29 +61,53 @@ func testChunk(c Chunk) error {
...
+	for it1.Next() {
+		ts, v := it1.At()
+		res1 = append(res1, pair{t: ts, v: v})
 	}
-	if it.Err() != nil {
-		return it.Err()
+	testutil.Ok(t, it1.Err())
+	testutil.Equals(t, all, res1)
+
+	// 2. Expand second iterator while reusing first one.
```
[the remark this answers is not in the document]
bwplotka: It is to ensure that reusing iterator will produce same results.

bwplotka:
@krasi-georgiev 

* I am adding tests for `func (s *verticalChainedSeries) ChunkIterator() ChunkIterator {` now
* I did not know about  #329 :scream:  I will sync on the discussions there and ensure I addressed all comments from there.

cc @free if you are happy with that. I am touching this place as well.

querier_test.go
krasi-georgiev: yep - kill it!

querier.go
```diff
@@ -60,6 +60,19 @@ type Series interface {
 	ChunkIterator() ChunkIterator
 }
 
+// ChunkIterator iterates over the chunk of a time series.
+type ChunkIterator interface {
+	// Seek advances the iterator forward to the given timestamp.
+	// It advances to the chunk with min time at t or first chunk with min time after t.
+	Seek(t int64) bool
+	// At returns the meta.
```
brian-brazil: "current value". This makes it sounds like it's dropping memes :)

chunkenc/chunk.go
```diff
@@ -56,11 +56,19 @@ type Appender interface {
 	Append(int64, float64)
 }
 
-// Iterator is a simple iterator that can only get the next value.
+// Iterator iterates over the data of a time series.
```
brian-brazil: I'd make it clearer that this is about samples, not chunks.

PR #667, opened 2019-07-22
add metric for tsdb size retention bytes

YaoZengzeng:

fixes: prometheus/prometheus#5790

### DIFF
--- a/db.go
+++ b/db.go
@@ -157,6 +157,7 @@ type dbMetrics struct {
 	startTime            prometheus.GaugeFunc
 	tombCleanTimer       prometheus.Histogram
 	blocksBytes          prometheus.Gauge
+	maxBytes             prometheus.Gauge
 	sizeRetentionCount   prometheus.Counter
 }
 
@@ -227,6 +228,10 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
 		Name: "prometheus_tsdb_storage_blocks_bytes",
 		Help: "The number of bytes that are currently used for local storage by all blocks.",
 	})
+	m.maxBytes = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "prometheus_tsdb_retention_limit_bytes",
+		Help: "Max number of bytes to be retained in the tsdb blocks, configured 0 means disabled",
+	})
 	m.sizeRetentionCount = prometheus.NewCounter(prometheus.CounterOpts{
 		Name: "prometheus_tsdb_size_retentions_total",
 		Help: "The number of times that blocks were deleted because the maximum number of bytes was exceeded.",
@@ -244,6 +249,7 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
 			m.startTime,
 			m.tombCleanTimer,
 			m.blocksBytes,
+			m.maxBytes,
 			m.sizeRetentionCount,
 		)
 	}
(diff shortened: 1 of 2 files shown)


YaoZengzeng:
@brian-brazil PTAL

krasi-georgiev:
Why do you need this?

YaoZengzeng:
@krasi-georgiev Just fix issue prometheus/prometheus#5790

db.go
```diff
@@ -227,6 +228,10 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
 		Name: "prometheus_tsdb_storage_blocks_bytes",
 		Help: "The number of bytes that are currently used for local storage by all blocks.",
 	})
+	m.sizeRetention = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "prometheus_tsdb_size_rentions_bytes",
```
brian-brazil: prometheus_tsdb_retention_byte_limit is what I'd used

YaoZengzeng:
@brian-brazil Updated, naming is always hard for me 😄 

But as @krasi-georgiev said, I'm also wondering if this metric is really necessary.

brian-brazil:
It's generally useful to have metrics like this for flags that set limits.

krasi-georgiev:
Yes I understand the use case, but seems a bit strange to add metrics that are constants like - version numbers, config flags etc.
Anyway if @brian-brazil 	is ok with this feel free to merge.

brian-brazil:
Config flags like this which are limits are useful for alerting (though this specific one isn't quite, as retention handling is automatic).

db.go
```diff
@@ -227,6 +228,10 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
 		Name: "prometheus_tsdb_storage_blocks_bytes",
 		Help: "The number of bytes that are currently used for local storage by all blocks.",
 	})
+	m.sizeRetention = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "prometheus_tsdb_retention_byte_limit",
+		Help: "Max number of bytes to be retained in the tsdb blocks, 0 or less means disabled",
```
brian-brazil: 0 means disabled.

Let's not make things harder for users than we have to. I'd also put the word configured in there.

YaoZengzeng: @brian-brazil Updated.

db.go
```diff
@@ -281,6 +287,7 @@ func Open(dir string, l log.Logger, r prometheus.Registerer, opts *Options) (db
 		chunkPool:   chunkenc.NewPool(),
 	}
 	db.metrics = newDBMetrics(db, r)
+	db.metrics.sizeRetention.Set(float64(opts.MaxBytes))
```
brian-brazil: Let's make sure that if this is <0 that this becomes 0.

db.go
```diff
@@ -227,6 +228,10 @@ func newDBMetrics(db *DB, r prometheus.Registerer) *dbMetrics {
 		Name: "prometheus_tsdb_storage_blocks_bytes",
 		Help: "The number of bytes that are currently used for local storage by all blocks.",
 	})
+	m.sizeRetention = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "prometheus_tsdb_retention_byte_limit",
```
brian-brazil: limit_bytes

krasi-georgiev:
Can you please add a test for this metric as well to ensure that the value is the same as the set value when opening a db.

look at the db size tests for some ideas.

YaoZengzeng:
@brian-brazil @krasi-georgiev PTAL

brian-brazil:
:+1:

YaoZengzeng:
@brian-brazil @krasi-georgiev Any new comments?

krasi-georgiev:
nothing on my side
LGTM

PR #668, opened 2019-07-24
cut 0.10.0

codesome:
This release brings some good allocation improvements in queries and compaction.


codesome:
Sticking these benchmark results here http://prombench.prometheus.io/grafana/d/7gmLoNDmz/prombench?orgId=1&var-RuleGroup=All&var-pr-number=5765&from=1563170890686&to=1563213975521

PR #669, opened 2019-07-25
remove redundant condition judgement whether cut new chunk

YaoZengzeng:
Actually w.n will never > w.segmentSize when WriteChunks()


krasi-georgiev:
Thanks

PR #673, opened 2019-08-04
WIP: Add tests for tsdb cli tool

obitech:
This PR introduces tests for the TSDB CLI tool, verifying the output is correct. 

A requirement for this is to factor out certain utility functions such as `createHead` or `CreateBlock` into the `testutil` package, so they can be imported by package `main`.

## Test `ls`

- `printBlocks()` now takes a `io.Writer` which can be tested with a `bytes.Buffer`. 
- in order to work around annoying `\t` expansion in `os.Stdout`, all white space is stripped from `expected` and `actual` in `main_test.go`

## TODO

- [x] Test `ls`
- [ ] Test `analyze`
- [ ] Delete original references of factored-out util functions & adjust imports in existing files


obitech:
I've just realized that there are now circular imports due to factoring out utility functions. Will have to deal with this first.

cmd/tsdb/main_test.go
```diff
@@ -0,0 +1,123 @@
...
+func TestCLIPrintBlocks(t *testing.T) {
+	db, closeFn := createTestRODBWithBlock(t)
+	defer closeFn()
+
+	var b bytes.Buffer
+	hr := false
+	tw := tabwriter.NewWriter(&b, 0, 0, 2, ' ', 0)
+	defer tw.Flush()
+
+	// Set table header
```
krasi-georgiev: don't forget full stops everywhere.

```suggestion
	// Set table header.
```

cmd/tsdb/main_test.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		os.RemoveAll(tmpdir)
+		t.Error(err)
+	}
+
+	safeDBOptions := *tsdb.DefaultOptions
+	safeDBOptions.RetentionDuration = 0
+
+	testutildb.CreateBlock(nil, tmpdir, testutildb.GenSeries(1, 1, 0, 1))
+	db, err := tsdb.Open(tmpdir, nil, nil, &safeDBOptions)
```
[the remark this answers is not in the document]
obitech: Forgot to remove this after refactoring, fixed now.

testutil/db/db.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/go-kit/kit/log"
+
+	"github.com/prometheus/tsdb"
+	"github.com/prometheus/tsdb/labels"
+	"github.com/prometheus/tsdb/testutil"
+	"github.com/prometheus/tsdb/tsdbutil"
+)
+
+// CreateBlock creates a block with given set of series and returns its dir.
+func CreateBlock(tb testing.TB, dir string, series []tsdb.Series) string {
```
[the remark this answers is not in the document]
obitech: I agree! This is going to be a bigger change though since just using the new function in the old test leads to import cycles (`tsdb` -> `testutil/db` -> `tsdb`). I'll work on testing the `analyze` function first, might be more functions that need to be moved out. I'll try to think of a solution after that, most likely involving some interface. Let me know if you have any suggestions!

obitech:
@krasi-georgiev Tests for the `analyze` cmd are done.

testutil/db/db.go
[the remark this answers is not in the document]
krasi-georgiev: in the past I have used [this](https://stackoverflow.com/questions/39028261/import-cycle-in-golang-with-test-packages). Not sure if it will help here.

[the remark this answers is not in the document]
obitech: I looked at that possibility but moving the existing tests into a new package would require moving lots of stuff around which seemed very nasty. The easiest way seems to be to just move `CreateBlock` and `GenSeries` back into the `tsdb` package (new file: `mocks.go` ) but mark them clearly as test functions: `MockCreateBlock` and `MockGenSeries`. I have also moved dependent functions and types into `mocks.go` and deleted them where they were defined previously.

mocks.go
```diff
@@ -0,0 +1,157 @@
...
+)
+
+type mockSeries struct {
+	labels   func() labels.Labels
+	iterator func() SeriesIterator
+}
+
+// MockCreateBlock creates a block with given set of series and returns its dir.
+// Intended for testing purposes.
+func MockCreateBlock(tb testing.TB, dir string, series []Series) string {
```
krasi-georgiev: I prefer the original name.  just need to find a good package name or put it one of the existing packages.

```suggestion
func CreateBlock(tb testing.TB, dir string, series []Series) string {
```

mocks.go
```diff
@@ -0,0 +1,157 @@
...
+		testutil.Ok(tb, it.Err())
+	}
+	err = app.Commit()
+	testutil.Ok(tb, err)
+	return head
+}
+
+// MockGenSeries generates series with a given number of labels and values.
+// Intended for testing purposes.
+func MockGenSeries(totalSeries, labelCount int, mint, maxt int64) []Series {
```
krasi-georgiev:
```suggestion
func GenSeries(totalSeries, labelCount int, mint, maxt int64) []Series {
```

mocks.go
obitech: Sure, I renamed it back to `CreateBlock`.

mocks.go
obitech: Fixed

obitech:
Should I squash into a single commit?

mocks.go
```diff
@@ -0,0 +1,157 @@
...
+)
+
+type mockSeries struct {
+	labels   func() labels.Labels
+	iterator func() SeriesIterator
+}
+
+// MockCreateBlock creates a block with given set of series and returns its dir.
+// Intended for testing purposes.
+func CreateBlock(tb testing.TB, dir string, series []Series) string {
```
krasi-georgiev: No way to put this in tsdbutil package ?

obitech: Not without massive refactoring I think. A lot of the functionality in `mocks.go` depends on stuff from the `tsdb` package: for example `CreateBlock` calls `createHead` which calls `Head()` from `head.go`.  Putting all these into `tsdbutil` would mean we have to import `tsdb` into `tsdbutil`, which is a circular import (since `tsdb` imports `tsdbutil` in a lot of places). 

To avoid this I can see three options:

1. Come up with some clever interfaces that we can implement in bock `package tsdb` (+ tests), as well as in the tests of `package main`.
2. Move all existing tests into a new package, then import from both `tsdb` and `tsdbutil`.
3. Export `CreateBlock` and `GenSeries` from `tsdb`.

I think option 1 and 2 require bigger changes in regards to architecture and refactoring and would increase the size of this PR by around x5 which might be out of scope for this PR. So I thought just leaving those functions in `tsdb` and exporting them is the least-intrusive way forward for this. What do you think?

krasi-georgiev:
> Should I squash into a single commit?

no need

mocks.go
krasi-georgiev: I am not sure, 
would be interested to see what @codesome @gouthamve 	@bwplotka 	think.

PR #675, opened 2019-08-06
code optimization

dmitsh:
Signed-off-by: Dmitry Shmulevich <dmitry.shmulevich@sysdig.com>

I've been running performance Prometheus performance profiling and noticed excessive CPU usage by labels.New() function.
This simple optimization has improved CPU usage by 5% in my tests.

### DIFF
--- a/labels/labels.go
+++ b/labels/labels.go
@@ -123,9 +123,9 @@ func (ls Labels) WithoutEmpty() Labels {
 // New returns a sorted Labels from the given labels.
 // The caller has to guarantee that all label names are unique.
 func New(ls ...Label) Labels {
-	set := make(Labels, 0, len(ls))
-	for _, l := range ls {
-		set = append(set, l)
+	set := make(Labels, len(ls))
+	for i, l := range ls {
+		set[i] = l
 	}
 	sort.Sort(set)
 


dmitsh:
@krasi-georgiev would you be able to restart the test? From the logs it seems some infra issue, not related to the code change. Are you the right person for this type of requests?

labels/labels.go
```diff
@@ -123,9 +123,9 @@ func (ls Labels) WithoutEmpty() Labels {
 // New returns a sorted Labels from the given labels.
 // The caller has to guarantee that all label names are unique.
 func New(ls ...Label) Labels {
-	set := make(Labels, 0, len(ls))
-	for _, l := range ls {
-		set = append(set, l)
+	set := make(Labels, len(ls))
+	for i, l := range ls {
+		set[i] = l
```
krasi-georgiev: That is strange I would think that the append would handle this the same way.

Can you maybe add a bench test and show the before and after this change?

csmarchbanks: Yeah, I am pretty sure the go compiler optimizes this away in recent versions. Also, if optimizing this, I think we could get rid of the loop and just do:
```
copy(set, ls)
```

csmarchbanks (commented):
I am curious what code path you are seeing this slowdown in? I don't see any usages of this function outside of test files.

PR #677, opened 2019-08-13
Update README to notify about repository move.

bwplotka:


codesome (approved):

codesome:
Oops, I was adding another comment.

We may as well add that any future issues/PRs related to storage engine needs to be opened in prometheus repo (that would be enforced by archiving I guess), and a note saying if you have PR in this repo, you need to port it to prometheus repo. This can be done after it is merged with prometheus.

